PR #59, opened 2021-03-28
[shares] Use orml crates

### LINKED ISSUE #26: Change "Shares" to ORML tokens and multicurrency
This is quite the refactor, but I think should ultimately be done to align Zeitgeist to other Polkadot ecosystem standards.

c410-f3r:
Fixes #26

#### `ZTG`

`ZTG` token is now described as an individual struct `Ztg` that can be incorporated into a pool share by simply adding a new enum `Asset` variant

```rust
enum Asset {
   SomeAsset(SomeAsse),
   Ztg(Ztg),
}

impl From<Ztg> for Asset {
   fn from(from: Ztg) -> Self {
       Asset::Ztg(from)
   }
}
```


lsaether:
Can you merge the latest changes in main?

c410-f3r:
Pretty sure it is already updated. Let me check again

lsaether:
```
// Mapper
[POOL_A | USER_A] -> SHARE_UUID_1
[POOL_B | USER_B] -> SHARE_UUID_2

// PoolAsests
[Pool_A] -> { ASSET_A, ASSET_B }
[Pool_B] -> { ASSET_C, ASSET_D }

// Shares
[SHARE_UUID_1 | ASSET_A] -> W
[SHARE_UUID_1 | ASSET_B] -> X
[SHARE_UUID_2 | ASSET_C] -> Y
[SHARE_UUID_2 | ASSET_D] -> Z
```

In this storage schema, I believe that `SHARE_UUID_*` is an unnecessary indirection. We don't need to store a `UUID` per Pool / Shares and User, but rather assume `SHARE_UUID` is external information that is provided by the user and can be generated deterministically in other pallets. It should be a `Hash` type I think, and it can be converted to the `Asset` type similar to how you are doing for the `Ztg` struct. Then we only need the `currency_id` which would be `Asset::Shares(some_hash)` for a particular share type.

zrml/shares/src/lib.rs
```diff
@@ -1,313 +1,143 @@
...
         Ok(())
     }
 
-    fn destroy_all(share_id: T::Hash) -> DispatchResult {
-        <Accounts<T>>::remove_prefix(share_id);
-        <TotalSupply<T>>::remove(share_id);
-
-        Ok(())
+    fn destroy_all(_share_id: Asset<Self::Hash>) -> DispatchResult {
+        todo!()
```
c410-f3r, line 82: Not sure if this is possible because to implement.

lsaether: Yeah doesn't seem possible without modifying orml_tokens to add it

c410-f3r: Everything else is good though. Perhaps 'shares' is no longer needed

lsaether: Oh wait, I think we could do something like what's done in frame_system [here](https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L419) By calling [`kill_prefix`](https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/unhashed.rs#L95) directly on the prefix of the storage

lsaether: NVM, the double map for [account storage in orml_tokens](https://github.com/open-web3-stack/open-runtime-module-library/blob/0.2.0/tokens/src/lib.rs#L188) is keyed first by `AccountId` which means there will not be a shared prefix for an entire shares type, like before. Bummer.

primitives/src/lib.rs
```diff
@@ -1,11 +1,16 @@
...
+mod asset;
+
+pub use asset::Asset;
 use sp_runtime::{
     generic,
     traits::{IdentifyAccount, Verify},
     MultiSignature,
 };
 
+pub const BASE: u128 = 10 * 18;
```
lsaether, line 12:
```suggestion
pub const BASE: u128 = 10_000_000_000
```

c410-f3r: Thanks!

c410-f3r:
**Still need to tweak `Asset`.**

The trait stuff worked. I guess `traits` and `shares` crates are not needed any more.

zrml/prediction-markets/src/tests.rs
```diff
@@ -138,15 +143,17 @@ fn it_allows_to_deploy_a_pool() {
             100 * BASE,
         ));
 
-        assert_ok!(Shares::wrap_native_currency(
+        assert_ok!(Balances::transfer(
             Origin::signed(BOB),
+            <Test as crate::Trait>::ModuleId::get().into_account(),
             100 * BASE
         ));
+        assert_ok!(Shares::generate(Asset::Ztg, &BOB, 100 * BASE));
```
c410-f3r, line 151: This is roughly the same as `wrap_native_currency`, but sure if the semantics are correct though

zrml/shares/src/lib.rs
```diff
@@ -1,313 +1,152 @@
...
-    fn destroy_all(share_id: T::Hash) -> DispatchResult {
-        <Accounts<T>>::remove_prefix(share_id);
-        <TotalSupply<T>>::remove(share_id);
-
+    fn destroy_all(asset: Asset<Self::Hash>) -> DispatchResult {
+        for account_id in T::Currency::accounts_by_asset(&asset) {
+            let reserved_balance = T::Currency::reserved_balance(asset, &account_id);
+            T::Currency::unreserve(asset, &account_id, reserved_balance);
+            let free_balance = T::Currency::free_balance(asset, &account_id);
+            T::Currency::slash(asset, &account_id, free_balance);
```
lsaether, line 89: I think what we want to do is instead add another method to the super trait that can access the `remove` method on the storage value directly so we just wipe it clean from storage.

zrml/prediction-markets/src/tests.rs
lsaether: Is `Asset::Ztg` distinct from Balances? Like does `orml_tokens` still have a separate storage for it from balances?

lsaether: If so, can we use `orml_currencies` to replace `zrml_shares` and implement a trait for it to have `destroy_all`?

zrml/shares/src/lib.rs
c410-f3r: Thanks. Just moved `destroy_all` logic into `ZeitgeistMultiReservableCurrency`.

zrml/prediction-markets/src/tests.rs
c410-f3r: Implementation-wise, `PredictionMarkets::Shares` has a separated storage attached to `zrml_shares::Module` just like the old one had, the only difference is the usage of `orml_tokens` instead of `pallet_balances`.
`zrml_shares` is now just a wrapper around the `orml_traits` trait methods, therefore, it doesn't seem necessary any more.

lsaether: Yep makes sense to remove `zrml_shares` now

lsaether: But, can we make Balances ZTG interoperable with tokens in pallets such as Swaps without needing to first convert Balances ZTG storage to Tokens ZTG storage?

c410-f3r: Both have separated storages, which makes sense since they are two different things. The only way to seamlessly exchange tokens between `Balances` and `Tokens` is, in my opinion, to have the same balance type `Asset`.
With this PR, every token `Ztg` token took from the balance must sync with the pool accordingly by issuing command with `Asset::Ztg`.

zrml/prediction-markets/src/lib.rs
```diff
@@ -77,11 +72,7 @@ pub trait Trait: frame_system::Trait + pallet_timestamp::Trait {
 
     type Currency: ReservableCurrency<Self::AccountId>;
 
-    type Shares: Shares<Self::AccountId, Self::Hash, Balance = BalanceOf<Self>>
-        + WrapperShares<Self::AccountId, Self::Hash, Balance = BalanceOf<Self>>;
-
-    /// The identifier of individual markets.
-    type MarketId: AtLeast32Bit + Parameter + Member + MaybeSerializeDeserialize + Default + Copy;
```
lsaether, line 84: Can we keep the MarketId type for documentation purposed even though it's kind of useless in practice?

zrml/prediction-markets/src/tests.rs
lsaether: That all makes sense, but my sticking point is the following:

- ZTG is Balances
- Swaps works with Tokens
- Using ZTG in Swaps means first creating the separate instance of "ZTG token" on Tokens

Please correct me if my assumption is wrong.

We can make it so that Swaps works with "ZTG Balances" without requiring to turn into "ZTG token" by using an implementation that calls either Balances or Tokens based on the provided `Asset` type. This is what ORML `MultiCurrency` trait and `orml_currencies` does. So I think we can finish by 1) removing Shares, 2) use `MultiReservableCurrency` everywhere `ReservableShares` was used 3) adding `orml_currencies`. WDYT?

c410-f3r: Things are clear now, thanks for the explanation. I will use `orml_currencies` for pools but it might also be extended to prediction markets and order books

Just for the record, `orml_currencies` methods are roughly composed by an addition storage and
```rust
if currency_id == T::GetNativeCurrencyId::get() {
	... stuff
} else {
	... more stuff
}
```
to track what is and isn't `native`. Not sure if the additional capacity and processing overhead will be beneficial. What will change? Faster native queries?

lsaether: It's O(1),  just adding a single storage read to the methods. It's definitely beneficial, both to the end user who doesn't need to manage two different types of ZTG, and for the chain's storage size since we only have 1 storage for ZTG instead of 2 (which could be at worst double the storage size, for no reason). 

Not sure if that answers your question or not, so let me know if it didn't.

c410-f3r: Thank you very much for the answer, I understand now.

zrml/prediction-markets/src/lib.rs
c410-f3r: No problem. Just a nit: `Asset` must have yet another type for `MarketId` but `MarketId` is not used everywhere like `Hash` so expect more trait associated types for `MarketId`.

c410-f3r:
Unfortunately, it isn't possible to use `orml_currencies` because the module doesn't expose the underlying storage to create our custom methods. Happy to follow any other suggestion in this regard

c410-f3r:
Sorry, I was brain farting. IT might be possible although I am not 100% sure.

c410-f3r:
All tests are passing, `orml_swaps` is using `orml_currencies` instances and `MarkerId` was re-introduced.

primitives/src/zeitgeist_currencies_extension.rs
```diff
@@ -0,0 +1,23 @@
...
+    Self::MultiCurrency: ZeitgeistMultiReservableCurrency<Self::AccountId>,
+{
+    fn destroy_all(currency_id: CurrencyIdOf<Self>) -> DispatchResult {
+        let accounts = Self::MultiCurrency::accounts_by_currency_id(currency_id);
+        // Destroy pool
+        Self::MultiCurrency::destroy_all(currency_id, accounts.iter().cloned());
+        // Give back to accounts
+        for (k0, v) in accounts {
+            Self::NativeCurrency::deposit(&k0, v.free + v.reserved)?;
+        }
```
lsaether, line 20:
```suggestion
```

This won't be needed. The operation that this is used for is when a market has been resolved and a whole `currency_id` should just simply be wiped from storage because they are the losing shares and therefore worthless.

primitives/src/zeitgeist_multi_reservable_currency.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn destroy_all<I>(currency_id: Self::CurrencyId, accounts: I)
+    where
+        I: Iterator<Item = (T::AccountId, AccountData<Self::Balance>)>,
+    {
+        let mut total = Self::Balance::zero();
+        for (k0, v) in accounts {
+            <Accounts<T>>::remove(k0, currency_id);
+            total += v.free + v.reserved;
+        }
+        <TotalIssuance<T>>::mutate(currency_id, |v| *v -= total);
```
lsaether, line 47:
```suggestion
        <TotalIssuance<T>>::remove();
```

runtime/src/lib.rs
```diff
@@ -327,8 +330,8 @@ construct_runtime!(
         Balances: pallet_balances::{Module, Call, Storage, Config<T>, Event<T>}, // 5
         TransactionPayment: pallet_transaction_payment::{Module, Storage}, // 6
         Sudo: pallet_sudo::{Module, Call, Config<T>, Storage, Event<T>}, // 7
+        Tokens: orml_tokens::{Event<T>},
```
lsaether, line 333: We need to add Currencies here too - I can do it

lsaether (approved):
Nice! I'll make the suggested changes in a follow up

PR #60, opened 2021-04-01
Add currencies to zeitgeist runtime

lsaether:


c410-f3r (approved):
LGFM. Thanks for filling the remaining changes

PR #61, opened 2021-04-02
Add parachain support

c410-f3r:
Parachain nodes are collators, i.e., they simply collect txs from users to send to kusama/polkadot validators and don't perform any validation/finalization logic (Aura/Granpa).


lsaether:
Have you tested this by connecting to a local Relay Chain?

We are going to need to maintain this as a second branch I think, since we need to also maintain our test network which runs on PoA consensus. Have you taken a look at [this video about maintaining dual consensus in a single codebase](https://youtu.be/MsViDtbXp4g), we should follow this approach imo until we can shed the necessity of needing a solo network for the testnet.

lsaether:
42c792a doesn't seem to actually speed up the CI. Before it would complete in 30 mins [see](https://github.com/zeitgeistpm/zeitgeist/runs/2248020145), now it takes over 45 mins [see](https://github.com/zeitgeistpm/zeitgeist/runs/2260344238)

c410-f3r:
> Have you tested this by connecting to a local Relay Chain?

We are going to need to maintain this as a second branch I think, since we need to also maintain our test network which runs on PoA consensus. Have you taken a look at this video about maintaining dual consensus in a single codebase, we should follow this approach imo until we can shed the necessity of needing a solo network for the testnet.

I will test connectivity today. Dual consensus probably won't be a problem

> 42c792a doesn't seem to actually speed up the CI. Before it would complete in 30 mins see, now it takes over 45 mins see

https://github.com/zeitgeistpm/zeitgeist/runs/2248020145 was before the introduction of this PR.
Polkadot, Cumulus, and Substrate together make the compilation take much longer and I guess the Parity guys still need the tweak performance and memory consumption in this regard.

c410-f3r:
Took longer than I originally expected.

* To bootstrap a local polkadot network with zeitgeist parachain, an auxiliary `parachain.sh` script was created in the `scripts` folder. Locally, everything was working accordingly with `https://polkadot.js.org/apps`.
* The old standalone node is now the default choice when building `node`. To build a parachain, just add the `parachain` feature -> `cargo build --features parachain`.
* CI is not building parachain because it takes too much time (+1 hour) and memory (+18GB) to compile

lsaether:
Cool! Thanks for changing over the pallet style to Frame v2 too. 

Let me check this out locally. It would be good to add the instructions to the README as well.

c410-f3r:
A lot has changed. If desirable, I can create multiple and smaller PRs

lsaether:
It's not necessary for this one, but in the future smaller PRs would be preferred simply because it's easier to review.

I tested this last night and the standalone node worked well, the `parachain.sh` script worked too, but the parachain collator wasn't producing blocks, any thoughts for why?

PR #66, opened 2021-04-08
Parachain improvements

c410-f3r:
Improves the local parachain development script by doing every necessary step but the parachain auction/addition. For now, it is necessary to manually register the Zeitgest parachain to produce blocks (https://substrate.dev/cumulus-workshop/#/en/3-parachains/2-register?id=registration-transaction).

Also adds documentation and removes some left-over code duplications from the previous parachain support PR.


lsaether (approved):
LGTM

PR #76, opened 2021-04-15
Implements scalar markets

### LINKED ISSUE #14: Implement Scalar markets

lsaether:
closes #14 

Adds an `Asset` variant to distinguish between a `CategoricalOutcome` asset and a `ScalarOutcome` asset. Removes the `Shares(Hash)` variant since using hash as identifiers has been deprecated since we moved to using `orml_currencies` instead.

Removes the need to add special fields for the different market types. Instead we use an enum that holds the relevant values per the type of market. Example: `MarketType::Scalar((u128, u128))` holds the `(lower_bound, upper_bound)` for that market. `MarketType::Categorical(u16)` holds the number of categories for that categorical market.

Scalar markets have a different payout logic, so that's being pattern matched inside of the `redeem` function now.

Trying to make the architecture of the pallet more extensible with anticipation for the implementation of the `CombinatorialOutcome` for combinatorial markets, which are right now just stubbed.

Still need to work on tests and finishing the logic.


primitives/src/asset.rs
```diff
@@ -14,9 +14,29 @@
     parity_scale_codec::Decode,
     parity_scale_codec::Encode,
 )]
-pub enum Asset<H, MI> {
-    Share(H),
-    PredictionMarketShare(MI, u16),
+pub enum Asset<MI> {
```
c410-f3r, line 17: Nice. One less type

zrml/prediction-markets/src/market.rs
```diff
@@ -25,8 +25,8 @@ pub enum MarketEnd<BlockNumber> {
 #[derive(Eq, PartialEq, Encode, Decode, Clone, RuntimeDebug)]
 pub enum MarketType {
     // A market with a number of categorical outcomes.
-    Categorical,
-    Scalar,
+    Categorical(u16),
+    Scalar((u128, u128)),
```
c410-f3r, line 29: Maybe use `RangeInclusive<u128>` or `Range<u128>` (exclusive) as a self-documented structure?

c410-f3r (commented):
Not very experient with the PM crate but the changes look promising for the upcoming combinatorial markets

c410-f3r (approved):

zrml/prediction-markets/src/lib.rs
```diff
@@ -232,26 +234,61 @@ mod pallet {
...
+            };
+
+            <Markets<T>>::insert(market_id.clone(), Some(market));
+
+            Self::deposit_event(Event::MarketCreated(market_id, sender));
+
+            Ok(())
+        }
+
+        #[pallet::weight(10_000)]
```
sea212, line 278: Weight is too low. For example, all functions we have so far in the Swaps pallet have a lower weight bound of 50 million (excl. db interactions)

zrml/prediction-markets/src/tests.rs
```diff
@@ -463,10 +499,14 @@ fn the_entire_market_lifecycle_works_with_timestamps() {
...
         );
 
-        assert_ok!(PredictionMarkets::report(Origin::signed(BOB), 0, 1,));
+        assert_ok!(PredictionMarkets::report(
+            Origin::signed(BOB),
+            0,
+            Outcome::Categorical(1)
+        ));
     });
 }
```
sea212, line 512: General: create_scalar_market test missing

node/src/rpc.rs
```diff
@@ -32,8 +32,7 @@ where
     C: Send + Sync + 'static,
     C::Api: substrate_frame_rpc_system::AccountNonceApi<Block, AccountId, Index>,
     C::Api: pallet_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>,
-    C::Api: zrml_prediction_markets_rpc::PredictionMarketsRuntimeApi<Block, MarketId, Hash>,
```
sea212, line 35: Curious: Why was this removed?

sea212 (changes requested):
In addition: The project does not compile with `runtime-benchmarks` feature, because the benchmarks for the swaps pallet use the old `Asset` type.

runtime/src/lib.rs
```diff
@@ -530,19 +530,13 @@ impl_runtime_apis! {
...
     }
 
-    impl zrml_prediction_markets_runtime_api::PredictionMarketsApi<Block, MarketId, Hash> for Runtime {
-        fn market_outcome_share_id(market_id: MarketId, outcome: u16) -> Asset<Hash, MarketId> {
-            PredictionMarkets::market_outcome_share_id(market_id, outcome)
-        }
-    }
-
-    impl zrml_swaps_runtime_api::SwapsApi<Block, PoolId, Hash, AccountId, Balance, MarketId>
+    impl zrml_swaps_runtime_api::SwapsApi<Block, PoolId, AccountId, Balance, MarketId>
```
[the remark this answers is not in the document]
lsaether: It's not longer needed. Originally it was there as more of a convenience method to get the hash of a `Asset::Share(hash)` now that we use asset types based on market index and outcome value, we don't need accesss to a hashing function.

node/src/rpc.rs
lsaether: answered above

runtime/src/lib.rs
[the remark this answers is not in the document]
sea212: That makes sense, thanks

PR #78, opened 2021-04-18
Inflation and companions

c410-f3r:
Initially intended for collator staking, this PR now embraces three different things that are related to inflation: Liquidity mining, crowdloans and of course, collator staking.
Please, keep in mind that all third-party pallets are still very young, lack documentation and doesn't provide yet the full set of intended features. 

## Staking for collators/nominators

Moonbeam implementation is being used for this area. All runtime and node parameters were included but one remaining thing is nimbus integration (https://github.com/PureStake/cumulus/tree/nimbus) as stated by https://moonbeam.network/announcements/moonbeam-collators-announcement.

## Crowdloan rewards

The Centrifuge implementation theoretically does everything we need: Provides a way to let users redeem rewards in a vesting period. The downside is that their implementation is still a work in progress so it might be worth to join forces for faster and generic development.
Although `cf-rewards` is the only reward implementation at this moment, `orml` will probably also try to create one in the future.

## Liquidity mining

These are the pre-conditions:
  1. Must start with a fixed pre-allocated amount of ZTG.
  2. For each year, a fixed predetermined percentage must be applied to the the fixed pre-allocated amount of ZTG to have a deterministic inflation upper-bound.

Given that an unknown and undetermined quantity of shares can bought or sold on any markets at any time, the next sections will show what was made to satisfy all constraints.

### Overview

Each block has a maximum allowed amount of ZTG that is distributed among the share owners of that same block. Over time this amount will increase until a market closes and then all rewards will be distributed accordingly.

### Genesis parameters

* Fixed amount of ZTG: Will form a pot/fund for transfers.
* Per block distribution: Fixed maximum amount of ZTG that a block can distribute among shares.

Only sudo can change the per block distribution parameter.

### Buy operations 

Example: Each block can distribute 1 ZTG. A block is being constructed with 1000 shares bought by several users on different markets so each share owner will receive `1 / 1000 = 0.001 ZTG` per share.
This deterministic distribution for each block enables the creation of a controlled inflation system where it is guaranteed that a certain number `N` won't never be exceed. Another example: With an initial fund of 100.000.000 ZTG for liquidity mining and a block distribution of 0.1 ZTG per block, no more than 3153600 ZTG will be distribute per year.
```
1 block = 6 seconds
1 minute = 60 blocks
1 hour = 3600 blocks
1 day = 86400 blocks
1 year = 31536000 blocks

31536000 blocks * 0.1 ZTG = 3153600 ZTG per year
```

### Sell operations 

Since a share value will fluctuate between blocks (more shares quantity equals less value and vice-versa), it would be very expensive to store all individual share values. Instead, when someone sells, the amount of subtracted shares will be used as a percentage compared to the total amount of owned shares.
Example: In a particular market, an account has 1000 shares totalling 50 ZTG and wants to sell 200 shares. The new ZTG balance will then be `50 * 0.8 (-20%) = 40 ZTG` with `1000 - 200 = 800 shares`.

### Considerations 

* Users won't be able to withdraw ZTG until the end of the market.
* `Swaps` uses the `LiquidityMining` pallet and calls methods like `add_shares_of_account_and_market` or `remove_shares_of_account_and_market`
* Still need to work on swap fees for liquidity providers
* Instead of providing ZTG, it might be worth providing a collateral ZTG (CZTG) or derivative that can be used in other DeFi block-chains for liquid staking, loans, etc...

## Misc

* Created forks because Moonbeam and Centrifuge are using outdated dependencies
* Didn't test anything yet. Maybe some integration tests are worth adding.
* WIP, so not ready for merging.|

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -413,6 +413,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "author-inherent"
+version = "0.6.0"
+source = "git+https://github.com/c410-f3r/moonbeam?branch=stuff#83ead5d2cb13a04633e89072afa6119b712caaf0"
+dependencies = [
+ "frame-support",
+ "frame-system",
+ "parity-scale-codec",
+ "sp-authorship",
+ "sp-inherents",
+ "sp-runtime",
+ "sp-std",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -806,6 +820,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "cf-reward"
+version = "0.0.1"
+source = "git+https://github.com/c410-f3r/centrifuge-chain?branch=stuff#89bc4ea3a39779ba86ba5e655c198069d6ce7fd1"
+dependencies = [
+ "crowdloan-claim",
+ "frame-support",
+ "frame-system",
+ "pallet-balances",
+ "pallet-indices",
+ "pallet-vesting",
+ "parity-scale-codec",
+ "serde",
+ "sp-runtime",
+]
+
+[[package]]
 name = "cfg-if"
 version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


c410-f3r:
Did some code and documentation updates

c410-f3r:
The changes have become quite large. For an easier review, this PR will be divided into three

PR #82, opened 2021-04-19
Implement benchmark and add weights for prediction market pallet

sea212:
Issue: #7 
Solution: Implement benchmarks and add weights for every dispatchable function in the prediction market pallet.
Progress:

**Benchmarks**
- [x] admin_destroy_market
- [x] admin_move_market_to_closed
- [x] admin_move_market_to_resolved
- [x] approve_market
- [x] buy_complete_set
- [x] cancel_pending_market
- [x] create_categorical_market
- [x] create_scalar_market
- [x] deploy_swap_pool_for_market
- [x] dispute
- [x] global_dispute _- function not implemented yet_
- [x] on_initialize
- [x] redeem_shares
- [x] reject_market
- [x] report
- [x] sell_complete_set

**Weights**
- [x] admin_destroy_market
- [x] admin_move_market_to_closed
- [x] admin_move_market_to_resolved
- [x] approve_market
- [x] buy_complete_set
- [x] cancel_pending_market
- [x] create_categorical_market
- [x] create_scalar_market
- [x] deploy_swap_pool_for_market
- [x] dispute
- [x] global_dispute _- function not implemented yet_
- [x] on_initialize
- [x] redeem_shares
- [x] reject_market
- [x] report
- [x] sell_complete_set


sea212:
I'll squash everything into 1 commit and rebase into main once it's approved

primitives/src/lib.rs
```diff
@@ -20,6 +20,8 @@ pub use zeitgeist_currencies_extension::ZeitgeistCurrenciesExtension;
 pub use zeitgeist_multi_reservable_currency::ZeitgeistMultiReservableCurrency;
 
 pub const BASE: u128 = 10_000_000_000;
+pub const MIN_LIQUIDITY: Balance = BASE * 100;
```
lsaether, line 23:
```suggestion
pub const MIN_LIQUIDITY: Balance = 100 * BASE;
```

easier to parse putting the amount first

zrml/prediction-markets/src/lib.rs
```diff
@@ -208,22 +262,27 @@ mod pallet {
         ///
         /// NOTE: This is the only way to create new shares.
         ///
-        #[pallet::weight(50_000_000)]
+        // Note: buy_complete_sets weight consumption is dependant on how many assets exists
```
lsaether, line 265:
```suggestion
        // Note: `buy_complete_set` weight consumption is dependent on how many assets exists
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -208,22 +262,27 @@ mod pallet {
...
+        // Unfortunately this information can only be retrieved with a storage call, therefore
+        // The worst-case scenario is assumed and the correct weight is calculated at the end of this function.
+        // This also occurs in numereous other functions
```
lsaether, line 268:
```suggestion
        // This also occurs in numerous other functions.
```

style => fixed typos
content => makes sense

zrml/prediction-markets/Cargo.toml
```diff
@@ -24,18 +28,20 @@ std = [
...
+    'serde',
     'sp-core/std',
     'sp-io/std',
     'sp-runtime/std',
+    'sp-std/std',
     'zeitgeist-primitives/std',
-    'zrml-swaps/std'
+    'zrml-swaps/std',
 ]
+runtime-benchmarks = ["frame-benchmarking", "orml-tokens", "pallet-balances"]
```
c410-f3r, line 41:
```suggestion
runtime-benchmarks = [
    "frame-benchmarking",
    "frame-support/runtime-benchmarks",
    "frame-system/runtime-benchmarks",
    "orml-tokens",
    "pallet-balances"
]
```

```
cargo test --features std,runtime-benchmarks --manifest-path zrml/prediction-markets/Cargo.toml --no-default-features
```

```
error[E0407]: method `add` is not a member of trait `$crate::traits::SortedMembers`
```

```
error[E0046]: not all trait items implemented, missing: `successful_origin`
```

After fixing compilation errors, `benchmarks::benchmark_tests::test_benchmarks` fails for some reason.

c410-f3r (commented):
Hey @sea212, thanks for the PR! 
Taking aside the `cargo test` stuff, everything else looks good IMO.

zrml/prediction-markets/Cargo.toml
sea212: @c410-f3r I have fixed this problem, but there is one point where the pallet behaves strange.
When the benchmark tests are invoked with the command you supplied, `deploy_swap_pool_for_market` with sufficient base currency ztg (max balance) and categorical asset is also called. But it exits with an error, "InsufficientBalance" for `Asset::Ztg`. I have checked that the account has enough base currency, but the pallet reads 0 Asset::Ztg. Seems like during the prediction market tests the swaps pallet is unable to associate Asset::Ztg with the base currency. When I deposit Asset::Ztg token to the calling account before calling `deploy_swap_pool_for_market` in the benchmark, everything works as expected.

lsaether: The reason for this is because in the mock for tests and benchmarks we [use the Tokens pallet instead of the Currencies pallet](https://github.com/zeitgeistpm/zeitgeist/pull/82/files#diff-2f51b5c84bbe63c6ee3df0a2b58d64c3fcc0cd5583ea23d2fb5b77c7464db9e5L172). We should use the Currencies pallet like how it's done in the the Swaps mock, [here](https://github.com/zeitgeistpm/zeitgeist/blob/main/zrml/swaps/src/mock.rs#L73) (we alias Shares to the Currencies pallet).

c410-f3r (approved):

zrml/prediction-markets/Cargo.toml
sea212: @lsaether done

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -0,0 +1,402 @@
...
+    Ok((caller, marketid))
+}
+
+benchmarks! {
+    admin_destroy_disputed_market{
+        // a = total accounts
+        let a in 0..10;
+        // b = num. accounts with assets
+        // Unfortunately frame-benchmarking does not allow to b = b.min(a) here
+        let b in 0..10;
```
lsaether, line 181: Why only benchmark up to 10 accounts? This seems like a rather low number compared to what could be worst case. I saw the issue you made about the bounds of the account and assets holders, shouldn't we benchmark closer to the bound?

sea212: I played around a bit with different values (account numbers and repetitions of benchmarks) and found out the following: A higher number increases the benchmark runtime significantly, while increasing the error due to a lower number of repetitions (the data points that are used to approximate the weight function weight(a) are less precise).
I assume that the reason behind this is, that the required weight per account is a linear function of degree 1, i.e. `fn(a) = ba^1 + ca^0`. If this assumption is correct (and the code mirrors that, because its behavior is independent from the actual number of accounts), the first few data points of the curve are already approximating the function and its constant gradient fairly well, provided that these few data points are very precise (many repetitions). If the function is linear and the gradient is well estimated, any value can be derived, up to infinity, assuming that at no point the function fn(a) is non-linear. At least practically (through tests of many different values), this behavior is shown.

lsaether: Makes sense, indeed it should be linear since for each iteration it performs a constant number of storage writes. Can you add this explanation to the documentation inside the benchmark file in case someone reading the code would have this question again in the future?

sea212: @lsaether done.

lsaether (approved):

PR #84, opened 2021-04-19
Follow up to scalar market PR

lsaether:
addresses issues leftover in #76 

- adds one test for scalar markets
- doesn't touch benchmarks, which is addressed in #82 
- also does a chunk of missing documentation


zrml/prediction-markets/src/lib.rs
```diff
@@ -508,41 +509,44 @@ mod pallet {
...
+                            }
+
+                            let payout_long: Percent =
+                                Percent::from_rational(final_value - low, high - low);
+                            (
+                                payout_long,
+                                Percent::from_parts(
+                                    Percent::one().deconstruct() - payout_long.deconstruct(),
+                                ),
+                            )
```
sea212, line 542: Suggestion: Use at least [Permill](https://substrate.dev/rustdocs/v3.0.0/sp_arithmetic/per_things/struct.Permill.html) for this, potentially even Perbill, because the precision for Percent is poor. In fact, the precision is 1/100 and the result is [floored to the next integer](https://crates.parity.io/sp_arithmetic/per_things/trait.PerThing.html#tymethod.from_rational), which leads to a potential inaccuracy of 0,99 (period) percent which in turn will be missing in the winning payouts.

sea212 (changes requested):
Looks good overall, but there are some things to consider before merging.

c410-f3r (approved):

zrml/prediction-markets/src/lib.rs
lsaether: Good point

sea212 (approved):

PR #88, opened 2021-04-21
Add bootnode for battery park

lsaether:
inconsequential

adds the bootnode for battery park so it doesn't need to be passed in with a cli flag

### DIFF
--- a/node/src/chain_spec/battery_park.rs
+++ b/node/src/chain_spec/battery_park.rs
@@ -42,7 +42,7 @@ pub fn battery_park_config(
             )
         },
         vec![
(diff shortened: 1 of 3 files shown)


node/src/chain_spec/battery_park.rs
```diff
@@ -41,7 +41,9 @@ pub fn battery_park_config(
                 wasm_binary,
             )
         },
-        vec![],
+        vec![
+            "/ip4/139.162.171.58/tcp/30333/p2p/12D3KooWPvu5rpH2FNYnAmiQ8X8XqkMiuSFTjH2jwMCSjoam7RGQ".parse().unwrap()
```
c410-f3r, line 45: When passing a bad peer id, `unwrap` will panic the current program execution but since this path is about node initialization, it shouldn't be a problem. If desirable, it is also possible to `.parse().map_err(|_| "Invalid boot node id")?`.

sea212 (approved):

PR #91, opened 2021-04-23
Enhance robustness of `u128` serialization/de-serialization

c410-f3r:
Tries to prevent runtime panics by tweaking some types, structures and crate features.

Some tests were created for RPC calls involving the **runtime** (RPC calls involving nodes would require a SDK) to prevent future panics related to serde `u128` serialization/de-serialization.

In order to create more tests with different inputs, it would be cool to do some external node interaction while recording the actual network activity to catch and reproduce possible errors -> https://support.box.com/hc/en-us/articles/360043696054-How-to-Generate-Network-Captures-for-Troubleshooting.

Capture d’écran de 2021-04-23 14-43-48

### DIFF
--- a/zrml/swaps/src/mock.rs
+++ b/zrml/swaps/src/mock.rs
@@ -7,8 +7,8 @@ use sp_runtime::{
     traits::{AccountIdConversion, BlakeTwo256, IdentityLookup},
 };
 use zeitgeist_primitives::{
-    AccountIdTest, Amount, Asset, Balance, BlockNumber, BlockTest, CurrencyId, Hash, Index,
-    MarketId, UncheckedExtrinsicTest, BASE,
+    AccountIdTest, Amount, Asset, Balance, BalanceInfo, BlockNumber, BlockTest, CurrencyId, Hash,
+    Index, MarketId, PoolId, UncheckedExtrinsicTest, BASE,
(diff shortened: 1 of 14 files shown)


zrml/swaps/src/mock.rs
```diff
@@ -161,3 +161,25 @@ impl ExtBuilder {
         t.into()
     }
 }
+
+sp_api::mock_impl_runtime_apis! {
+    impl zrml_swaps_runtime_api::SwapsApi<Block, PoolId, AccountIdTest, Balance, MarketId>
```
lsaether, line 166: Can we add `get_spot_prices` too, now that we have it?

lsaether, line 166: Can we expose `get_spot_prices` now too, since we have it?

lsaether (approved):

lsaether (approved):

lsaether:
@c410-f3r Can you create an issue for this so we can track it?

> In order to create more tests with different inputs, it would be cool to do some external node interaction while recording the actual network activity to catch and reproduce possible errors -> https://support.box.com/hc/en-us/articles/360043696054-How-to-Generate-Network-Captures-for-Troubleshooting.

PR #92, opened 2021-04-23
Properly fix `u128` serialization for RPC calls

c410-f3r:
@lsaether You were right, the problem was related to types or more specifically, types used for serialization, including `Asset::PoolShare(u128)`.

To trigger the panic with the current code-base, just type:
```
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "swaps_poolSharesId", "params": [0]}' http://localhost:9933/
```

### DIFF
--- a/zrml/swaps/src/tests.rs
+++ b/zrml/swaps/src/tests.rs
@@ -4,9 +4,7 @@ use crate::{
 };
 use frame_support::{assert_noop, assert_ok};
 use orml_traits::MultiCurrency;
-use sp_api::BlockId;
 use zeitgeist_primitives::{Asset, MarketId, BASE};
-use zrml_swaps_runtime_api::SwapsApi;
 
 pub const ASSET_A: Asset<MarketId> = Asset::CategoricalOutcome(0, 65);
 pub const ASSET_B: Asset<MarketId> = Asset::CategoricalOutcome(0, 66);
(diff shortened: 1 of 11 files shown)


zrml/swaps/src/tests.rs
```diff
@@ -31,17 +29,6 @@ const _100: u128 = 100 * BASE;
...
-fn rpc_serialization_deserialization_do_not_panic() {
-    ExtBuilder::default().build().execute_with(|| {
-        let runtime = Runtime;
-        let block_id = BlockId::Number(0);
-        let _ = runtime.get_spot_price(&block_id, 1, Asset::PoolShare(0), Asset::PoolShare(1));
-        let _ = runtime.pool_account_id(&block_id, 1);
-        let _ = runtime.pool_shares_id(&block_id, 1);
-    });
-}
-
```
lsaether, line 44: So the test doesn't work?

c410-f3r: Thought this test could catch the bad serialization error but it doesn't and it probably won't because the real issues lies when interacting with nodes, not the runtime.
```
User --- (calls) ---> Node --- (calls) ---> Runtime
```
Therefore, it isn't testing anything useful. Nevertheless, I can rollback this removal if desirable.

lsaether: No let's remove it since it doesn't test what it should be testing. We will need to test this using an integration test in the JS libraries.

lsaether (approved):

PR #103, opened 2021-04-27
Fix local parachain deployment

c410-f3r:
* Brings the latest versions of `polkadot`, `cumulus` and `substrate`
* Fix local parachain environment for development

Something, between `substrate` and `polkadot-apps` (`polkadot-api`? `cumulus`? A bad configuration?) is not allowing calls to `parasSudoWrapper` and the only way I could find to locally register parachains was to include them in the polkadot chain specification.

Capture d’écran de 2021-04-27 14-36-13

Taking this issue asise, our parachain is working and we can now register on Rococo.


node/src/cli/cli_parachain.rs
```diff
@@ -19,12 +19,17 @@ pub struct RelayChainCli {
 }
 
 impl RelayChainCli {
-    /// Create a new instance of `Self`.
+    /// Parse the relay chain CLI parameters using the para chain `Configuration`.
```
lsaether, line 22:
```suggestion
    /// Parse the relay chain CLI parameters using the parachain `Configuration`.
```

lsaether (approved):

PR #105, opened 2021-04-28
Benchmark orderbook-v1

sea212:
Issue: #99 
Solution: Implement benchmarks and add weights for every dispatchable function in the orderbook-v1 pallet.
Progress:

**Benchmarks**
- [x] cancel_order
- [x] fill_order
- [x] make_order

**Weights**
- [x] cancel_order
- [x] fill_order
- [x] make_order


lsaether (approved):
Interesting to note that asks require more reads / writes than bids do

c410-f3r:
`sp_std::marker::PhantomData` and `core::marker::PhantomData` are the same but it doesn't matter. In any case, `sp_std` is required by the benchmark macros IIRC

c410-f3r (approved):

PR #106, opened 2021-04-28
Enhance test coverage

c410-f3r:
The current CI checking for tests doesn't take into account some features like `parachain` and `runtime-benchmarks`, which brings building failures into `master` that are eventually fixed over time but shouldn't even exist.

To give a better testing coverage, a new `all-sequential.sh` script were included for dev and CI usage. So, once `./scripts/tests/all-sequential.sh` is successful, it is guaranteed that CI will also succeed.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,9 +11,14 @@ env:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        scripts: [
+          "misc",
+          "parachain",
+          "standalone"
+        ]
     steps:
     - uses: actions/checkout@v2
     - name: Initializes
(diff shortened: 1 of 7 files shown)


scripts/tests.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cargo fmt --all -- --check
+
+test_package_with_feature primitives std
+
+test_package_with_feature zrml/swaps std
+
+test_package_with_feature zrml/orderbook-v1 std
+
+test_package_with_feature zrml/prediction-markets std
```
lsaether, line 21: It's easy to forget to add a line here when a new pallet is created, can we make sure we include all pallets in zrml?

Somethig like
```sh
for package in zrml/*
do
  test_package_with_feature "$package" std
done
```

lsaether:
Now we don't build the WASM blob in the CI. Will the tests always guarantee that it will build?

c410-f3r:
> This change will not build the WASM blob in the CI. Will the tests always guarantee that it builds?

`make make build-wasm` can also be incorporated in the script. Just a sec

scripts/tests.sh
c410-f3r: Oh yeah, much better! Thanks

c410-f3r:
On CI, the tests were divided into 4 different parts, summing a total of `01:22:14` with the longest (parachain) taking `00:39:39`. Taking these values into consideration, we would have to run this workflow an average of 47 times every day to reach the free-tier limit.

lsaether (approved):

Issue #70, opened 2021-04-10
Store pallet parameter_types! in primitives crate


sea212:
The pallet configuration of the Zeitgeist pallets in the runtime, test and benchmark environment currently uses hard-coded values. If one value changes, for example *MaxAssets*, some tests and benchmarks become invalid, leading to invalid checks of correctness and potentially invalid automatic generation of weights.

Extend the primitives create, such that it contains the pallet configuration values of the Zeitgeist pallets. Update the runtime, tests and benchmarks to fetch the values from the primitives crate

PR #112, opened 2021-04-30
Bundle parameters

sea212:
Issue: #70

This PR moves the `parameter_types!` into the *primitives* crate. Now the runtime and the tests (to a certain degree) share the same parameters. For example, when changes to `MaxCategories` are applied, the benchmarks and tests in the prediction markets pallet will also use the new values automatically.
This PR also introduces structural changes: The *primitives* crates now has three public modules, *constants*, *types* and *traits*, through which the necessary values can be retrieved from the *primitives* crate.


c410-f3r (approved):
The fewer duplications, the better. Thanks for tackling this subject using a more ergometric and less error-prone approach.

lsaether (approved):
Looks great! Thanks!

Issue #89, opened 2021-04-21
sell_complete_set can corrupt the storage


sea212:
Issue: See title
https://github.com/zeitgeistpm/zeitgeist/blob/1355f47d8a3a792a7693f90de89548be20b99439/zrml/prediction-markets/src/lib.rs#L687-L696

Potential solutions: transact after every asset has been checked or use transactional attribute.

PR #117, opened 2021-05-04
Fix potential storage corruption in sell_complete_set

sea212:
Issue: #89 

Fix: Verify first, write last. This PR also updates `TransactionByteFee` from `1 * MILLICENTS` to `10 * MILLICENTS`, based on the example of Polkadot.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -96,7 +96,7 @@ parameter_types! {
   pub const MaxLocks: u32 = 50;
   pub const MinimumPeriod: u64 = SLOT_DURATION / 2;
   pub const SS58Prefix: u8 = 42; // @TODO: Change back to 73 once https://github.com/paritytech/substrate/pull/8509 is merged
-  pub const TransactionByteFee: Balance = 1 * MILLICENTS;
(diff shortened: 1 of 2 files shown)


c410-f3r (approved):
It would be nice to audit all runtime parameters before mainnet because some of them were simply copied and pasted from reference implementations.

Not sure if you guys talked about it but the addition of `transactional` for some PM call entry-points might make sense. For example:  `buy_complete_set`.

sea212:
> It would be nice to audit all runtime parameters before mainnet because some of them were simply copied and pasted from reference implementations.
> 
> Not sure if you guys talked about it but the addition of `transactional` for some PM call entry-points might make sense. For example: `buy_complete_set`.

I totally agree, we should definitely review and imo. also be able to proof why those values are optimal for mainnet launch.

Give me a couple of minutes, I'll update `buy_complete_set`

c410-f3r:
Thanks @sea212 !

lsaether (approved):
Thanks!

I would be quite worried if we are in a situation where `buy_complete_set` needs to revert using transactional since it would require inflation in ZTG to ever reach an amount of outcome assets that would trigger an error. But since we haven't formally proven this is the case, it's probably safer to add it on.

Issue #83, opened 2021-04-19
Some dispatchable functions allow arbitrarily sized vectors


sea212:
**Attack vector**
Some dispatchable functions take an arbitrarily sized vector as at least one of their arguments, that is not checked before inserting it into the storage database. Transactions (that invoke normal dispatches, i.e. "user" functions) can include about [3,75MB of arbitrary data](https://github.com/zeitgeistpm/zeitgeist/blob/68779e9edb5f1ba1eb3fb0a5def618d528593da0/runtime/src/lib.rs#L128) per block.
Since the [fee for 1 Byte](https://github.com/zeitgeistpm/zeitgeist/blob/68779e9edb5f1ba1eb3fb0a5def618d528593da0/runtime/src/lib.rs#L124) is one unit of currency and the weight calculation for storage insertions does not depend on the amount of data inserted, abusing this could lead to virtually free (in terms of transaction fees) storage population.

*Having the option to fill the block in that cheap way introduces the next vulnerability: DOS attacks. Because including such a transaction will cost only a fraction of the function call itself using the current configuration (~3,75mil fee for arbitrary data inclusion in contrast to 50+ mil weight for function execution and therefore [50+ mil fee](https://substrate.dev/rustdocs/v3.0.0/src/frame_support/weights.rs.html#662-733)), the blocks can be filled virtually for free and thus stall the execution of other extrinsics. This is covered in issue #86*.

*Note: The fee per byte issue and the possibility to insert arbitrarily sized data structures in transactions are two separate issues that should each be tackled non-exclusively. Having a reasonable fee per byte would raise the cost, but still allow to insert arbitrary amounts of data into the storage*.

**Vulnerable functions**
- [prediction_markets/create_categorical_market](https://github.com/zeitgeistpm/zeitgeist/blob/68779e9edb5f1ba1eb3fb0a5def618d528593da0/zrml/prediction-markets/src/lib.rs#L232)
- [prediction_markets/create_scalar_market](https://github.com/zeitgeistpm/zeitgeist/blob/68779e9edb5f1ba1eb3fb0a5def618d528593da0/zrml/prediction-markets/src/lib.rs#L283)

**Potential solutions**
- Introduce a maximum vector size that is checked during the execution of the function
- Use an array with a fixed size
[...]

Issue #119, opened 2021-05-05
Add proper docstrings to pallets


sea212:
**Issue**
Docstrings missing or incomplete for zrml-prediction-markets and zrml-orderbook-v1

PR #120, opened 2021-05-05
Define maximum metadata size

sea212:
Issue: #83 

**Solution**
This PR introduces a static length type, that represents a [multihash](multiformats.io/multihash/) (used in IPFS CIDv1) for the `metadata` in `create_categorical_market` and `create_scalar_market`. The changes have to be discussed, details follow below.
In addition, weight parameters for some functions have been changed. Before the maximum values were calculated wrong, which lead to overweight transactions. This has been fixed. If you are interested in the relatively accurately calculated system capacity, search for *0.75* in the discord *tech* channel.

**CIDv1 multiformat**
Data in IPFS is referenced using an old CIDv0 or CIDv1 multiformat. I suggest we apply the new CIDv1 standard.
According to [multiformat Github repository](https://github.com/multiformats/cid#how-does-it-work):
> \<cidv1> ::= \<mb>\<multicodec-cidv1>\<mc>\<mh>
> \# or, expanded:
> \<cidv1> ::= \<multibase-prefix>\<multicodec-cidv1>\<multicodec-content-type>\<multihash-content-address>

**Distinguishing which project does what**
Those are the elements, split into what different components of the zeitgeist project should do:
- multibase-prefix - appendend in SDK (depends on encoding)
- multicodec-cid - appendend in SDK (currently [CIDv1 0x01](https://github.com/multiformats/multicodec/blob/master/table.csv#L3))
- multicodec-content-type - appended in SDK (we can use [json 0x0200](https://github.com/multiformats/multicodec/blob/master/table.csv#L114) once it is ready, but this type is currently not supported and instead the [raw content type 0x85](https://github.com/multiformats/multicodec/blob/master/table.csv#L38) is used for json files)
- multihash-content-address - hash pointer to the actual data including information about the hash algorithm -> stored in chain storage

**Analysis of the used hash function for metadata IPFS references**
I suggest we use sha3-384 for ipfs hashes, which is already a [permanent multihash](https://github.com/multiformats/multicodec/blob/master/table.csv#L12). I have choosen sha3-384 over sha3-256, because according to [this paper](https://arxiv.org/pdf/quant-ph/9705002.pdf), the security in terms of collision resistance against quantum computers for sha3-256 could be only about 85 bits, which could realistically lead to collision attacks and consequently overshadow our market metadata in the near future. In contrast to sha3-256, sha3-384 still delivers 128 bit security against collision attacks, which is sufficient. I have decided against sha3-512 to save storage space in the long-term, the extra security is not required based on the current state of research. Using sha3-512 would requires 16 bytes more per market, in the very long term we want to save those if possible.

**Result**
This leads to a total requirement of
1 + 1 + 48 = 50 bytes [multihash](multiformats.io/multihash/)
for the market metadata within an extrinsic. 1 byte for the [sha3-384 multihash identifier (0x15)](https://github.com/multiformats/multicodec/blob/master/table.csv#L12), 1 byte for the length (0x30) and 48 bytes for the hash.

This format constraints the hash functions to hash functions which have an 1 byte identifier in the [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv). The constraints regarding the hash functions are just limitations in regards to the signature of the dispatchable function, internally our chain will store this data as arbitrary length byte vectors, effectively allowing us to upgrade the function signature to support any hash function, while at the same time avoiding arbitrary storage population due to the limitations in the metadata function parameter.

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -1,4 +1,5 @@
 [dependencies]
+arbitrary = { version = "1.0", default-features = false, optional = true }
 frame-support = { branch = "rococo-v1", default-features = false, git = "https://github.com/paritytech/substrate" }
 frame-system = { branch = "rococo-v1", default-features = false, git = "https://github.com/paritytech/substrate" }
 orml-currencies = { default-features = false, git = "https://github.com/open-web3-stack/open-runtime-module-library" }
@@ -12,6 +13,7 @@ sp-runtime = { branch = "rococo-v1", default-features = false, git = "https://gi
 [features]
 default = ["std"]
 std = [
+    "arbitrary",
     "frame-support/std",
     "frame-system/std",
     "orml-currencies/std",
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -13,7 +13,6 @@ sp-runtime = { branch = "rococo-v1", default-features = false, git = "https://gi
 [features]
 default = ["std"]
 std = [
-    "arbitrary",
     "frame-support/std",
     "frame-system/std",
     "orml-currencies/std",
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -92,7 +92,7 @@ mod pallet {
         traits::{Swaps, ZeitgeistMultiReservableCurrency},
         types::{
             Asset, Market, MarketCreation, MarketDispute, MarketEnd, MarketStatus, MarketType,
-            OutcomeReport, PoolId, PoolStatus, Report, ScalarPosition,
+            MultiHash, OutcomeReport, PoolId, PoolStatus, Report, ScalarPosition,
         },
     };
 
@@ -204,8 +204,8 @@ mod pallet {
         ////
         #[pallet::weight(T::WeightInfo::admin_move_market_to_resolved_overhead()
             .saturating_add(T::WeightInfo::internal_resolve_categorical_reported(
-                80_000,
-                80_000,
+                4_200,
+                4_200,
                 T::MaxCategories::get() as u32
             ).saturating_sub(T::WeightInfo::internal_resolve_scalar_reported())
         ))]
@@ -308,7 +308,7 @@ mod pallet {
             origin: OriginFor<T>,
             oracle: T::AccountId,
             end: MarketEnd<T::BlockNumber>,
-            metadata: Vec<u8>,
+            metadata: MultiHashSha384,
             creation: MarketCreation,
             categories: u16,
         ) -> DispatchResult {
@@ -322,6 +322,11 @@ mod pallet {
             ensure!(
                 categories <= T::MaxCategories::get(),
                 <Error<T>>::TooManyCategories
+            );
+            // Require sha3-384 as multihash.
+            ensure!(
+                metadata[0] == 0x15 && metadata[1] == 0x30,
+                <Error<T>>::InvalidMultihash
             );
 
             let status: MarketStatus = match creation {
@@ -338,13 +343,14 @@ mod pallet {
             };
 
             let market_id = Self::get_next_market_id()?;
+
             let market = Market {
                 creator: sender.clone(),
                 creation,
                 creator_fee: 0,
                 oracle,
                 end,
-                metadata,
+                metadata: Vec::from(metadata),
                 market_type: MarketType::Categorical(categories),
                 status,
                 report: None,
@@ -363,7 +369,7 @@ mod pallet {
             origin: OriginFor<T>,
             oracle: T::AccountId,
             end: MarketEnd<T::BlockNumber>,
-            metadata: Vec<u8>,
+            metadata: MultiHashSha384,
             creation: MarketCreation,
             outcome_range: (u128, u128),
         ) -> DispatchResult {
@@ -371,6 +377,11 @@ mod pallet {
             Self::ensure_create_market_end(end)?;
 
             ensure!(outcome_range.0 < outcome_range.1, "Invalid range provided.");
+            // Require sha3-384 as multihash.
+            ensure!(
+                metadata[0] == 0x15 && metadata[1] == 0x30,
+                <Error<T>>::InvalidMultihash
+            );
 
             let status: MarketStatus = match creation {
                 MarketCreation::Permissionless => {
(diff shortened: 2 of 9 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -328,6 +328,11 @@ mod pallet {
                 categories <= T::MaxCategories::get(),
                 <Error<T>>::TooManyCategories
             );
+            // Require sha3-384 as multihash.
+            ensure!(
+                metadata[0] == 0x15 && metadata[1] == 0x30,
+                <Error<T>>::InvalidMultihash
+            );
```
sea212, line 335: Our chain filters out any calls that contain metadata, which does not use the multihash identifier byte and length byte for sha3-384.

This ensures that all calls at least have the correct identifiers, potentially serving as a protection against invalid calls beforehand and therefore market creation with invalid metadata. It also enforces the hash function to sha3-384, denying any call that supplies metadata which does not use the corresponding identifiers. For anybody who just wants to play around locally, this may not be trivial, but this can be improved by adding nice docstrings (#119). Alternatively, we can increase the static size of the metadata array by 3 bytes and removing the checks, effectively allowing any hash function with up to 48 bytes digest size to be used.

What do you prefer? My position is we think about which data types we want once and then enforce it, which leads to uniformity and less error-prone usage (in case someone who uses it knows how it should be used - docs).

c410-f3r:
Need more time to review and think about the proposed changes

zrml/prediction-markets/src/lib.rs
lsaether: > For anybody who just wants to play around locally, this may not be trivial, but this can be improved by adding nice docstrings (#119).

A couple things we could do here,

1) make `Market.metadata` an `Option` so that it's possible to submit `None`. But if it is provided it must be a well-formatted multihash. 
2) make `Market.metadata` an new enum, that can extended over time:

```rust
enum Metadata {
  MultihashSha3_384([u8;50]),
  <new formats we want to support in the future>
  None,
}
```

lsaether: > My position is we think about which data types we want once and then enforce it, which leads to uniformity and less error-prone usage (in case someone who uses it knows how it should be used - docs).

I agree, if we do support new variants for metadata in the future, they should always be well specified and enforced in chain. In an ideal world, we would also enforce the JSON is correct with the expected fields. I see this possibly being the case if instead of adding to IPFS and submitting to Zeitgeist in two steps, we added the content to IPFS from the runtime, but I would leave this for a later improvement and pending further discussion.

lsaether (commented):
I see, so this PR keeps metadata as a `Vec`, which works nicely for backwards compatibility. Generally I would say that we should change the type of `Market.metadata` since we're only enforcing one type of data in there anyway.

lsaether:
To me this seems like a beneficial change and I'm all for it. The main complication is ensuring the downstream libraries like SDK and proto adapt to the change.

lsaether (approved):

sea212:
> I see, so this PR keeps metadata as a `Vec`, which works nicely for backwards compatibility. Generally I would say that we should change the type of `Market.metadata` since we're only enforcing one type of data in there anyway.

@lsaether I am not sure if we want this, because this only complicates the resolution in the SDK. Now the vector can be copied as-is in the SDK and be concatenated to the remaining information to create a CIDv1 string. If we enforce a specific type, the SDK has to distinguish between any different types. In addition, adding new types breaks the SDK temporarily. If it is a just a vec, the SDK does not have to worry about it - it always receives a valid multihash in the form of a (byte) vector and just has to concatenate it.

zrml/prediction-markets/src/lib.rs
sea212: @lsaether So no metadata at all is also a valid option?

lsaether:
I think in effect it would be delaying changes in the SDK rather than obviating the eventual need for them, but I see what you're saying. It's also beneficial that we wouldn't need to do a storage migration if we intend to deploy this as a runtime upgrade to the currently running BP testnet.

zrml/prediction-markets/src/lib.rs
lsaether: Yeah, I think it would be okay, the market would be pretty useless but it's at the cost of the creator. Also it would aid in testing to not need to have a local tool handy to create the multihash.

sea212:
Please don't merge yet, I'll add the new metadata enum.

c410-f3r:
`Multihash` is generic over several different hashes and `MultiHashSha384` might sound confusing: Is it Multihash, multihash over 384 bytes of data or the Sha384 version of a multihash? @lsaether's suggestion is really nice to distinguish between different hashes
```rust
enum Hash {
  Multihash(ArrayVec<u8, 512>),
  Sha256([u8; 256]),
  Multihash384([u8; 1 + 1+ 384]),
  ...
}

fn entrypoint(hash: Option<Hash>) { ... }
```
In any case, these are just nits and I don't have a strong opinion here

lsaether:
@sea212 Can you update the `types.json`?

@morkeltry This will need a change in the SDK after it gets merged.

sea212:
Have added a MultiHash enum (and implemented default `Arbitrary` trait for the fuzzer @c410-f3r) and updated the types.json @lsaether . Also a follow up issue was created that encompasses changing the storage layout to match the latest metadata type and adding a proper migration #148 .

@morkeltry I have created an issue in the tools project related to that, please check and resolve it: https://github.com/zeitgeistpm/tools/issues/89

primitives/Cargo.toml
```diff
@@ -12,6 +13,7 @@ sp-runtime = { branch = "rococo-v1", default-features = false, git = "https://gi
 [features]
 default = ["std"]
 std = [
+    "arbitrary",
```
c410-f3r, line 16: `arbitrary` is a development dependency that will be shipped on production builds. So, it is necessary to remove this line, add `#[cfg(feature = "arbitrary")]` to `primitives/src/types.rs` and then also add `zeitgeist-primitives = { features = ["arbitrary"], path = "../../../primitives" }` to the fuzz project.

It is also possible to manually construct `MultiHash` inside the fuzz project.

c410-f3r (approved):
Taking aside the below comment, everything else LGTM.

lsaether (approved):
All good from my side. Added the necessary changes in https://github.com/zeitgeistpm/tools/pull/90

PR #121, opened 2021-05-07
Add chainspecs and make the docker images publish on pushes to main

lsaether:
should be good to go if CI passes


node/src/cli.rs
```diff
@@ -143,9 +143,13 @@ pub fn load_spec(
             #[cfg(feature = "parachain")]
             para_id,
         )?),
-        "battery_park" => Box::new(crate::chain_spec::battery_park_config(
-            #[cfg(feature = "parachain")]
-            para_id,
```
c410-f3r, line 148: Rust-based specifications are useful for development and small tweaks although most (all?) parameters can also be modified directly in the `json` variant. For deployment, it would just be a matter of passing the canonical spec file `--chain path/to/file.json`.
Polkadot uses the same strategy for all runtimes. For example, Rococo -> https://github.com/paritytech/polkadot/blob/f9d5ad2a771319ec59265f434789b8080a442c9c/node/service/src/chain_spec.rs#L1000

In any case, this method is fine and will probably be sufficient

node/res/bp.json
```diff
@@ -0,0 +1,60 @@
+{
```
c410-f3r, line 1: Never deployed non-raw specs before and I am not sure if they will work. Probably will?

sea212 (commented):
When I build `cargo build --release` and run the chain `./target/release/zeitgeist --chain battery_park --tmp` I get the following error:
> Error: Service(Client(Storage("Error parsing embedded file: BadLength at line 31 column 13")))

node/res/bp_parachain.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "Zeitgeist Battery Park",
+  "id": "battery_park",
+  "chainType": "Live",
+  "bootNodes": [
+    "/ip4/139.162.171.58/tcp/30333/p2p/12D3KooWPvu5rpH2FNYnAmiQ8X8XqkMiuSFTjH2jwMCSjoam7RGQ"
+  ],
+  "telemetryEndpoints": [
+    [
+      "/dns/telemetry.polkadot.io/tcp/443/x-parity-wss/%2Fsubmit%2F",
```
sea212: Should this not be:
> wss://telemetry.polkadot.io/submit/

? In case of yes, it also has to be changed in node/res/bp.json

node/res/bp_parachain.json
```diff
@@ -0,0 +1,48 @@
...
...8cb22fab6227757905c45ec441c5c16fc8b64d6f160dc2c8c3445a4147c12b13f21913ca765709dab2c58b5567a8931b6b38d0da2164c776431ec29ab30d5db1513135c82429dcac0d89f55f2a7100fd4b2565beb02eb2a506a78d2196656e54c6d85d19e67430bf74749b3bac76c31fb9b45d9c3e34e203ef3d44db952eb451a786b9ebcc1ca9c3273a6e5ef6208e45d6e73fa86716d2e9a2ade5ab464cde655acd45c7e4d1801ed2d3626bab3ba05e3b41f2a29450865a40132f002991412938cb5af6ff8e93350a7074e39115f85f2288f44fe229c775174be5b832bb4b149636604b0f883b68a39b442fc044b4272b74b034560ad3c934a607eb017c1bb51509638366b61b3c029eaa03f3d1c3b66ce9d7358241f6e20ef73c842c6d48402df04c80b7916619f713d36121ec3c2cfcc8d862b03e2d6a80d0159cf138369c7a5bb811c26c8ce599ea5824eb97d92016897bd8ec3a912c3c0528e41a0038b90f8cd318ab4c16059f461803ee8419dfc0f2e3d54dbe4278dc449ee668fe6de2e0f9af103ce06baa5efe99d99c8fd65e23baa29b40857c890438a07bd5410c8f490482313a557b2a1b01f2eae6526043f933d3e5b606511ef29b85296423d082c1bf5f50e4f3d7b7129088a4392281a578fefb9e4f6d19089d59e27a3af9e38309d0f592f778bd7fe3474c69dded71857d1a1006f86e689dced68bd76988a56e2205e43ca6e0afa9d14289bd3289fe2f8135e93cedb9f99734ecd518bde28628a4a7933514ae6878154daf6ddf2b3865ad906b33a85b8f58f00cb3823b636f2ab3a5bcc92ddbf91897c0f767adca6337d28a6063ba239ee6
```
sea212: Where is the WASM blob (genesis @ 0x3a636f6465)?

I also think we have to use the raw format to specify genesis contents.

lsaether:
> When I build `cargo build --release` and run the chain `./target/release/zeitgeist --chain battery_park --tmp` I get the following error:
> 
> > Error: Service(Client(Storage("Error parsing embedded file: BadLength at line 31 column 13")))

thanks! this is because I got too excited putting the faucet address in the chainspec and pasted it twice https://github.com/zeitgeistpm/zeitgeist/pull/121/files#diff-aa880398e6fdc54607fb3a86c03a70d6fb3a86d2c7af4c649670a07626a573dfR30

node/res/bp_parachain.json
lsaether: Should be fine. It's a valid multiaddress and was also what was generated from the chainspec generation from the rust code. Will test it out though

node/res/bp.json
lsaether: Yep, they do

lsaether:
@c410-f3r I added Battery Park Staging back to the rust code we can continue to use it for development while we have the JSON spec for the public testnet.

c410-f3r (approved):
Thanks @lsaether

PR #122, opened 2021-05-09
Liquidity mining pallet

c410-f3r:
cc #78


c410-f3r:
With the latest master commit, this PR does not depend on #78 anymore

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+                    e
+                } else {
+                    // trying to retrieve an account that doesn't have any bought shares
+                    continue;
+                };
+
+                let sold_shares_pct = if let Some(el) = sold_shares.checked_mul(&_100_balance) {
+                    el
+                } else {
+                    // TODO: Should subtract some amount and delay a portion to the next block
```
sea212, line 221: Convert into issue and remove from code please.

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+                    el
+                } else {
+                    // `total_shares` is zero
+                    continue;
+                };
+
+                let total_balance_pct = if let Some(el) = values.total_balance.checked_mul(&pct) {
+                    el
+                } else {
+                    // TODO: Account has a lot of balance. What should be done here?
```
sea212, line 235: I think we can ignore that case for now, since nobody will ever reach such an enormous balance. If we reach this scenario, we would most likely have to worry about other things (exploits that were abused).

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+            for (account, pool_id, bought_shares) in remaining_bought_shares {
+                <BlockBoughtShares<T>>::insert(account, pool_id, bought_shares);
+            }
+        }
+
+        fn remove_incentives_based_on_sold_shares() {
+            let _100_balance = BalanceOf::<T>::from(100u8);
+            let block_sold_shares = <BlockSoldShares<T>>::iter().collect::<Vec<_>>();
+
+            for (account_id, pool_id, sold_shares) in block_sold_shares {
```
sea212, line 210: Can you explain to me please why this construct is necessary here? I mean why is 
```rust
for (account_id, pool_id, sold_shares) in <BlockSoldShares<T>>::iter() {
```
not sufficient? In the current case we iterate two times over every element if I understand it correctly.

sea212 (commented):
Thanks for implementing that feature Caio. Overall this looks good. I have left some comments, but there is nothing tragic. We have to work on some tests and many benchmarks & weights:

**missing tests (major)**
    liquidity-mining/distribute_incentives_based_on_bought_shares
    liquidity-mining/remove_incentives_based_on_sold_shares
    
**missing tests (minor)**
    liquidity-mining/add_shares_of_account_and_pool
    liquidity-mining/remove_shares_of_account_and_pool

**missing benchmarks**
    liquidity-mining/set_per_block_distribution
    
**missing weight calculations**
    liquidity-mining/set_per_block_distribution
    
**inspection of validity and possible adjustment of old benchmarks**
    swaps/pool_exit
    swaps/pool_exit_with_exact_asset_amount
    swaps/pool_join
    swaps/pool_join_with_exact_asset_amount
    swaps/swap_exact_amount_in
    swaps/swap_exact_amount_out
    
**update of weights**
    swaps/pool_exit
    swaps/pool_exit_with_exact_asset_amount
    swaps/pool_join
    swaps/pool_join_with_exact_asset_amount
    swaps/swap_exact_amount_in
    swaps/swap_exact_amount_out

node/src/chain_spec/mod.rs
```diff
@@ -62,6 +68,13 @@ fn generic_genesis(
         parachain_info: zeitgeist_runtime::ParachainInfoConfig {
             parachain_id: acs.parachain_id,
         },
+        zrml_liquidity_mining: {
+            let perbill = Perbill::from_percent(LIQUIDITY_MINING_PCT);
```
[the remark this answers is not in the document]
c410-f3r: Nice catch! This probably came from previous development iterations

zrml/liquidity-mining/src/liquidity_mining.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait LiquidityMining<AI, B> {
+    /// Returns `Err` if `shares` plus the current amount of shares overflows.
+    fn add_shares_of_account_and_pool(account_id: &AI, pool_id: u128, shares: B) -> DispatchResult;
+
+    /// Removes a given `pool_id` from the storage distributing incentives to all
+    /// related accounts.
+    fn remove_pool_distributing_incentives(pool_id: u128) -> DispatchResult;
+
+    /// Ignores if `shares` is greater than the stored amount of shares.
+    fn remove_shares_of_account_and_pool(account_id: &AI, pool_id: u128, shares: B);
```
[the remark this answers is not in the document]
c410-f3r: Changed. The remaining `u128` should be handled by #127

zrml/liquidity-mining/src/lib.rs
c410-f3r: My bad, it doesn't. `vec` conversation only matters when calculating bought shares.

zrml/liquidity-mining/src/lib.rs
c410-f3r: Yeah, you are right but it still concerts me (theoretically).
Reverted to saturating operations and added some documentation describing the situation

c410-f3r:
Thanks for the review, @sea212

> Thanks for implementing that feature Caio. Overall this looks good. I have left some comments, but there is nothing tragic. We have to work on some tests and many benchmarks & weights:
> 
> **missing tests (major)**
> liquidity-mining/distribute_incentives_based_on_bought_shares
> liquidity-mining/remove_incentives_based_on_sold_shares
> 
> **missing tests (minor)**
> liquidity-mining/add_shares_of_account_and_pool
> liquidity-mining/remove_shares_of_account_and_pool
> 
> **missing benchmarks**
> liquidity-mining/set_per_block_distribution
> 
> **missing weight calculations**
> liquidity-mining/set_per_block_distribution
> 
> **inspection of validity and possible adjustment of old benchmarks**
> swaps/pool_exit
> swaps/pool_exit_with_exact_asset_amount
> swaps/pool_join
> swaps/pool_join_with_exact_asset_amount
> swaps/swap_exact_amount_in
> swaps/swap_exact_amount_out
> 
> **update of weights**
> swaps/pool_exit
> swaps/pool_exit_with_exact_asset_amount
> swaps/pool_join
> swaps/pool_join_with_exact_asset_amount
> swaps/swap_exact_amount_in
> swaps/swap_exact_amount_out


Like I said, sorry for not including benchmarks. They are now present.

`distribute_incentives_based_on_bought_shares` and `remove_incentives_based_on_sold_shares` were already tested when calling `on_finalize` and now they are explicitly called to better express the test's intention.

Changed `SomeStorage::insert` to `add_shares_of_account_and_pool/remove_shares_of_account_and_pool` since they are very similar.

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+        type Currency: ReservableCurrency<Self::AccountId>;
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+        type PalletId: Get<PalletId>;
+        type WeightInfo: WeightInfoZeitgeist;
+    }
+
+    #[pallet::event]
+    pub enum Event<T>
+    where
+        T: Config, {}
```
c410-f3r, line 70: An issue will be open to add useful events

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+
+        #[frame_support::transactional]
+        fn remove_pool_distributing_incentives(pool_id: PoolId) -> DispatchResult {
+            let pallet_account_id = Pallet::<T>::pallet_account_id();
+            for (account_id, values) in <OwnedValues<T>>::drain_prefix(pool_id) {
+                T::Currency::transfer(
+                    &pallet_account_id,
+                    &account_id,
+                    values.total_balance,
+                    ExistenceRequirement::KeepAlive,
```
lsaether, line 245: `AllowDeath` should be fine here.

lsaether (changes requested):
Ok, so to make sure I understand the current functionality:

- When a user joins a pool, they trigger the `add_shares_of_account_and_pool` which at the end of the block will record the `PoolShare`s that they have in storage.
- When a user exits a pool, they trigger `remove_shares_of_account_and_pool` which will remove them from consideration of receiving rewards _and does not pay any LM rewards_.
- In order to pay out rewards to LPs, the `remove_pool_distributing_incentives` must be called. This can be triggered when a market resolves, but that part is  not yet implemented.

So current behavior is that __when a LP exits a pool early, they do not receive rewards__, right? We will probably want to change this so that they do receive some rewards, but not as much as they would have received if they had stated for the entire length of the pool. Along these lines, we need to also make sure that LPs which join earlier are rewarded more than LPs join later, since AFAIK right now it would be considered equal rewards if LPs joined at the beginning of the market or one block before the end of the market.

c410-f3r:
> In order to pay out rewards to LPs, the remove_pool_distributing_incentives must be called. This can be triggered when a market resolves, but that part is not yet implemented.

Indeed!

> So current behavior is that when a LP exits a pool early, they do not receive rewards, right?
Yes. If LP removes all shares, then nothing will be provided when market resolves

> AFAIK right now it would be considered equal rewards if LPs joined at the beginning of the market or one block before the end of the market.
Yes, all shares have the same percentage and value according to the number of total shares of a block.

> Along these lines, we need to also make sure that LPs which join earlier are rewarded more than LPs join later
A logarithm function will do the job.

> We will probably want to change this so that they do receive some rewards, but not as much as they would have received if they had stated for the entire length of the pool
This is odd... Anyone with a significant amount of ZTG could abuse this feature, providing and un-provinding liquidity, to create more ZTG out of thin air.

lsaether:
> This is odd... Anyone with a significant amount of ZTG could abuse this feature, providing and un-provinding liquidity, to create more ZTG out of thin air.

It should be something like continuous payments, like every block the LP gets rewarded with X amount of ZTG. If the LP stays from 1000 blocks, then they would get rewarded 1000*X, but if the LP just joined for one block they get 1*X - `early_exit_fee`. Then at the end of the market, the summed `early_exit_fees` get redistributed to all LPs that stayed through until the end.

c410-f3r:
Took me awhile to finish because I had to re-think and re-evaluate incentives. 

## Providing liquidity

When an account is providing liquidity, the given incentive will be stabilised by a framework that can be adapted with different types of algorithms. This framework is composed of two steps:

1. `Market normalization`: Each market will output a number of equal range. How this output is produced is up to the chosen algorithm.
2. `Percentage calculation`: Based on the normalization values, each market will have an associated percentage summing a total of 100%.

To encourage early providers, an algorithm that takes the remaining blocks as percentages is currently being used so let's see an example with two markets:

```rust
// Amount of reserved ZTG for the current block
let block_incentives = 1000;
// Block that is being produced
let current_block = 5;

// Start and end blocks of market A
let market_a_period = [1, 11];
// Start and end blocks of market B
let market_b_period = [1, 21];

let market_a_total_blocks = 11 - 1 = 10;
let market_a_remaining_blocks = 11 - 5 = 6
let market_b_total_blocks = 21 - 1 = 20;
let market_b_remaining_blocks = 21 - 5 = 16

let market_a_normalized_value = 6 * 100 / 10 = 60
let market_b_normalized_value = 16 * 100 / 20 = 80

let market_a_percentage = 60 * 100 / 140 = 42,857%
let market_b_percentage = 80 * 100 / 140 = 57.143%

let market_a_incentives = 1000 * 42,857% = 428.57
let market_b_incentives = 1000 * 57.143% = 571.43
```

As you can see, Market `A` receives less incentives because its completion will happen before market `B`.

## Perpetual liquidity

Is an incentive that each LP will receive by simply joining a pool. Currently, it is calculated based on a pre-fixed percentage of 0.5% applied to the amount of received incentives in a block.
At the end of the market, accounts will receive the full amount of perpetual liquidity if they stayed the whole campaign or its relative percentage otherwise.
```txt
User A receives 1 ZTG as incentive -----> 0.995 ZTG goes to ordinary balance
                                   |
                                   |----> 0.005 ZTG goes to perpetual balance
```

primitives/src/pool.rs
```diff
@@ -11,6 +11,7 @@ use alloc::{collections::BTreeMap, vec::Vec};
 )]
 pub struct Pool<Balance, MarketId> {
     pub assets: Vec<Asset<MarketId>>,
+    pub market_id: MarketId,
```
c410-f3r, line 14: Temporary. The market will be fetched on `market-commons` once  #167 is merged

zrml/prediction-markets/src/lib.rs
```diff
@@ -362,6 +360,12 @@ mod pallet {
             };
 
             <Markets<T>>::insert(market_id, Some(market));
+            if let MarketEnd::Block(block) = end {
```
c410-f3r, line 363: Is it possible to remove `MarketEnd::Time`?

c410-f3r:
This PR was created before the discussion about big PRs so do you guys want me to split it into different and smaller PRs?

sea212:
> This PR was created before the discussion about big PRs so do you guys want me to split it into different and smaller PRs?

I am fine with reviewing this last big PR.

primitives/src/constants/ztg.rs
```diff
@@ -9,6 +9,9 @@ pub const COLLATORS: u128 = 10_000_000;
 /// Total ZTG amount for community incentives
 pub const COMMUNITY_INCENTIVES: u128 = 2_000_000;
 
+/// Total ZTG amount for liquidity mining
+pub const LIQUIDITY_MINING: u128 = 10_000_000;
```
lsaether, line 13: Is this used anywhere? If so, how is it used?

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,106 @@
...
+        incentive.checked_div(&total_bought_shares)
+    }
+
+    // How much incentive each market will receive
+    fn markets_incentives(
+        per_block_incentives: BalanceOf<T>,
+        curr_block: T::BlockNumber,
+    ) -> Vec<(T::MarketId, BalanceOf<T>)> {
+        let mut normalized_total = BalanceOf::<T>::from(0u8);
+        let normalized_valeus: Vec<_> = <Markets<T>>::iter()
```
lsaether, line 74:
```suggestion
        let normalized_values: Vec<_> = <Markets<T>>::iter()
```

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,106 @@
...
+            .filter_map(|(market_id, period)| {
+                let normalized_value = Self::normalize_market(curr_block, period)?;
+                normalized_total = normalized_total.saturating_add(normalized_value);
+                Some((market_id, normalized_value))
+            })
+            .collect();
+        normalized_valeus
```
lsaether, line 81:
```suggestion
        normalized_values
```

primitives/src/constants/ztg.rs
c410-f3r: This constant is the total amount of ZTG reserved for the liquidity mining pallet account or in other words, a fund/pot where incentives will be transferred to accounts.

It is defined on the genesis block at https://github.com/zeitgeistpm/zeitgeist/pull/122/files#diff-d0ab461d3f1322691c297bdf74c56e023de548c8555915ed6b60b0a792129341R130-R136

lsaether: Is the pot funded any other ways, i.e. via inflation? This I suppose can be a future addition

c410-f3r: For now it is a fixed amount that can be changed using a root account but yeah, inflation could be added in a following PR.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1379,6 +1392,8 @@ mod pallet {
                 T::Swap::set_pool_as_stale(&market.market_type, pool_id, outcome_report)?;
             }
 
+            T::LiquidityMining::distribute_market_incentives(market_id)?;
```
sea212, line 1395: I see that this function is called in courts resolve and prediction markets resolve functions. Is there any way that this function could be called twice for one market?

zrml/swaps/src/lib.rs
```diff
@@ -398,6 +429,9 @@ mod pallet {
                     )?.saturated_into();
                     ensure!(asset_amount_out >= min_asset_amount_out, Error::<T>::LimitOut);
 
+                    T::LiquidityMining::remove_shares(&who, &pool.market_id, asset_amount_out);
+                    T::LiquidityMining::add_shares(who_clone, pool.market_id, asset_amount_in);
```
sea212, line 433: I am a bit confused here. So everyone who trades Asset A OUT for Asset B IN, automatically becomes a liquidity provider and earns amount IN shares? `add_shares` and `remove_shares` in liquidity provider pallet also need some documentation. What are those shares? LP shares? Assets?

zrml/swaps/src/migrations/_0_1_2_add_pool_market_id.rs
```diff
@@ -0,0 +1,110 @@
...
+    let storage_version = <Pallet<T>>::storage_version().unwrap_or(pallet_version);
+
+    if storage_version == pallet_version {
+        let _ = <Pools<T>>::translate::<Option<PreviousPoolTy<T>>, _>(|_, pool_opt| {
+            weight = weight.saturating_add(T::DbWeight::get().reads(1));
+            if let Some(pool) = pool_opt {
+                weight = weight.saturating_add(T::DbWeight::get().writes(1));
+                Some(Some(Pool {
+                    assets: pool.assets,
+                    market_id: T::MarketId::from(0u8),
```
sea212, line 46: The old pools will have an invalid associated `market_id`, what does this imply?

sea212 (changes requested):

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        <BlockBoughtShares<T>>::remove_all();
+    }
+
+    // ZTG value of one bought share for the current block being produced. Or in other words:
+    // Determines how much a share will be worth given the amount of ZTG for liquidity
+    // mining and the total number of bought shares for the current block.
+    //
+    // `None` result means no-one purchased a share.
+    fn bought_share_value(
```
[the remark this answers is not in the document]
c410-f3r: Agreed. This talk of "the compiler is smart enough to inline" is not always true.

node/src/chain_spec/mod.rs
```diff
@@ -25,7 +30,7 @@ const DEFAULT_COLLATOR_INFLATION_INFO: parachain_staking::InflationInfo<Balance>
     let round_millisecs = DefaultBlocksPerRound::get() as u64 * MILLISECS_PER_BLOCK;
     let rounds_per_year = millisecs_per_year / round_millisecs;
 
-    let annual_inflation = Perbill::from_percent(ztg::STAKING);
+    let annual_inflation = Perbill::from_percent(ztg::STAKING_PCT);
```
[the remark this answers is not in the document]
c410-f3r: Jesus Christ. Another good candidate from a integration test

zrml/court/src/lib.rs
```diff
@@ -521,13 +521,15 @@ mod pallet {
...
+        fn manage_pool(
             market: &Market<T::AccountId, T::BlockNumber>,
             market_id: &T::MarketId,
             outcome_report: &OutcomeReport,
         ) -> DispatchResult {
             let pool_id = Self::market_pool_id(market_id)?;
 
+            T::LiquidityMining::distribute_market_incentives(market_id)?;
+
             T::Swaps::set_pool_as_stale(&market.market_type, pool_id, outcome_report)?;
```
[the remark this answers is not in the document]
c410-f3r: Yes because I couldn't find a good motive to guard against non-active pools.
In any case, it is better to play safe so the guard is now added.

zrml/prediction-markets/src/lib.rs
c410-f3r: `T::LiquidityMining::distribute_market_incentives(market_id)?;` must be called once for every market resolution and this constraint will be enforced when #167 is merged

zrml/swaps/src/migrations/_0_1_2_add_pool_market_id.rs
c410-f3r: The Swaps pallet is not aware of anything regarding markets, in fact, PM uses Swaps and not the other way around so there is no place to take market information from Swaps. This migration is just a way to enforce backwards compatibility.

Once #167 is merged, common logic between markets, pools, liquidity mining and court will be used inside `market-commons` so another migration must be created afterwards to correctly map swaps <-> migrations

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,106 @@
...
+    // early liquidity providers. For example, a market starts at 0 and ends at 10. If the current
+    // block is at 7, then the output is 30.
+    fn normalize_market(
+        curr_block: T::BlockNumber,
+        [start, end]: [T::BlockNumber; 2],
+    ) -> Option<BalanceOf<T>> {
+        let total_blocks = end.saturating_sub(start);
+        let remaining_blocks = end.saturating_sub(curr_block);
+        let ptd = calculate_perthousand(remaining_blocks, &total_blocks)?;
+        Some(perthousand_to_balance(ptd))
```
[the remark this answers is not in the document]
c410-f3r: The final result is actually `300` units (30%/300ptd of 10) instead of `3000` units (300%/3000ptd of 10).

But yes, it is per-thousand because percent incurs a greater lost of accuracy. For example, 
https://github.com/zeitgeistpm/zeitgeist/pull/122/files#diff-6d701ba4fa5e808263982c3ef607ad265031bb560ca867c28423f302da40faf4R152-R153
would be 
```rust
        // Market 0 has 42% of `per_block_incentives`
        let market_incentives_0 = 420000000;
```
with percentages.

> Is it possible that remaining_blocks can become 0 (current = end)? In that case we'd have a zero division error.

All per-thousands are calculated through the unified `calculate_perthousand` function that guards against divisions by zero -> https://github.com/zeitgeistpm/zeitgeist/pull/122/files#diff-2fa2c31f9a820cbd97164a96acb2d7648da09ec2226e576c034554d1b1b8e1cbR8-R16.

zrml/liquidity-mining/src/owned_values_params.rs
```diff
@@ -0,0 +1,29 @@
...
+    Debug,
+    Default,
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    parity_scale_codec::Decode,
+    parity_scale_codec::Encode,
+)]
+pub struct OwnedValuesParams<BAL, BNR> {
```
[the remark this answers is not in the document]
c410-f3r: With `OwnedValuesParams<BalanceOf<T>, T::BlockNumber>`, the compiler already knowns that both `BAL` and `BNR` implement `CheckedDiv + From<u16> + Saturating` -> https://github.com/zeitgeistpm/zeitgeist/pull/122/files#diff-be1627b38629a5ebacab4e35ecebea6efc65cc628067a7e9f77f5a60458cff78R242-R250

zrml/swaps/src/lib.rs
c410-f3r: My bad. These functions receive pool shares instead of asset amounts.
Another reason to create smaller PRs: They are also easier to review and less error-prone for the author

c410-f3r: Wait a second. `asset_amount_in` should be `shares_amount_in`...

lsaether: No, only adding and removing liquidity is considered for liquidity mining.

c410-f3r: Thanks guys. Just removed liquidity functions from swapping operations

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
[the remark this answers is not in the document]
sea212: If the remaining blocks can become zero (current = end), we should return 0% instead of a zero division error.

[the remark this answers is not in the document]
c410-f3r: Oh, I see now. Sorry for the misunderstanding and thanks for spotting this behaviour

zrml/swaps/src/migrations.rs
```diff
@@ -1,2 +1 @@
-pub mod _0_1_1_add_pool_status;
-pub mod _0_1_2_add_pool_market_id;
```
c410-f3r: Ppl on https://docs.zeitgeist.pm/battery-park#get-connected are still using the same initial fb127223ea8990bb27819dbbb9b15a46d7ffea73 build and no new release was published since then. Therefore, these two migrations are now merged into one.

lsaether:
@c410-f3r Can you change it back to make the migration on 0_1_2? I want to make an upgrade to 0_1_1 before we will merge this in along with the court and other more significant changes.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1366,8 +1378,9 @@ mod pallet {
             Ok([total_accounts, total_asset_accounts, total_categories])
         }
 
-        // If a market has a pool that is `Active`, then changes from `Active` to `Stale`.
-        fn manage_pool_staleness(
+        // * Distributes rewards to liquidity providers
+        // * If a market has a pool that is `Active`, then changes from `Active` to `Stale`.
+        fn manage_pool(
```
lsaether, line 1383:
```suggestion
        fn change_pool_to_stale(
```

zrml/prediction-markets/src/lib.rs
lsaether, line 1395: `distribute_market_incentives` is marked transactional but the calling function is not, which means that a pool can still be changed to stale without distributing rewards. We should probably make the calling function transactional as well.

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+use alloc::vec::Vec;
+use core::marker::PhantomData;
+use sp_runtime::traits::{CheckedDiv, Saturating};
+
+// Per-thousand that every incentive should transfer to the perpetual balance.
+// Currently is 0.1%
+const PERPETUAL_PCT: u8 = 1;
+
+pub struct DistributeIncentivesBasedOnBoughtShares<T>(PhantomData<T>);
```
lsaether, line 13: IMO we should rename this since it does not distribute the incentives (that's done later) but it adds incentive tracking data to storage. I would suggest `TrackIncentivesBasedOnBoughtShares`

zrml/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+            market_id: Self::MarketId,
+            shares: Self::Balance,
+        ) {
+            <BlockBoughtShares<T>>::mutate(market_id, account_id, |total_shares| {
+                *total_shares = total_shares.saturating_add(shares);
+            })
+        }
+
+        #[frame_support::transactional]
+        fn distribute_market_incentives(market_id: &Self::MarketId) -> DispatchResult {
```
lsaether, line 160: This needs to check that the pallet account has funds in it to distribute rewards. The pool could be zero (which it will likely start out with when we first deploy this to the testnet.

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
```diff
@@ -0,0 +1,108 @@
...
+            })
+            .collect()
+    }
+
+    // Takes any market parameter and outputs a number that will be used as a percentage
+    // to calculate how much incentives each individual market will receive.
+    //
+    // In this case, the output is the percentage of the remaining number of blocks to stimulate
+    // early liquidity providers. For example, a market starts at 0 and ends at 10. If the current
+    // block is at 7, then the output is 30.
```
lsaether, line 97: Can you give some additional examples of this? Not sure I totally understand how this will differ the payout to LPs that enter at different blocks.

zrml/prediction-markets/src/lib.rs
c410-f3r: Added `#[frame_support::transactional]`

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
c410-f3r: Changed `DistributeIncentivesBasedOnBoughtShares` to `TrackIncentivesBasedOnBoughtShares` and `RemoveIncentivesBasedOnSoldShares` to `TrackIncentivesBasedOnSoldShares`

zrml/liquidity-mining/src/lib.rs
c410-f3r: Added `ensure_can_withdrawn` but `#[frame_support::transactional]` should theoretically take care of this scenario

zrml/prediction-markets/src/lib.rs
c410-f3r: Changed

zrml/liquidity-mining/src/distribute_incentives_based_on_bought_shares.rs
c410-f3r: Added one more example. Please, feel free to discuss any doubts about the algorithm

zrml/liquidity-mining/src/lib.rs
lsaether: Ok, can we return early or some other way handle the case that there are no funds to make the payout of the rewards? Since when we first deploy this there will be no funds being used to pay out rewards, now this will prevent a pool from being move to stale

c410-f3r: `ensure_can_withdraw` will return an early error if the fund does not have enough funds. Should the `ensure_can_withdraw` check return an early `Ok` instead?
It is also possible to ignore `distribute_market_incentives` within `set_pool_to_stale` using `let _ = T::LiquidityMining::distribute_market_incentives(market_id);` instead of `T::LiquidityMining::distribute_market_incentives(market_id)?;`

lsaether: Yes, it should return an early `Ok` and also log an event `LiquidityRewardsNotDistributed` so we know when it happens. The behaviour should be either everyone gets distributed rewards or no one is distributed rewards, but either way the pool still moves to `Stale`.

c410-f3r: Done

As stated by https://github.com/zeitgeistpm/zeitgeist/pull/122#discussion_r636214190, I would like to create (if possible) events to this pallet in a following PR because the changes are already too big.

lsaether: Which issue did you create?

lsaether: #197

lsaether (approved):

c410-f3r:
Re-based from master and opened https://github.com/zeitgeistpm/zeitgeist/issues/208

lsaether:
I'll merge this in after the next release (should be later today).

c410-f3r:
Rebased

PR #123, opened 2021-05-10
Add pool_exit test

lsaether:


lsaether:
Ok... so tests pass locally but fail on CI. #progress

lsaether:
Issue is not related to internal functionality of the Swaps pallet like I first suspected -- but instead to an overlooked interaction between PredictionMarkets and Swap. Being tracked in #125

Issue #125, opened 2021-05-10
Ensure all liquidity has been withdrawn from pools before destroying assets in market resolution


lsaether:
If [an asset has a balance of 0](https://github.com/zeitgeistpm/zeitgeist/blob/4217f3845caa7664192469748e57a34dcc9e946b/zrml/swaps/src/macros.rs#L119-L127) it will cause the `remove_liquidity` function to fail, making the liquidity stuck in the pool forever. We should then ensure that liquidity is removed from pools before market resolution happens, or alternatively not destroy the assets during market resolution.

This could be resolved by transforming a `Pool` into a `StalePool` when market resolution starts. There is not trading allowed on `StalePool` and it's just a bag of assets that's waiting to be claimed back by the LPs.

PR #127, opened 2021-05-11
Add `PoolStatus` parameter

c410-f3r:


zrml/prediction-markets/src/lib.rs
```diff
@@ -559,6 +559,13 @@ mod pallet {
         ) -> DispatchResultWithPostInfo {
             let sender = ensure_signed(origin)?;
 
+            let mapped_pool_id = if let Ok(Some(el)) = <MarketToSwapPool<T>>::try_get(market_id) {
+                el
+            } else {
+                return Err(Error::<T>::PoolDoesNotExist.into());
+            };
+            T::Swap::set_pool_status(mapped_pool_id, false)?;
+
```
lsaether, line 568: This is not the right place for this code, it should trigger before someone calls `redeem_shares`. Ideally it's called in the `on_initilaize` hook when a market ends, but we can also trigger it when a market is reported too to avoid having to do another iteration in the `on_initialize` hook

c410-f3r:
Ok, took me a while to figure out what needed to be done. Everything should be finished tomorrow

primitives/src/zeitgeist_multi_reservable_currency.rs
```diff
@@ -21,24 +21,16 @@ where
...
-                    }
-
-                    if k1 == currency_id {
-                        Some((k0, v))
-                    } else {
-                        None
-                    }
-                })
-                .collect(),
-        )
```
c410-f3r, line 41: IINM, `(u32, Vec<(T::AccountId, AccountData<T::Balance>)>)` here is the same of
```
let rslt = accounts_by_currency_id(something);
(rslt.len(), rslt)
```

primitives/src/zeitgeist_currencies_extension.rs
```diff
@@ -1,19 +0,0 @@
...
-where
-    Self::MultiCurrency: ZeitgeistMultiReservableCurrency<Self::AccountId>,
-{
-    /// Destroy all storage items for a `currency_id`.
-    fn destroy_all(currency_id: CurrencyIdOf<Self>) -> DispatchResult {
-        let (_, accounts) = Self::MultiCurrency::accounts_by_currency_id(currency_id);
-        Self::MultiCurrency::destroy_all(currency_id, accounts.iter().cloned());
-        Ok(())
-    }
-}
```
c410-f3r, line 19: Replaced by the additional `ZeitgeistMultiReservableCurrency` implementation

zrml/prediction-markets/src/lib.rs
```diff
@@ -1167,11 +1172,49 @@ mod pallet {
         pub(crate) fn internal_resolve(market_id: &T::MarketId) -> Result<Weight, DispatchError> {
             let market = Self::market_by_id(market_id)?;
             let report = market.report.clone().ok_or_else(|| NO_REPORT)?;
-            let mut total_accounts = 0u32;
-            let mut total_asset_accounts = 0u32;
-            let mut total_categories = 0u32;
+            let total_accounts = 0u32;
+            let total_asset_accounts = 0u32;
+            let total_categories = 0u32;
```
c410-f3r, line 1177: @sea212 Liquidity removal must happen before market resolution so these parameters should be removed or transferred to other entry-point. Any suggestions?

zrml/prediction-markets/src/lib.rs
```diff
@@ -1259,28 +1290,6 @@ mod pallet {
...
-                                total_accounts = total_accs;
-                            }
-
-                            total_asset_accounts = total_asset_accounts
-                                .saturating_add(accounts.len().saturated_into());
-                            T::Shares::destroy_all(*asset, accounts.iter().cloned());
-                        }
-                    }
-                }
-            }
```
c410-f3r, line 1283: Removal of loosing categorical assets now happens when calling the `Swap::set_pool_as_stale` function

c410-f3r (commented):
On PM, a new active pool is created when deployed, is marked as `Stale` when **reported** (not sure how it will interact with disputes) and finally modified to `Inactive` on `internal_resolve` when liquidity is withdrawn.
We should add a timer because it is not guaranteed that **all** liquidity will be removed by **all** accounts.

primitives/src/zeitgeist_multi_reservable_currency.rs
sea212: This is not the same, since the `u32` that was returned before was the number of total accounts and `rslt.len()` is the number of accounts that hold a specific `CurrencyId`. Please revert.

primitives/src/swaps.rs
```diff
@@ -1,15 +1,40 @@
...
+    /// * `weights`: These are the denormalized weights (the raw weights).
     fn create_pool(
         creator: AccountId,
         assets: Vec<Asset<Self::MarketId>>,
         swap_fee: Self::Balance,
         weights: Vec<u128>,
     ) -> Result<u128, DispatchError>;
+
+    /// Returns the assets of a pool.
+    fn pool_assets(pool_id: u128) -> Result<Vec<Asset<Self::MarketId>>, DispatchError>;
```
sea212, line 26: Maybe we can create a type alias `PoolId` for `u128` like we have for market `MarketId`. Not sure if we ever change this, but u64 is likely sufficient and most 64bit processors have instructions to directly manipulate them in one register, therefore being more efficient in handling them.

zrml/swaps/src/lib.rs
```diff
@@ -733,5 +739,56 @@ mod pallet {
...
+                }
+                for asset in assets_iter {
+                    let accounts_len = manage_asset(asset, winning_asset_idx);
+                    total_asset_accounts = total_asset_accounts.saturating_add(accounts_len);
+                }
+            }
+
+            Self::set_pool_status(pool_id, PoolStatus::Stale)?;
+            Ok([total_accounts, total_asset_accounts])
+        }
```
sea212, line 792: The snippet delivers a wrong number: `total_accounts` really carries the number of accounts for `first_asset`, not the total number of accounts.

zrml/prediction-markets/src/lib.rs
sea212: We have to transfer this to the function that removes liquidity. As I see it currently, the function `report` does it now. We have to reflect the changes in the benchmarks:
1) Think about which logical paths the functions `report` and `internal_resolve` take now on from which variables they depend
2) Update benchmarks
3) Update `report` and `internal_resolve` to return overcharged weight
4) Benchmark new weights and include into `weight.rs`

I can handle the 4th step after you're done, so all the weights were consistently calculated on the same hardware. We'll implement an automatic solution for that soon.

sea212 (changes requested):
Thanks Ciao for tackling that issue. I think some changes are required, mainly:
1) Investigate into invalid total_accounts number
2) Add updated benchmarks for changed functions
3) Add updated weight return calculation for changed functions
4) Fix tests: `cargo test --features std,runtime-benchmarks --manifest-path zrml/prediction-markets/Cargo.toml --no-default-features`:
```
running 19 tests
test mock::__construct_runtime_integrity_test::runtime_integrity_tests ... ok
test tests::it_allows_the_advisory_origin_to_reject_markets ... ok
test tests::it_allows_to_buy_a_complete_set ... ok
test tests::it_allows_to_sell_a_complete_set ... ok
test tests::it_allows_to_deploy_a_pool ... ok
test tests::it_creates_binary_markets ... ok
test tests::it_allows_sudo_to_destroy_markets ... ok
test tests::it_does_not_create_market_with_too_many_categories ... ok
test tests::it_does_not_create_market_with_too_few_categories ... ok
test tests::it_allows_advisory_origin_to_approve_markets ... ok
test tests::it_allows_to_report_the_outcome_of_a_market ... FAILED
test tests::full_scalar_market_lifecycle ... FAILED
test tests::it_allows_to_redeem_shares ... FAILED
test tests::the_entire_market_lifecycle_works_with_timestamps ... FAILED
test tests::it_correctly_resolves_a_market_that_was_reported_on ... FAILED
test tests::it_resolves_a_disputed_market ... FAILED
test tests::it_allows_to_dispute_the_outcome_of_a_market ... FAILED
test benchmarks::benchmark_tests::test_benchmarks ... FAILED
test tests::it_allows_anyone_to_report_an_unreported_market ... FAILED
```

primitives/src/zeitgeist_multi_reservable_currency.rs
c410-f3r: OK, in any case `total` was and always will be zero so is this parameter really necessary?
If so, I will fix this snippet to return the total number of accounts.

sea212: @c410-f3r What a bummer! In that case can you restructure the code that it takes the actual total number after `<Accounts<T>>::iter().filter_map` was evaluated?
```
        let mut total: u32 = 0;
        let accounts = <Accounts<T>>::iter()
        .filter_map(|(k0, k1, v)| {
            if total < u32::MAX {
                total += 1;
            }

            if k1 == currency_id {
                Some((k0, v))
            } else {
                None
            }
        })
        .collect();
        (total, accounts)
```

sea212: I don't like this whole construct but we need to count the number of total accounts currently to properly calculate the weight. Unfortunately getting the number of total accounts does require iterating over every account (to my knowledge), so I have decided to return this information from this function to avoid one additional iteration over every account.

c410-f3r: Oh, don't worry. I was concerned about possible unintentional leftovers but this snippet is indeed used and expected in other places.
Thanks for the review 👍. Tomorrow everything will probably be fixed.

primitives/src/swaps.rs
c410-f3r: > Maybe we can create a type alias PoolId for u128 like we have for market MarketId

Done

> Not sure if we ever change this, but u64 is likely sufficient and most 64bit processors have instructions to directly manipulate them in one register, therefore being more efficient in handling them.

For runtime, `wasm32-unknown-unknown` has a pointer size of 32 bits and the change to `u64` should probably create more optimized assembly, although I am not sure about its feasibility.

c410-f3r:
To fix https://github.com/zeitgeistpm/zeitgeist/issues/125, non-winning assets are now removed from the pool's resolved market and that is basically only it.
Everything else are improvements: Additional tests, pool status (Active, Stale), minor refactors and docs.

Again, it took longer than expected but on the bright side, I am more used to the PM pallet internals.

zrml/prediction-markets/src/lib.rs
c410-f3r: Sorry @sea212, I was a bit confused about what should be done and these parameters transfer shouldn't be necessary anymore

zrml/prediction-markets/src/lib.rs
```diff
@@ -1328,6 +1314,66 @@ mod pallet {
...
+                    for asset in assets_iter {
+                        let _ = manage_asset(asset, winning_asset_idx);
+                    }
+                }
+            }
+
+            Ok([total_accounts, total_asset_accounts, total_categories])
+        }
+
+        // If a market has a pool that is `Active`, then changes from `Active` to `Stable`.
```
lsaether, line 1361:
```suggestion
        // If a market has a pool that is `Active`, then changes from `Active` to `Stale`.
```

zrml/swaps/src/lib.rs
```diff
@@ -733,5 +736,38 @@ mod pallet {
...
+                        } else {
+                            return Err(Error::<T>::PoolDoesNotExist.into());
+                        };
+                        pool_mut.assets.retain(|el| {
+                            if let &Asset::CategoricalOutcome(_, idx) = el {
+                                idx == winning_asset_idx
+                            } else {
+                                false
+                            }
+                        });
```
lsaether, line 763:
```suggestion
                        pool_mut.assets.retain(|el| {
                            if let &Asset::CategoricalOutcome(_, idx) = el {
                                idx == winning_asset_idx
                            } else if let &Asset::Ztg = el {
                              true
                            } else {
                                false
                            }
                        });
```

Ztg should be retained too

lsaether (changes requested):
The stale pool should retain ZTG too, otherwise the changes look very good

c410-f3r:
Thanks @lsaether !

lsaether:
@c410-f3r  Last thing, could you bump the runtime spec? We'll deploy this to Battery Park

c410-f3r:
> @c410-f3r Last thing, could you bump the runtime spec? We'll deploy this to Battery Park

Done!

sea212:
Since the Pool storage layout has changed we also have to think about [storage migrations](https://hackmd.io/BQt-gvEdT66Kbw0j5ySlWw#2-Writing-the-Migration).

c410-f3r:
Like here?
```
pub type Pools<T: Config> = StorageMap<_, Blake2_128Concat, u128, Pool<BalanceOf<T>, T::MarketId>>;
```

I will revert this and any other breaking stuff to avoid storage fallout

sea212:
Yes, for example the [struct Pool](https://github.com/zeitgeistpm/zeitgeist/blob/e2273c2ef3e77ef9a849090be9de791e93b127e6/primitives/src/pool.rs#L14) has been changed (new field added), which is used in storage.

c410-f3r:
Pool status is orthogonal to #125, so it is possible to first merge stuff that fixes liquidity withdraw and then write migrations for pool status

c410-f3r:
Needs merging of https://github.com/zeitgeistpm/zeitgeist/pull/140

lsaether:
Please update the `types.json` then alert @morkeltry that there will be an upcoming change to types that's needed in the SDK.

c410-f3r:
Done

c410-f3r:
A migration was added to avoid storage conflict

zrml/swaps/src/migrations.rs
```diff
@@ -0,0 +1,110 @@
...
+        sp_runtime::RuntimeDebug,
+    )]
+    struct PreviousPool<B, MI> {
+        assets: Vec<Asset<MI>>,
+        swap_fee: B,
+        total_weight: u128,
+        weights: BTreeMap<Asset<MI>, u128>,
+    }
+
+    pub fn migrate<T>() -> Weight
```
lsaether, line 27: Should this bump the pallet storage version?

lsaether:
Overall looks good, just had a question about whether we need to explicitly set the new pallet storage version as part of the migration logic. Also, create #147 as another enhancement to build on top of this, so that we can manually manage the pools that are already expired using Sudo on the testnet.

zrml/swaps/src/migrations.rs
c410-f3r: Some pallets do explicitly use storage versioning with an enum:

```rust
enum StorageVersion {
  V1,
  V2,
  ...
}
```

And others simply use the crate metadata version as a way to track upgrades, which is the chosen approach simply because it is a little easier, the default choice since https://github.com/paritytech/substrate/pull/7208 and we are temporarily migration a testnet

What I couldn't find with version-based migrations is the exposure of public API/Storage versions. In any case, I can use an `enum` to track storage versions if desirable.

sea212 (approved):
Thanks for adding this feature, looks good overall. Some weights have to be re-benchmarked but the difference should be only minor, so we can skip it for now (I'll add an automatic benchmarking system soon anyways).
Besides that I share Logans concern, but I'll comment on that rather than opening a new conversation.

zrml/swaps/src/migrations.rs
sea212: I see that the current variant is convenient in that it does not require to add a new storage value, but I think this will make updates more complicated overall. Please correct me if I am wrong, but as I see it now we have to update the pallet version for the next runtime upgrade, even if no changes to the pallet were made, because otherwise this condition:
https://github.com/zeitgeistpm/zeitgeist/blob/a144ed5a1c9cfcffdefd0f7e8beafd7797bd2ca5/zrml/swaps/src/migrations.rs#L35-L40
will be true again and the migration will be executed.

zrml/swaps/src/migrations.rs
```diff
@@ -0,0 +1,110 @@
+pub mod _0_1_1 {
```
sea212, line 1: I'd give this an more descriptive name additionally, such we know which migration did what in the future. For example: `pub mod _0_1_1_add_pool_status`

c410-f3r: Nice

zrml/swaps/src/migrations.rs
sea212: Ah, I see what was my error in thought. The storage_version is a value inside storage that contains the pallet version before the runtime upgrade and it is updated after a runtime upgrade (in case the pallet version changed). This is smooth!

lsaether (approved):

zrml/swaps/src/migrations.rs
```diff
@@ -29,18 +29,14 @@ pub mod _0_1_1_add_pool_status {
         T: Config,
     {
         let mut weight: Weight = 0;
-        let storage_version = <Pallet<T>>::storage_version().unwrap_or(PalletVersion {
+        let _0_1_0 = PalletVersion {
```
lsaether, line 32: `version_0_1_0` ?

c410-f3r:
Thank you guys

Issue #135, opened 2021-05-17
Maintain a `types.json` that defines chain types


lsaether:
In order to easily share with people our chain types, we should maintain a `types.json` file inside of this repository and make sure it's maintained any time types are changed in the rust code. This would be beneficial in the fact that we can easily share types to others, and also for SDK development to be able to reference something that is not the source code (which can be opaque and hard to follow).

lsaether:
completed in #136

PR #136, opened 2021-05-17
Add types.json

sea212:
**Issue:** #135 

This PR adds the types.json. It renamed the "templates" folder into "misc", where the old weight template and the newly added *types.json* are located.

Currently a type conflict for AccountData between the balances pallet and the tokens pallet exists. The *types.json* supplied in this PR favors the AccountData type for the tokens pallet. The balances can currently be retrieved by querying the system pallet


c410-f3r (approved):
Lots of custom types 👀

Hope https://github.com/paritytech/substrate/pull/8615 will help in the near future

lsaether (approved):
Thanks!

lsaether:
> Currently a type conflict for AccountData between the balances pallet and the tokens pallet exists. The types.json supplied in this PR favors the AccountData type for the tokens pallet. The balances can currently be retrieved by querying the system pallet

Does this have a solution inside the `types.json` or can we only fix it on the upper layers using a type alias?

sea212:
@lsaether Good point, seems that type aliases also work in the types.json. In case a app does not support this, it will be just another type that we don't use.

Issue #138, opened 2021-05-20
types.json has conflict between AccounData of balances and tokens pallet


sea212:
**Issue**
See title

**Solution**
Add type alias

PR #139, opened 2021-05-20
Add type alias for tokens Accountdata

sea212:
Issue: #138

### DIFF
--- a/misc/types.json
+++ b/misc/types.json
@@ -1,132 +1,144 @@
 {
-    "Address": "MultiAddress",
-    "Amount": "i128",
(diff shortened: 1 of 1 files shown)


lsaether:
Cool! Is this a self-contained solution or does it only work because we've already created [that alias in the type definitions](https://github.com/zeitgeistpm/tools/blob/main/packages/type-defs/src/index.ts#L7-L15) which Polkadot-JS Apps has registered? Either way, I think it's okay, type aliases should probably not come up very often.

lsaether (approved):

PR #144, opened 2021-05-21
Add court pallet

c410-f3r:
The court pallet is basically the current dispute logic inside the PM pallet. This separations of responsibilities will be useful for upcoming additions and enhancements.

Instead of proving everything at once, changes will be delivered in batches to keep PRs small. By doing so, reviews and merges are likely to become faster and easier 🤞.

This PR simply copy, paste and adapt disputing logic from the PM pallet and moves common types from `prediction-markets/market.rs` to `zeitgeist-primitives/market.rs` and that is it. No runtime or nodes changes, not even the original PM module internals have been modified.


lsaether:
The next steps seems to be exploring the interface between prediction markets and the `DisputeHandler` of which the Court pallet implements.

The approach I'm thinking of is that the `dispute` function remains on the PredictionMarkets pallet, but inside of the dispute function we have something like:

```rust
/// in PredictionMarkets pallet

fn dispute() {
   ...
   T::DisputeHandler::on_dispute(market_id);
}
```

This way, although the `DisputeHandler` will be tightly coupled to the PredictionMarkets pallet, it would be theoretically possible to swap out one pallet that implement the `DisputeHandler` interface with another one so that we have a modular way to handle disputes in markets.

Additional to this, resolution should also be handled primary by the PredictionMarkets pallet, with another hook on `DisputeHandler`, like `DisputeHandler::on_resolution(market_id)`. 

The reason I'm suggesting this kind of architecture is because it's foreseeable that at some point in the future we would want to do a complete swap of one type of Court logic for a new type of Court logic. So having a `DisputeHandler` trait would be the right way to manage this. I'm open for discussion though, in case I'm missing a detail that would make this type of architecture undesirable.

c410-f3r:
Thanks for all the information @lsaether 

Just added the suggested trait to the pallet structure. Next steps will probably be: Add tests, PM integration, runtime addition, documentation, benchmarks and other possible enhancements.

lsaether (approved):

lsaether:
Merging as this does not currently alter any parts of the runtime, and is an intermittent step toward further work on the Court pallet.

Issue #131, opened 2021-05-14
Genesis storage root hash differs


sea212:
The most recent source code will compile into a binary that creates a different storage root hash in genesis, although the same chainspec is applied.

c410-f3r:
With the upcoming integration of new runtime pallets and initial configurations (staking, crowdloan, Liquidity mining, ...), I am afraid more genesis conflicts will appear because most of them do modify initial state.

lsaether:
Closed in #146

PR #146, opened 2021-05-26
Add complete chain spec

sea212:
**Issue:** #131 

**Problem**
TL;DR: The chain spec was incomplete. Export raw chain spec after genesis block was created.

Our previous chain spec was incomplete. The question arises, why the genesis state root hash was the same for the same source files (on different architectures) but different when changes occurred. If we inspect fb127223ea8990bb27819dbbb9b15a46d7ffea73 (the chain spec was generated from this) and ebc44c4f58ef40ed1a5f9a88ccd2cb946780e3d3, we see that the only changes were two constants being changed, `spec_version` and `MaxCategories`. Those constants are usually included as constants in the runtime and the runtime was correctly fixed to the one in the chain spec. After further inspection, it becomes clear that the genesis state root hash differed at this point:
```
"0x26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8": "0x3c247a6569746765697374", <-- newer (spec version update)
"0x26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8": "0x38247a6569746765697374", <-- older
```

What is this? By hashing the pallet names and their storage key names, it turned out that the key `0x26aa394eea5630e07c48ae0c9558cef7f9cce9c888469bb1a0dceaa129672ef8` refers to: `System.LastRuntimeUpgrade`. In the newer version it was this:
```
0x3c247a6569746765697374 = {
  specVersion: 15,
  specName: zeitgeist
}
```

Here we can see that a simple adjustment of `spec_version` changed the storage and therefore implicitly the state root. 
_Personal side note: Isn't it funny that changing the spec version to signal changes for updates did break our updates in the first place?_
Since this storage data was not included in the chain spec, it was written into storage by some initialization code in the system pallet.

**Solution**
To generate a complete raw chain specification, you *can* follow these steps:

1. Run the node for the relevant chain and interrupt it immediately after it printed the genesis state root hash: `zeitgeist --chain relevant_chain -d temp/path/to/chain/data`
2. Execute this command to generate the chain spec: `zeitgeist export-state --chain relevant_chain -d temp/path/to/chain/data -- 0 > relevant_chain.json`
3. Delete the temporary chain folder: `rm -r temp/path/to/chain/data`


lsaether:
Cool! Thanks for investigating this and finding the answer. So I guess this answers the question @c410-f3r  raised earlier on whether the JSON chainspec will work, and now we know that it doesn't. Only the raw chainspec does. 

I actually did not realize this was the case. My intuition is that the JSON chainspec is really not a chainspec, but a genesis configuration, and therefore it re-creates the state from the config on startup. You would still think the two of them would lead to the creation of the same genesis hash... but here we are 🤷 

I tested this already and it works well.

lsaether (approved):

c410-f3r (approved):
Glad this issue probably won't be a problem anymore. Thanks

PR #152, opened 2021-05-29
Improve `buy_complete_set` documentation

### LINKED ISSUE #143: Glitch issue
Guys, I am going to try to explain it once again. I am just trying to help because if this is how it goes live it can mess up hard on you. Also because I do not want to get any kind of restricction since i have been following for a while and ones of the first ones on the telegram group.

When you buy tokens directly from https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Froc.zeitgeist.pm#/extrinsicsyour  it is meant to be 1:1 but since you are selecting "buy complete set" you are getting 1:9

Let me explain. when I saw the broadcast and saw the possibility for som arbitrage with Slot 2 and shiden (cost was 1.22) I bought the complete set of 47 ZBT and sold them there for I got 56ish ZBT

So far so good....
[7:58]
Until I found out I also got 47 Karura, 47 River, 47 Khala....etc

And did the cycle over and over again until getting 2000ish ZBT just in half an hour
[7:59]
At the moment there was several horses arround 50% (after selling all thats what got horses really behind)
[8:00]
I do not know anything about coding but a lot about predictions marckets. And here are 2 ways you can solve this issue
[8:02]
1 not buying the whole (complete)set for only 1 ZBT you should be able to buy only 1 token(or karura or shiden or, etc... but not all) of the set
[8:04]
2nd option is to doing it as bets, if you think that is right as it is meant to work, then the Sum of all horses together can not be higher than 100% IF all together are over a 100% here is when we can take profit since we will be given 1 token of each when buying the whole set.
[8:07]
I hope I was able to explain myself, I submitted a ticket a couple of days ago and reply that this was how it was meant to work. But really it is not and the only afected will be you, not the community and in live net you can get a lot of losses. 

Now I been doing it because is a testnet, and because I was told it was Ok, I hope I do not get any kind of disqualification for this since being telling it since discovered it and as mentioned told it was ok. But it is important to fix it
[8:07]
I can share you my wallet address as well so you can track the movements of it and verify all this info
[8:07]
Sorry for the long post...

c410-f3r:
Fixes #143

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -259,7 +259,7 @@ mod pallet {
         /// function on a categorical market with five different outcomes, five different shares
         /// will be transferred to the callee.
         ///
-        /// The amount of each share will equals the provided `amount` parameter.
+        /// The amount of each share will equal the provided `amount` parameter.
         ///
         /// NOTE: This is the only way to create new shares.
         // Note: `buy_complete_set` weight consumption is dependent on how many assets exists.
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -255,10 +255,13 @@ mod pallet {
             Ok(())
         }
 
-        /// Generates a complete set of outcome shares for a market.
+        /// Buys the complete set of outcome shares of a market. For example, when calling this
+        /// function on a categorical market with five different outcomes, five different shares
+        /// will be transferred to the callee.
+        ///
+        /// The amount of each share will equals the provided `amount` parameter.
         ///
         /// NOTE: This is the only way to create new shares.
-        ///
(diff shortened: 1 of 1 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -255,10 +255,13 @@ mod pallet {
             Ok(())
         }
 
-        /// Generates a complete set of outcome shares for a market.
+        /// Buys the complete set of outcome shares of a market. For example, when calling this
+        /// function on a categorical market with five different outcomes, five different shares
+        /// will be transferred to the callee.
         ///
-        /// NOTE: This is the only way to create new shares.
+        /// The amount of each share will equals the provided `amount` parameter.
```
lsaether, line 262:
```suggestion
        /// The amount of each share will equal the provided `amount` parameter.
```

lsaether (approved):
Great, thank you for this.

Issue #96, opened 2021-04-26
Rikiddo Pallet


lsaether:
Tracking issue for implementation of the Rikiddo pallet.

PR #155, opened 2021-05-31
[LSD-LMSR] Add LSD-LMSR skeleton

sea212:
Issue: #158

This PR creates the scaffold for the LSD-LSMR pallet as preparation for #96


sea212:
The trait, types and constants modules have to be moved out into a separate crate before integrating this into the runtime (to avoid circular dependencies). Would you prefer them to be contained within zeitgeist-primitives or an extra crate?

sea212:
Please do not review yet, just figured out the design needs some adjustments.

zrml/lsd-lmsr/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+            + Default
+            + Copy
+            + MaybeSerializeDeserialize
+            + Debug;
+
+        /// The type that offers timestamping functionality
+        type Timestamp: Time;
+
+        /// The type that will be used for the fractional part of the fixed point numbers
+        type FractionalType: LeEqU128;
```
sea212, line 45: I was trying to the deduce the type that determines the number of fractional bits from the number of one unit of token (in our case BASE) automatically and use it in storage, but failed after thinking through multiple options. I think since this number should not change during the whole lifetime of the chain it is okay to supply it this way.

zrml/lsd-lmsr/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[dependencies]
+frame-support = { branch = "polkadot-v0.9.4", default-features = false, git = "https://github.com/paritytech/substrate" }
+frame-system = { branch = "polkadot-v0.9.4", default-features = false, git = "https://github.com/paritytech/substrate" }
+parity-scale-codec = { default-features = false, features = ["derive"], version = "2.0" }
+sp-runtime = { branch = "polkadot-v0.9.4", default-features = false, git = "https://github.com/paritytech/substrate" }
+sp-std = { branch = "polkadot-v0.9.4", default-features = false, git = "https://github.com/paritytech/substrate" }
+substrate-fixed = { default-features = false, features = ["serde"], git = "https://github.com/encointer/substrate-fixed", package = "substrate-fixed" }
```
c410-f3r, line 7:
```suggestion
substrate-fixed = { default-features = false, features = ["serde"], git = "https://github.com/encointer/substrate-fixed" }
```

zrml/lsd-lmsr/src/constants.rs
```diff
@@ -0,0 +1,36 @@
...
+/// One hour.
+pub const EMA_SHORT: Timespan = Timespan::Hours(1);
+
+/// Default long EMA period (in seconds)
+/// Six hours.
+pub const EMA_LONG: Timespan = Timespan::Hours(6);
+
+/// Default smoothing factor for EMA calculation
+/// 2.0
+pub const SMOOTHING: FixedU32<U24> = <FixedU32<U24>>::from_bits(0x0200_0000);
```
c410-f3r, line 15: It is a pity `from_num` is not a constant function.

Can you answer me a question? The default substrate `FixedU128` structure has 18 digits and an internal `2_000_000_000_000_000_000` decimal actually means `2.0`. So does `U24` here means the number of decimal digits (fractions)?

Given that:

- `u32` can only hold a maximum of `4294967295` decimal.
- `0x0200_0000` hex is `33554432` decimal but in fact represents `2.0`

I am a bit confused about how `substrate_fixed::Fixed(U/I)(32/64/128)` works compared to the vanilla substrate version.

c410-f3r (approved):
Just some nits

zrml/lsd-lmsr/src/lib.rs
c410-f3r: This type would be beneficial if two different runtimes are using two different `FractionalType`. For example, for some reason polkadot's runtime is using `FractionalType = U30` and kusama's runtime is using `FractionalType = U24` or we only have one runtime but want to test different fractions inside a pallet's mock (can't imagine why this would be necessary).

In my opinion, I think it is probably better to stick with a fixed fraction

zrml/lsd-lmsr/src/constants.rs
sea212: I agree in regards to `from_num` not being a constant function, now it's complicated to interpret the numbers.

The `U` type specifies the number of fractional bits. It is used to move the comma in the bit representation. For `FixedU32<U24>`, we have `32-24 = 8` of the highest bits for the integer part, and the remaining `24` bits for the fractional part. The lowest precision is `1.0 / pow(2, fractional_bits)`, so for `U24` this is `1.0 / 2.0^24.0 =  0.00000006`.

c410-f3r: Oh, fractions are bit related. Thanks for the explanation

Even if some cases aren't so straightforward, it is probably better to offset where possible, for example, `0x0200_0000` == `2 << 24`

zrml/lsd-lmsr/src/constants.rs
```diff
@@ -0,0 +1,36 @@
...
+
+// --- Default configuration for FeeSigmoidConfig struct ---
+/// Initial fee f
+/// 0.03
+pub const INITIAL_FEE: FixedU32<U32> = <FixedU32<U32>>::from_bits(0x07AE_1478);
```
c410-f3r, line 20: `INITIAL_FEE` is actually `0.0299999993`. Is this intended?

sea212: > `INITIAL_FEE` is actually `0.0299999993`. Is this intended?

No, assigned the correct value to it and some other constants now. Thanks!

zrml/lsd-lmsr/src/constants.rs
```diff
@@ -0,0 +1,36 @@
...
+/// 2.0
+pub const SMOOTHING: FixedU32<U24> = <FixedU32<U24>>::from_bits(0x0200_0000);
+
+// --- Default configuration for FeeSigmoidConfig struct ---
+/// Initial fee f
+/// 0.03
+pub const INITIAL_FEE: FixedU32<U32> = <FixedU32<U32>>::from_bits(0x07AE_147B);
+
+/// Minimal revenue w (proportion of initial fee f)
+/// f * β = 0.03 * 0.07 = 0.021
```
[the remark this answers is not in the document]
sea212: Oh there is a typo in the comment, β was 0.7 and the initial fee f 0.03 and the minimal revenue 0.021 = 2.1% .
Based on your comment the initial fee is now 0.005, so the minimal revenue should be `w = 0.005 * 0.07 = 0.0035` (0.35%), right?

lsaether (approved):

PR #156, opened 2021-06-01
[Court] Add remaining tweaks for PM integration

c410-f3r:
Locally, the `court` pallet is already integrated into the PM pallet but all changes would create a sizeable PR. Therefore, this PR only touches self-contained `court` modifications for an easier review.

### DIFF
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -18,6 +18,11 @@ pub const DOLLARS: Balance = BASE / 100; // 100_000_000
 pub const CENTS: Balance = DOLLARS / 100; // 1_000_000
 pub const MILLICENTS: Balance = CENTS / 1000; // 1_000
 
+// Court parameters
+parameter_types! {
+    pub const CourtPalletId: PalletId = PalletId(*b"zge/cout");
+}
+
 // Prediction Market parameters
 parameter_types! {
     pub const AdvisoryBond: Balance = 25 * DOLLARS;
@@ -33,6 +38,11 @@ parameter_types! {
     pub const ValidityBond: Balance = 50 * DOLLARS;
 }
 
+// Staking parameters
+parameter_types! {
+    pub const DefaultBlocksPerRound: u32 = 2 * BLOCKS_PER_MINUTE as u32;
+}
+
 // Swaps parameters
 parameter_types! {
     pub const ExitFee: Balance = 0;
@@ -44,9 +54,4 @@ parameter_types! {
     pub const MinLiquidity: Balance = 100 * BASE;
     pub const MinWeight: Balance = BASE;
     pub const SwapsPalletId: PalletId = PalletId(*b"zge/swap");
-}
-
-// Collator staking
-parameter_types! {
-    pub const DefaultBlocksPerRound: u32 = 2 * BLOCKS_PER_MINUTE as u32;
 }
--- a/zrml/court/src/court_pallet_api.rs
+++ b/zrml/court/src/court_pallet_api.rs
@@ -1,11 +1,79 @@
-use frame_support::dispatch::{DispatchError, DispatchResultWithPostInfo, Weight};
-use zeitgeist_primitives::types::OutcomeReport;
+use crate::ResolutionCounters;
+use alloc::vec::Vec;
+use frame_support::dispatch::{DispatchError, DispatchResultWithPostInfo};
+use zeitgeist_primitives::types::{Market, MarketDispute, OutcomeReport};
 
 /// Court - Pallet Api
 pub trait CourtPalletApi {
+    type AccountId;
     type BlockNumber;
     type MarketId;
     type Origin;
+
+    // Market
+
+    /// Gets a market from the storage.
+    fn market(
+        market_id: &Self::MarketId,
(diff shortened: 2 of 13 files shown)


primitives/src/constants.rs
```diff
@@ -18,6 +18,11 @@ pub const DOLLARS: Balance = BASE / 100; // 100_000_000
 pub const CENTS: Balance = DOLLARS / 100; // 1_000_000
 pub const MILLICENTS: Balance = CENTS / 1000; // 1_000
 
+// Court parameters
+parameter_types! {
+    pub const CourtPalletId: PalletId = PalletId(*b"zge/cout");
+}
+
```
c410-f3r, line 25: Used by tests across different crates and runtime definitions

zrml/court/src/court_pallet_api.rs
```diff
@@ -1,12 +1,80 @@
...
+        market: Market<Self::AccountId, Self::BlockNumber>,
+    );
+
+    /// Mutates a given market storage
+    fn mutate_market<F>(market_id: &Self::MarketId, cb: F) -> Result<(), DispatchError>
+    where
+        F: FnOnce(&mut Market<Self::AccountId, Self::BlockNumber>);
+
+    /// Removes a market from the storage.
+    fn remove_market(market_id: &Self::MarketId) -> Result<(), DispatchError>;
```
c410-f3r, line 32: Both `court` and `PM` pallets use the `Markets` storage so it is necessary to create an common entry-point. In this case, `court` holds `Markets` and `PM` consumes the API

zrml/court/src/lib.rs
```diff
@@ -415,19 +533,12 @@ mod pallet {
...
-                Ok(0)
-            } else {
-                Ok(0)
-            }
+            Ok(ResolutionCounters {
+                total_accounts,
+                total_asset_accounts,
+                total_categories,
+                total_disputes,
+            })
```
c410-f3r, line 541: There is no `call` method so the pertinent information for benchmarks is stored in this structure for `PM`.

**Overall, the benchmark history is something that still needs some adjustments**

zrml/court/src/lib.rs
```diff
@@ -260,54 +392,43 @@ mod pallet {
...
     /// been issued.
     #[pallet::storage]
-    #[pallet::getter(fn markets)]
-    pub type Markets<T: Config> = StorageMap<
-        _,
-        Blake2_128Concat,
-        T::MarketId,
-        Option<Market<T::AccountId, T::BlockNumber>>,
-        ValueQuery,
-    >;
```
c410-f3r, line 300: Optional queries are being used because, IIRC, it is impossible to distinguish between `vec![]` being a stored value without results or `vec![]` being the returned value of an non-existent query where `T: Default`.

zrml/court/src/lib.rs
```diff
@@ -42,6 +45,8 @@ mod pallet {
     pub const NOT_RESOLVED: DispatchError = DispatchError::Other("Resolved outcome does not exist");
     pub const OUTCOME_MISMATCH: DispatchError =
         DispatchError::Other("Submitted outcome does not match market type");
+    pub const UNKNOWN_BLOCK: DispatchError = DispatchError::Other("Unknown block");
+    pub const UNKNOWN_MARKET_ID: DispatchError = DispatchError::Other("Unknown market id");
```
sea212, line 49: Would be nice to have those in the error enum.

sea212 (commented):
Looks good overall, would be nice to have a different approach for shared storage though.

zrml/court/src/court_pallet_api.rs
c410-f3r: Yes, this schema is semantically wrong, although that's what I could formulate given the constraints.
This subject needs more research and https://github.com/paritytech/substrate/issues/8320 might have some answers so I will open a related PR afterwards if it is not a problem.

zrml/court/src/lib.rs
c410-f3r: Done

zrml/court/src/court_pallet_api.rs
lsaether: Yeah, I don't think this is the storage format we want in the final form, it makes Prediction Markets way too tightly coupled to Courts storage and would make replacing the Court pallet with something different in the future more difficult. The ideal interface between PM and Court should be that if we wanted to swap Courts with a <DISPUTE_MECHANISM>, it should be more or less a drop in replacement without impacting any markets in storage or possibly even requiring any migrations. The maximal version of this would be having multiple parallel dispute options at once, and they can be selected by the person making the dispute.

c410-f3r: Research of alternative methods to share storage would take more time so I simply created another pallet called `market-commons` that holds `Markets` and any other future common parameter between `PM` and `court`.

lsaether (approved):
> Research of alternative methods to share storage would take more time so I simply created another pallet called market-commons that holds Markets and any other future common parameter between PM and court.

This sounds reasonable to me and satisfies all constraints. 

However, it is worth noting that when we deploy this it would require a migration of storage from the PM pallet to MC pallet.

lsaether:
As this does not impact any current runtime logic yet, I will merge this.

PR #142, opened 2021-05-20
Update to polkadot v0.9.2

c410-f3r:
For Kusama parachains and also because Rococo was reset to `0.9.x`

- [x] Wait for Moonbeam upgrade

PR #164, opened 2021-06-03
Update to polkadot 0.9.3

c410-f3r:
To my surprise `WasmBuilder::build` automatically imports all features but "std" and "default" declared on `Cargo.toml` (https://github.com/paritytech/substrate/blob/d6e4db6c30667bf3524abced8428401ab141b07a/utils/wasm-builder/src/wasm_project.rs#L316-L322). So a structure like:
```toml
parachain-std = [ .. ]
std = [ .. ]
```
Won't work because WASM is a `no_std` environment and `parachain-std` will turn on `std` intrinsics.

Since the activation of an optional feature also implies the activation of its dependency, the only way I could find to separate `standalone` and `parachain` dependencies was by using the `weak-dep-features` cargo feature.

This PR is almost the same as #142. Everything is compiling and collators are producing blocks locally.


lsaether:
Awesome! Thanks

lsaether (approved):
Well, that's pretty interesting. Do you know which package it was that was causing the problem?

c410-f3r:
Thanks! With polkadot out of the standalone dependencies, we probably won't face more out of space errors

PR #167, opened 2021-06-04
Integrate Court pallet

c410-f3r:
- [x] Fix tests
- [x] Fix benchmarks
- [x] Add migrations (Postponed)


zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -329,39 +331,45 @@ benchmarks! {
...
         for i in 0..c.min(d) {
-            let _ = Call::<T>::dispute(marketid, OutcomeReport::Categorical(i.saturated_into()))
-                .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
+            let _ = T::Court::on_dispute(
+                RawOrigin::Signed(caller.clone()).into(),
+                marketid,
+                OutcomeReport::Categorical(i.saturated_into())
+            );
         }
-    }: { Pallet::<T>::internal_resolve(&marketid)? }
```
c410-f3r, line 341: `internal_resolve` is a private function of `Court` and can't be called here. On the other hand, benchmark transposition to the `Court` pallet can't be applied because of the lack of some functions like https://github.com/zeitgeistpm/zeitgeist/blob/37aaf94def04d1323362dbea3c193cf09b093901/zrml/prediction-markets/src/benchmarks.rs#L78.

Not sure how to proceed here. Can you provide some thoughts @sea212 ?

zrml/prediction-markets/src/lib.rs
```diff
@@ -214,19 +215,20 @@ mod pallet {
...
-            )
-            .into())
+            Ok(Some(0u64).into())
+            //let weight = Self::internal_resolve(&market_id)?;
+            //Ok(Some(
+            //    weight
+            //        .saturating_add(T::WeightInfo::admin_move_market_to_resolved_overhead())
+            //        .saturating_sub(T::WeightInfo::internal_resolve_scalar_reported()),
+            //)
+            //.into())
```
c410-f3r, line 231: These and other benchmarks comments will be evaluated once https://github.com/zeitgeistpm/zeitgeist/pull/167#discussion_r652274415 is resolved

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,138 @@
...
+                T::MarketId,
+                Option<PoolId>,
+                Blake2_128Concat,
+            >(PM, b"MarketToSwapPool")
+            .filter_map(|(k, v)| Some((k, v?)))
+            {
+                weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+                T::MarketCommons::insert_market_pool(k, v);
+            }
+        }
```
sea212, line 67: We have to remove the old storage from the prediction markets pallet, otherwise we have a zombie storage in the prediction market pallet that is not used but consumes disk space. I think the easiest way is to move the complete prefix inside a storage: [`move_storage_from_pallet`](https://releases.parity.io/substrate-rustdoc/frame_support/storage/migration/fn.move_storage_from_pallet.html). You can also remove the complete storage prefix, after you completed the migration (could be useful if you only want to move certain (k,v) pairs): [`remove_storage_prefix`](https://releases.parity.io/substrate-rustdoc/frame_support/storage/migration/fn.remove_storage_prefix.html)

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,138 @@
...
+                );
+
+                PredictionMarkets::on_runtime_upgrade();
+
+                assert_eq!(Court::market_ids_per_dispute_block(&0).unwrap(), vec![1]);
+                assert_eq!(Court::market_ids_per_report_block(&0).unwrap(), vec![1]);
+
+                assert_eq!(MarketCommons::market(&0).unwrap(), DEFAULT_MARKET);
+                assert_eq!(MarketCommons::market_pool(&0).unwrap(), 1);
+            })
```
sea212, line 135: Some additional tests that ensure, that the prediction market pallet does not have those values in storage anymore.

zrml/prediction-markets/src/benchmarks.rs
sea212: You can benchmark the `internal_resolve` variants in the court pallet. Then you can include the resulting weights module in the court pallet `pub mod weights`. After that you can include the weights module from the court pallet in the prediction market pallet and use it to calculate the actual weight in the relevant functions (`admin_move_market_to_closed`, `admin_move_market_to_resolved`):

```rust
// zrml/prediction-markets/src/lib.rs
// rename might be required
use zrml_court::weights::{WeightInfo, WeightInfoZeitgeist};

// -- snip --

let court_internal_resolve_weight = WeightInfo::<T>::internal_resolve_categorical_disputed(a, b, c, d);
// do weight adjustments
```

You also have to think about restructuring the benchmark functions based on the changes you made and about adding new benchmarks. For example, `on_initialize()` now became very complex and new benchmarks are required for a proper weight adjustment:

```rust
fn on_initialize(now: T::BlockNumber) -> Weight {
    let _ = T::Court::on_resolution(now);
    0
}
```

it calls the court pallet which in turn calls the (externally) private `internal_resolve` which is very complex. Instead of weight 0, it should return the proper benchmarked weight, which does depend on the weight calculations of the court pallet.

sea212 (changes requested):

c410-f3r:
Thanks for the review @sea212 
I will reach back once everything is properly addressed

zrml/court/src/benchmarks.rs
```diff
@@ -0,0 +1,205 @@
...
+fn create_close_and_report_market<T: Config>(
+    permission: MarketCreation,
+    options: MarketType,
+    _outcome: OutcomeReport,
+) -> Result<(T::AccountId, MarketIdOf<T>), &'static str> {
+    let (caller, marketid) = create_market_common::<T>(permission, options)?;
+    //    let _ = Call::<T>::admin_move_market_to_closed(marketid)
+    //        .dispatch_bypass_filter(T::ApprovalOrigin::successful_origin())?;
+    //    let _ = Call::<T>::report(marketid, outcome)
+    //        .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
```
c410-f3r, line 67: @sea212 Just for the sake of benchmarks, commented methods will have to be moved from PM to `MarketCommons`. Can I simply create local functions that mimic them?

Otherwise, what semantically prevents moving other call methods like `report`, `sell_complete_set`, etc?

sea212: Posting the result of the private discussion:

I think we have to rethink the benchmarks. All this moving around is likely not necessary. For all the changed functions, we have to think again about: What are the logical paths the call can take and on which variables does its execution length depend on? Based on this question we have to rewrite the benchmarks for those functions.
One example is the internal_resolve function. Before it was called within the PM pallet when an admin resolved the market or at on_initialize for each market that was reported or disputed. Now the question arises: How did this functions change and which logical paths can they take now and on which variables does their execution time depend?
The on_initialize function now calls T::Court::on_resolution(now). So we have to understand what this function does in regards to the criteria I mentioned and write appropriate benchmarks for the PM pallet that initialize the storage appropriately for the different cases and then run the benchmark by calling on_intialize. Within on_initialize, we have to distinguish the logical paths and invoke the weight function for the appropriate case.
I just saw that you already created and returned a ResolutionCounters struct, that contains all the relevant parameters that influence the execution time. That is great, now we only have to determine which weight function to call and insert the parameters there (for example, was it a scalar or a categorical market? Does it even make a difference now after the changes or can we unite them?)

c410-f3r:
> I think I already have enough to provide a MVP for these modified benchmarks.
Since it will take some time, can we merge https://github.com/zeitgeistpm/zeitgeist/pull/167 without the finished benchmarks? It would be nice because liquidity mining will use some of the MarketCommons structure.

I think it is fine to merge this and create a issue to adjust those benchmarks and add new weights. The issue must be resolved until the next runtime upgrade, when the new weights will take effect.

runtime/src/lib.rs
```diff
@@ -146,16 +146,21 @@ macro_rules! create_zeitgeist_runtime {
...
                 Timestamp: pallet_timestamp::{Call, Pallet, Storage, Inherent},
                 Tokens: orml_tokens::{Config<T>, Event<T>, Pallet, Storage},
                 TransactionPayment: pallet_transaction_payment::{Pallet, Storage},
 
+                // Zeitgeist
+
+                Court: zrml_court::{Event<T>, Pallet, Storage},
+                MarketCommons: zrml_market_commons::{Pallet, Storage},
+                Orderbook: zrml_orderbook_v1::{Call, Event<T>, Pallet, Storage},
+                PredictionMarkets: zrml_prediction_markets::{Call, Event<T>, Pallet, Storage},
```
sea212, line 161: Requires increment of `spec_version` and `transaction_version` before merge

sea212 (commented):
Most of the concerns have been addressed, including the most critical issue regarding the zombie storage. Since further work depends on the merge, imo it is okay to merge this after fixing merge conflicts and bumping `spec_version` and `transaction_version`. Please create an issue regarding the benchmark adjustments and resolve it asap.

c410-f3r:
Runtimes versions were updated and a new issue was opened to track benchmarks (https://github.com/zeitgeistpm/zeitgeist/issues/195)

runtime/src/lib.rs
lsaether: Why remove the explicit indices on the pallets?

c410-f3r: Bad re-base. Indices were re-introduced

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,178 @@
...
+    const MARKET_TO_SWAP_POOL: &[u8] = b"MarketToSwapPool";
+    const MARKETS: &[u8] = b"Markets";
+    const PM: &[u8] = b"PredictionMarket";
+
+    pub fn migrate<T>() -> Weight
+    where
+        T: Config,
+    {
+        let mut weight: Weight = 0;
+        let version_0_1_0 = PalletVersion { major: 0, minor: 1, patch: 0 };
```
lsaether, line 25: change to 0.1.1

zrml/prediction-markets/Cargo.toml
```diff
@@ -50,10 +52,12 @@ std = [
...
+    'zrml-court/std',
+    'zrml-market-commons/std',
 ]
 
 [package]
 authors = ["Zeitgeist PM <contact@zeitgeist.pm>"]
 edition = "2018"
 name = "zrml-prediction-markets"
-version = "0.1.0"
+version = "0.1.1"
```
lsaether, line 63: change to 0.1.2

zrml/market-commons/src/lib.rs
```diff
@@ -87,11 +129,32 @@ mod pallet {
...
     pub type Markets<T: Config> =
         StorageMap<_, Blake2_128Concat, T::MarketId, Market<T::AccountId, T::BlockNumber>>;
+
+    /// The number of markets that have been created (including removed markets) and the next
+    /// identifier for a created market.
+    #[pallet::storage]
+    #[pallet::getter(fn market_count)]
+    pub type MarketCounter<T: Config> = StorageValue<_, T::MarketId>;
+
+    /// Maps a market id to a related pool id. It is upper to the caller to keep and sync valid
```
lsaether, line 156:
```suggestion
    /// Maps a market id to a related pool id. It is up to the caller to keep and sync valid
```

c410-f3r: Fixed

lsaether (commented):
Overall all the logic migration looks good. However, I want to be more clear about some things, we should rename the trait `CourtsPalletApi` to `DisputeApi` and rename the "court" pallet to `simple-disputes` since this doesn't implement the Court logic yet.

c410-f3r:
> Overall all the logic migration looks good. However, I want to be more clear about some things, we should rename the trait `CourtsPalletApi` to `DisputeApi` and rename the "court" pallet to `simple-disputes` since this doesn't implement the Court logic yet.

Renamed

lsaether (approved):

PR #168, opened 2021-06-05
[Swaps] Add `admin_set_pool_as_stale` call

### LINKED ISSUE #147: Add an admin function for manually changing active pools to stale ones
Now that we have stale pools, we should add an admin function in case we need to turn active pools to stale ones, especially for pools that were created for market that expired before including the next runtime upgrade 0.1.1.

Expose `manage_pool_staleness` as a public dispatchable that is only allowed for the Root origin.

c410-f3r:
Fixes #147


lsaether (approved):
Thx!

PR #169, opened 2021-06-05
Remove `zrml-prediction-markets-rpc` crate

c410-f3r:


lsaether (approved):
Great, that was stale code for a while.

PR #172, opened 2021-06-09
Decouple benchmark dependencies from node

c410-f3r:
Taking aside the benchmark decoupling from the `node` crate, this PR also tries to remove unnecessary dependencies to speed-up builds and improve CI times.

Locally, `cargo check` went down from ~971 to ~948 crates.


node/src/service.rs
```diff
@@ -10,9 +10,17 @@ pub use service_standalone::{new_full, new_light, new_partial};
...
   zeitgeist_runtime::api::dispatch,
   zeitgeist_runtime::native_version,
   frame_benchmarking::benchmarking::HostFunctions,
 );
+
+#[cfg(not(feature = "runtime-benchmark"))]
+native_executor_instance!(
+  pub Executor,
+  zeitgeist_runtime::api::dispatch,
+  zeitgeist_runtime::native_version,
```
c410-f3r, line 25: IIRC, this removal shouldn't be a problem. Still need to double check it

scripts/tests/fuzz.sh
```diff
@@ -6,6 +6,6 @@ set -euxo pipefail
 
 RUNS="${RUNS:-50000}"
 
-RUST_BACKTRACE=1 cargo fuzz run --fuzz-dir zrml/prediction-markets/fuzz full_workflow -- -runs=$RUNS
+RUST_BACKTRACE=1 cargo fuzz run --fuzz-dir zrml/prediction-markets/fuzz pm_full_workflow -- -runs=$RUNS
 
-RUST_BACKTRACE=1 cargo fuzz run --fuzz-dir zrml/swaps/fuzz full_workflow -- -runs=$RUNS
+RUST_BACKTRACE=1 cargo fuzz run --fuzz-dir zrml/swaps/fuzz swaps_full_workflow -- -runs=$RUNS
```
sea212, line 11: This is a nice fix you included along the way.

scripts/tests/misc.sh
```diff
@@ -13,6 +13,11 @@ test_package_with_feature primitives std
 
 for package in zrml/*
 do
+  if [[ "$package" == "zrml/market-commons" ]]; then
+    test_package_with_feature "$package" std
+    continue
+  fi
+
```
sea212, line 20: There might be multiple pallets in the future that don't have the runtime-benchmarks feature, but I think it is reasonable to fix it when the second pallet without that feature is included.

zrml/court/Cargo.toml
```diff
@@ -18,15 +18,20 @@ zrml-swaps = { path = "../swaps" }
 
 [features]
 default = ["std"]
+runtime-benchmarks = [
+    "frame-benchmarking",
+    "frame-support/runtime-benchmarks",
+    "frame-system/runtime-benchmarks"
+]
 std = [
+    "frame-benchmarking?/std",
```
sea212, line 27: What does the `?` operator do in this context? (I was not able to find it in the cargo book)

sea212 (approved):
Thanks, the computers love you for this!

zrml/court/Cargo.toml
c410-f3r: When activating an optional feature like `serde/std`, Cargo unconditionally also brings the whole dependency and this situation was causing some problems on CI, thus the reason for https://github.com/zeitgeistpm/zeitgeist/pull/157.

In this example, `frame-benchmarking?/std` means that the `frame-benchmarking` crate won't be included when calling `cargo build --features std`. https://github.com/rust-lang/cargo/issues/8832

sea212: Sweet, thanks for clarifying!

lsaether (approved):

PR #174, opened 2021-06-10
Remaining fixes for parachain deployment

c410-f3r:
### What is necessary to become a collator

An account needs to have a minimum quantity of ZBP, get the instance keys through `author_rotateKeys`, join the set of available candidates calling `ParachainStake::joinCandidates` and then finally map its address to the instance account using `AuthorMapping::addAssociation`. More details are available at https://docs.moonbeam.network/node-operators/networks/collator/

### The problem

It is necessary to have at least one associated collator with sufficient funds to produce blocks at the genesis configuration. Otherwise, no transaction will ever be processed.
How this initial collator will be configured is still a mystery (at least for me) because `ParachainStake::joinCandidates` and `AuthorMapping::addAssociation` are transactions so it is like trying to register the first collator without a collator to produce the first block.

### Current behaviour for local test-nets

Collators with pre-inserted session keys are launched according to the chain specification. For example, `local_testnet` has Alice as an initial collator:

https://github.com/zeitgeistpm/zeitgeist/blob/d2badc863a54c73a2b90551ba1467d8fe8117228/node/src/chain_spec/local_testnet.rs#L26-L30

So it is just a matter of calling `zeitgeist --alice --chain local_testnet ...`, where `--alice` automatically inserts session keys.

### What was done for live-nets

A pre-created account with sufficient funds was created and inserted at the genesis, which is easy but different combinations of pre-created session keys along side different cli parameters (`--node-key`, etc) combinations where tried without success. For example:

```rust
 ( 
    hex![SOME_ACCOUNT_HEX],
    hex![SOME_SESSION_HEX],
    3000, 
), 
```

```bash
zeitgeist --chain battery_park_staging --node-key SOME_KEY_HEX ...
```

### Summary

Once I figure out how to set-up an initial collator, it should be possible to deploy on Rococo or any other Polkadot network


c410-f3r:
After several failed attempts, I leaning towards temporally having Alice as a collator to at least test the parachain on a live system. Otherwise, more time will be needed to launch on Rococo.

If someone knows how to boot a node with a pre-populated session keys like `--alice/--bob/--fred` does, feel free to answer it. No-one on the Element channel responded yet.

lsaether:
> If someone knows how to boot a node with a pre-populated session keys like --alice/--bob/--fred does, feel free to answer it. No-one on the Element channel responded yet.

Can you use `author_insertKey` with the keys initially set in the genesis config?

c410-f3r:
Already tried it without success. Anyway, I will try again.

https://github.com/PureStake/moonbeam/blob/56b9e95a9b634695f59a7c699bc68a5cfb695f03/runtime/moonbeam/src/lib.rs#L605-L615

sea212 (approved):

scripts/parachain/local.sh
```diff
@@ -91,13 +90,22 @@ start_collator() {
...
+    $LAUNCH_PARACHAIN_CMD & node_pid=$!
+
+    sleep 30
+    DATA='{ "id":1, "jsonrpc":"2.0", "method":"author_insertKey", "params":["nmbs", "'"$seed"'", "'"$public_key"'"] }'
+    curl -H 'Content-Type: application/json' --data "$DATA" localhost:$collator_rpc_port
+
+    sleep 5
+    kill $node_pid
+    sleep 5
+    $LAUNCH_PARACHAIN_CMD
```
c410-f3r, line 106: This part of the script waits some time to let to node start up and then inserts the provided `$seed` and `$public key` parameters into the collator's keystore. The node is stopped and finally restarted again.
It is possible to see that the keys were really added by querying `author_hasKeys`.

Keys were also added manually through `apps`  but the end result was the same.

scripts/parachain/local.sh
```diff
@@ -91,13 +90,22 @@ start_collator() {
...
 }
 
 # Feel free to comment, add or remove collators. Just remember that #Validators > #Collators
 
-start_collator --alice 30333 9933 9944 32000 8200 9200 &> /dev/null & node_pid=$!
-start_collator --bob 30334 9934 9945 32001 8201 9201
\ No newline at end of file
+start_collator 30333 9933 9944 32000 8200 9200 "swap ketchup reopen speak diary estate jealous ostrich crater useless jaguar attack" "0x7225b9e36cc33da17d7f97053c7110559f21967afc09867345367b7de241d03c"
```
c410-f3r, line 111: * "swap ketchup reopen speak diary estate jealous ostrich crater useless jaguar attack" is the account id seed
* "0x7225b9e36cc33da17d7f97053c7110559f21967afc09867345367b7de241d03c" is the session key

scripts/parachain/local.sh
```diff
@@ -59,29 +59,28 @@ start_validator() {
...
 
-  local relay_chain_port=$5
-  local relay_chain_rpc_port=$6
-  local relay_chain_ws_port=$7
+  local seed=$7
+  local public_key=$8
 
-  ../release/zeitgeist \
-    $author \
+  ../release/zeitgeist purge-chain --chain=./zeitgeist-raw.json -y
```
c410-f3r, line 77: This line can be commented to persist storage between multiple script calls

c410-f3r (commented):
Deploy is currently being tested using the `scripts/parachain/local.sh` script with the `battery_park_staging` specification instead of `local`.
Once we figure out how to solve this initial collator issue, all fixes will be ported to the actual deployment script, aka, `./scripts/parachain/rococo.sh`.

c410-f3r:
After bashing my head against the wall for several days alongside an infinite amount of debugging, I was finally able to produce blocks in a non-local environment. 

Capture d’écran de 2021-06-20 20-08-21

Tomorrow all changes will be pushed to the remote instance

lsaether:
Can you bump the [transaction version](https://crates.parity.io/sp_version/struct.RuntimeVersion.html#structfield.transaction_version) in the runtime?

c410-f3r:
@

> Can you bump the [transaction version](https://crates.parity.io/sp_version/struct.RuntimeVersion.html#structfield.transaction_version) in the runtime?

Done

Can I merge? A new docker image needs to be build in order to deploy

lsaether (approved):

lsaether:
Yep, you can merge once CI is green

c410-f3r:
Changes are live -> https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Froc.zeitgeist.pm#/explorer

PR #177, opened 2021-06-12
[`Swaps`] Replace macros with functions

c410-f3r:
This is something I was trying to do for a long time. Although powerful, procedural macros can sometimes be hard to read and maintain, which was usually the case for the macros on the swaps pallets. Beyond that, `Rustfmt` doesn't play well with macros and perhaps the most important of all, interactive code debugging is usually very painful.

Also re-introduces the new Rustfmt rules of https://github.com/zeitgeistpm/zeitgeist/pull/142/commits/cc51ec564a9d8f2dc5f9739072942a93e79c3110. I think it is an overall improvement but feel free to discuss any contradictions.


sea212 (approved):
Thanks, definitely makes working on/with the pallet a bit easier.

Review would have been a bit more straightforward if those were two PRs: One for general refractoring and one for refractoring of the macros in the swaps pallet.

lsaether (approved):

Issue #109, opened 2021-04-29
[Benchmark] Setup cloud instance for benchmarking


sea212:
With reference to #63, Logan suggested:

> Once we have set up the benchmarking, we can set up a cloud instance that we can run the benchmarks on in order to get a standard pricing.

sea212:
Some Options:

The benchmarks can be executed automatically as a whole, because our custom weight template generates the WeightInfo struct and the trait. We could setup a vm that runs a cronjob (for example) and listens to our github project. Whenever a change in the pallets occur, it could trigger a pull, build, execute a complete benchmark, test the modified project and afterwards push the freshly generated weight files or discard in case more recent changes were pushed. This can also be buffered in one additional branch, that temporarily carries merges into the main branch, such that the automatically generated weights are not pushed directly into main and can be reviewed before including them into main.
We should also always review the tests and benchmarks when the pallet logic changes. I would even go so far to prohibit merges into main before they have been verified. Otherwise, in the case of benchmarks, we could run into the danger to generate invalid weight functions.

The alternative to a fully automatic approach would be to manually trigger specific benchmarks on a reference machine, after verifying and potentially changing the benchmarks after an update of the pallet logic, ultimately including potential changes to the benchmarks itself and changes to the weights as an indicator of an executed verification

sea212:
Solved in #181

PR #181, opened 2021-06-15
[CICD] Add benchmark workflow

sea212:
**Issue**
#109 

**Solution**
A workflow is created that contains a bash script that executes benchmarks.

1. Watch for pushes on a *stage-benchmark* branch
2. Clone the latest state of the *stage-benchmark* branch
3. Run (optimized) runtime-benchmarks tests
4. Build optimized zeitgeist binary with runtime-benchmarks features
5. Run our benchmark script - For every benchmark command it:
   a.  Checks if the *stage-benchmark* branch contains new commits and if yes, aborts the benchmarks and does NOT commit changes
   b. Runs the benchmark
6. Checks the branch again for new commits and if none are existent, pushes the updated weights to the *stage-benchmark* branch

We can also run this script directly on the main branch, but I suggest that we use a staging branch to:
1. Verify that everything works appropriately (it was tested already, but who knows)
2. Squash the two commits, the commit that contains the PR merge and the "update weights" commit, into one commit and merge only one commit into the main branch. If we don't do this, every or almost every second commit on the main branch will be "Update weights".

This workflow can be used now, but I suggest to not use it due to the low hardware specs of the default runner and the uncertainty regarding the availability of the hardware resources. More info on #187 and #185. In addition, creating an action that is ready for the marketplace and public usage is also a nice-to-have #184.

You can inspect a [successfully executed benchmark workflow](https://github.com/zeitgeistpm/zeitgeist/runs/2848623725?check_suite_focus=true) in the past runs of this repository and its corresponding automatically committed new weights at commit hash [c6754e2cc608a2301d1f956f6eac8a4fe94f9a2e](https://github.com/zeitgeistpm/zeitgeist/pull/181/commits/c6754e2cc608a2301d1f956f6eac8a4fe94f9a2e)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,86 @@
...
+          done
+
+      - name: Commit updated weights
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Update weights
+          commit_options: '--no-verify'
+          file_pattern: zrml/*/src/weights.rs
+          commit_author: zeitgeist-benchmark-bot <zeitgeist-benchmark-bot@no-reply.zeitgeist.pm>
+          status_options: '--untracked-files=no'
```
c410-f3r, line 86: Cool

c410-f3r (approved):
Although it is still necessary to create benchmark methods for future pallet call additions (as well as their weights on the function signatures), these calls won't necessarily need to be benchmarked locally.

sea212:
> Although it is still necessary to create benchmark methods for future pallet call additions (as well as their weights on the function signatures), these calls won't necessarily need to be benchmarked locally.

Exactly. We still have to write the benchmarks, create a fast run such that the weights file is generated and include the weights in the pallet calls. After pushing to the *stage-benchmark* branch, the bot will commit reasonable weights.
If new pallets are added, we also have to add the associated benchmark command to the workflow.
It would be possible to automatically parse every existent pallet and generate benchmark commands automatically, but then we'd lose control about per-pallet benchmark *step* and *repeat* counts.

sea212:
We could also use a different workflow trigger, for example when a PR is declared as approved by a label. Alternatively we could introduce a new state "benchmark required" that the workflow will listen on and remove after benchmarks were executed and the new weights were pushed. There are numerous alternatives to using a staging branch. What do you think about that?

lsaether (approved):
LGTM

sea212:
@lsaether @c410-f3r What do you think about this:
> We could also use a different workflow trigger, for example when a PR is declared as approved by a label. Alternatively we could introduce a new state "benchmark required" that the workflow will listen on and remove after benchmarks were executed and the new weights were pushed. There are numerous alternatives to using a staging branch. What do you think about that?

Personally I like the idea to introduce new state that triggers the workflow.

lsaether:
I like this idea too, it would make the entire process more automated and feel more like magic 🪄

How easy is it to make a workflow trigger on a label though? If it's not so easy maybe what we can do is add the label, make it manual trigger, then the label would be replaced with a 'benchmark-complete' label when it's done.

sea212:
> How easy is it to make a workflow trigger on a label though? If it's not so easy maybe what we can do is add the label, make it manual trigger, then the label would be replaced with a 'benchmark-complete' label when it's done.

Okay, I will merge this now and open an issue that tackles this. It should be not too hard, since GitHub offers an event for this (pull_request:labeled).

Issue #186, opened 2021-06-17
[Runtime] Expose storage on the Aura pallet


lsaether:
https://github.com/zeitgeistpm/zeitgeist/blob/main/runtime/src/lib.rs#L183

Aura has storage, we should expose it.

sea212:
Solved in #189.

PR #189, opened 2021-06-18
Expose Aura storage

sea212:
Issue: #186


c410-f3r (approved):
Not particularly aware of the consequences (if any) of this change but in any case, it won't affect parachains.
Another issue should probably be created to figure out migration checks for parachains

PR #194, opened 2021-06-23
Bump crates versions to 0.1.1

lsaether:
Bumps all the versions from 0.1.0 to 0.1.1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11745,7 +11745,7 @@ dependencies = [
 
 [[package]]
 name = "zeitgeist-node"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "cumulus-client-cli",
  "cumulus-client-collator",
@@ -11807,7 +11807,7 @@ dependencies = [
 
 [[package]]
 name = "zeitgeist-primitives"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "arbitrary",
  "frame-support",
@@ -11823,7 +11823,7 @@ dependencies = [
 
 [[package]]
 name = "zeitgeist-runtime"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "cumulus-pallet-dmp-queue",
  "cumulus-pallet-parachain-system",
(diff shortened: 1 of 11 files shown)


lsaether:
Since we depend on the pallet versions when doing migrations, I think this means we will need to perform another runtime upgrade to ensure all pallets that were just bumped have the same version in the on-chain runtime.

This means we will need to bump the `spec_version` to 17.

@sea212 @c410-f3r thoughts?

sea212:
I think no runtime upgrade is required in this context for the following reason:
When a migration is written, the stored version (previous) version is compared with the current version (during the runtime upgrade). Then the version is stored in storage. Example from swaps:
```rust
        let version_0_1_0 = PalletVersion { major: 0, minor: 1, patch: 0 };
        let storage_version = <Pallet<T>>::storage_version().unwrap_or(version_0_1_0);

        if storage_version == version_0_1_0
```

Bumping the versions would only imply that existing migrations from other branches that were not merged into main yet would have to be adjusted to only migrate when the previous version was 0.1.1

Polkadot has another approach that prevents such issues: They remove old migrations before every release.

c410-f3r:
Taking into consideration that all these crates were modified, then a bump is welcome as a historical register of changes but I am not sure if these new versions will impact the storage migration.

lsaether:
Ok, I just want to not run into the situation that we falsely depend on storage version being `0.1.1` when it wasn't updated. Also it's not clear to me if the storage version follows the crate version automatically or if it needs to be manually set in a migration (which means we didn't set it in the migration either).

Still working on finding the best way to read the on-chain pallet storage version.

sea212:
The crate version is automatically set after a runtime upgrade (as far as I know even if the hook does not exist on the pallet).
The release strategy document suggest to bump every version once before a runtime upgrade, even if no changes occur. This way we can always check the previous unified version during migrations on a new release. This approach does not allow arbitrary runtime upgrade without bumping the versions, because the migrations check the previous pallet version. This way, if multiple runtime upgrades occur in between releases, the pallet will have the same version and therefore execute the migrations multiple times.

A more clean approach would be to bump the pallet versions independently (threat them like independent libraries) and only use a unified version number for the runtime and all crates that depend on it (e.g. node). Using our current way to check when a migration will be executed, this is not possible. We would have to check all versions of the pallet from the previous runtime upgrade and then write migrations for that specific versions. It would also deny anybody to perform an arbitrary runtime upgrade on their local chains or separate networks. Something different, for example like a classic dedicated storage value that references the storage version, is required for that.

PR #199, opened 2021-06-24
[Rikiddo] Implement Rikiddo helper functions

### LINKED ISSUE #159: [Rikiddo] Implement and test helper functions
Implement helper functions (EMA calculation, dynamic fee calculation, etc.) and implement tests for those.

sea212:
closes #159


zrml/rikiddo/src/types/ema_market_volume.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    fn calculate_ema(&mut self, volume: &TimestampedVolume<F>) -> Result<Option<F>, &'static str> {
+        // Overflow is impossible here (the library ensures that multiplier ∊ [0,1])
+        let volume_times_multiplier = if let Some(res) = volume.volume.checked_mul(self.multiplier)
+        {
+            res
+        } else {
+            return Err("[EmaMarketVolume] Overflow during calculation: volume * multiplier");
+        };
```
sea212, line 71: I decided to include overflow checks, even if they seem impossible to me, for the following reasons:
1) Although I am highly confident that those calculations cannot overflow, I still can be wrong. Therefore I prefer to initially leave them in the pallet, to avoid a node halt due to a panic. If the struct should be kept like this and was tested enough, we could replace them for some minor performance gains if desired.
2) The alternative to use saturated arithmetic does not seem the right way to me, because we would have no way to be informed about misbehavior of the functions except by seeing that the market behaves strange, and we would not have the information where exactly the problem occured.

sea212:
@lsaether @c410-f3r Just realized that the design is improvable a bit, please don't review yet, I'll push final updates within 1-2 hours.

sea212:
@lsaether @c410-f3r Ready for review now.

zrml/rikiddo/src/types.rs
```diff
@@ -1,109 +1,49 @@
...
+/*
 #[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
-pub struct LsdLmsrSigmoidMV<FI: FixedUnsigned, FE: LsdlmsrFee<FI>, MA: MarketAverage<FI>> {
+pub struct RikiddoSigmoidMV<FI: Fixed, FE: Sigmoid<FI>, MA: MarketAverage<FI>> {
     pub fees: FE,
     pub ma_short: MA,
     pub ma_long: MA,
     pub _marker: PhantomData<FI>,
 }
+*/
```
c410-f3r, line 49: If this snippet is not going to be used, then it is probably better to remove it

zrml/rikiddo/src/lib.rs
```diff
@@ -53,14 +53,14 @@ mod pallet {
...
     #[pallet::storage]
     pub type LmsrPerPool<T: Config> = StorageMap<
         _,
         Twox64Concat,
         u128,
-        LsdLmsrSigmoidMV<
+        RikiddoMV<
             FixedU128<T::FractionalType>,
             FeeSigmoid,
             EmaMarketVolume<FixedU128<T::FractionalType>>,
```
c410-f3r, line 66: More commented code. Guess they will be used in the future

zrml/rikiddo/src/types/sigmoid_fee.rs
```diff
@@ -0,0 +1,72 @@
+use crate::{
```
c410-f3r, line 1: Can you write a quick root doc (`//*`) about the sigmoid fee and its responsibility on Rikiddo?
I known about the white-paper and other documentations but something small with one or two lines would benefit other developers that are seeing this code for the first time

c410-f3r (approved):

zrml/rikiddo/src/types/ema_market_volume.rs
c410-f3r: Yes, better play safe than sorry

zrml/rikiddo/src/types.rs
sea212: This will be implemented in the next PR

zrml/rikiddo/src/lib.rs
sea212: This is correct, will be used in one PR after the next one.

zrml/rikiddo/src/types/sigmoid_fee.rs
sea212: A completed documentation will be added at #217 , is it okay from your perspective to leave it like this until then?

c410-f3r: OK. Didn't see the issue.

Issue #198, opened 2021-06-24
Add tagging to Docker image creation


lsaether:
Whenever we push a tag, we should create an identical tag for the Docker image to publish to Docker hub.

sea212:
Solved in #201.

PR #201, opened 2021-06-25
[CICD] Add tag trigger to docker deployment workflow

sea212:
Issue: #198

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -3,7 +3,8 @@ name: Docker Hub
 on:
   workflow_dispatch:
   push:
-    branches: [ main ]
+    branches: [ sea212-docker-push-on-tag ]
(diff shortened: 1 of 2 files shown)


sea212:
Was tested on commit c6ea147d46cfdb9a355e5d0f78fac2038c825bf2
When changes on the specified branch occur: https://github.com/zeitgeistpm/zeitgeist/runs/2914223229?check_suite_focus=true
When a tag is pushed: https://github.com/zeitgeistpm/zeitgeist/runs/2914260396?check_suite_focus=true
Manual trigger: https://github.com/zeitgeistpm/zeitgeist/runs/2914337471?check_suite_focus=true

Currently the workflow **triggers for any tag**. We can also make it trigger only when the tag has the format "vX.Y.Z". What do you prefer?

lsaether (approved):

PR #202, opened 2021-06-28
Add integration test for migrations

c410-f3r:
Depends on https://github.com/zeitgeistpm/tools/pull/105


scripts/runtime-upgrade/test_runtime_upgrade.sh
```diff
@@ -1,7 +1,24 @@
...
+sudo docker container stop migration
+sudo docker container rm migration
+cp -r /tmp/migration /tmp/migration-copy
+
+# Run new node with latest changes
+
+cargo build --bin zeitgeist --release
+./target/release/zeitgeist export-state --base-path /tmp/migration-copy --chain battery_park --pruning archive $BLOCK > /tmp/test-upgrade.json
+sed -i "s/\"$KEY_TO_SEARCH\": \".*\"/\"$KEY_TO_SEARCH\": \"$ALICE_PK\"/" /tmp/test-upgrade.json
+sed -i "s/\"0x5c0d1176a568c1f92944340dbfed9e9c530ebca703c85910e7164cb7d1c9e47b\": \".*\"/\"0x5c0d1176a568c1f92944340dbfed9e9c530ebca703c85910e7164cb7d1c9e47b\": \"0x04d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d\"/" /tmp/test-upgrade.json
```
lsaether, line 23: Can you make sure to also change the name, id, protocol id, and bootnodes so that this testing chain doesn't conflict with the actual Battery Park in any way?

.github/workflows/migration.yml
```diff
@@ -0,0 +1,24 @@
+name: Migration
+
+on:
+  workflow_dispatch:
+    inputs:
+      block:
+        description: 'Block number'    
+        required: true
```
c410-f3r, line 8: Adds a new manual job that requires the block number as input -> https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

integration-tests/tests/block.ts
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'chai';
+import SDK from "@zeitgeistpm/sdk";
```
c410-f3r, line 2: https://github.com/zeitgeistpm/tools/pull/105 was merged but it is cumbersome to release a new version.

@lsaether
```
npm install some_git_project#some_branch --save
```
Points to the root directory but not to `packages/sdk`. Not sure how to fetch a TS subdirectory using npm.

scripts/runtime-upgrade/test_runtime_upgrade.sh
```diff
@@ -1,7 +1,39 @@
...
+rm -rf /tmp/migration-copy
+
+# Run new node with latest changes
+
+cargo build --bin zeitgeist --release
+cp -r /tmp/migration /tmp/migration-copy
+./target/release/zeitgeist export-state --base-path /tmp/migration-copy --chain battery_park --pruning archive $BLOCK > /tmp/test-upgrade.json
+
+sed -i '/"bootNodes": \[/,/\]/c\ \ "bootNodes": [],' /tmp/test-upgrade.json
+sed -i "s/\"chainType\": \"Live\"/\"chainType\": \"Local\"/" /tmp/test-upgrade.json
```
c410-f3r, line 32: `Sudo` tab only shows when changing `chainType` to `Local`

scripts/runtime-upgrade/test_runtime_upgrade.sh
```diff
@@ -1,7 +1,39 @@
...
+
+BLOCK=$1
+ID=battery_park_runtime_upgrade_test
+PRODUCTION_IMAGE=zeitgeistpm/zeitgeist-node:latest
+
+# Sync with live network
+#
+# Comment this whole section if `/tmp/migration` is already synced.
+
+sudo docker run -d --name migration -v /tmp/migration:/zeitgeist/data $PRODUCTION_IMAGE --base-path /zeitgeist/data --chain battery_park --pruning archive
```
c410-f3r, line 19: After trying different types of combinations for days, the only way I could find to bind host's `/tmp/migration` to container's `/zeitgeist/data` was using `--privileged`.

Even with the usage of the same UID/GIG along side the correct folder/files permissions, Substrate will throw `Permission Denied`

c410-f3r (commented):
I was able to update the runtime but the storage was wiped out so this matter still need more investigation

scripts/runtime-upgrade/test_runtime_upgrade.sh
c410-f3r: Indeed. Thanks!

zrml/prediction-markets/src/migrations.rs
```diff
@@ -17,7 +17,7 @@ pub mod _0_1_1_move_storage_to_simple_disputes_and_market_commons {
     const MARKET_IDS_PER_REPORT_BLOCK: &[u8] = b"MarketIdsPerReportBlock";
     const MARKET_TO_SWAP_POOL: &[u8] = b"MarketToSwapPool";
     const MARKETS: &[u8] = b"Markets";
-    const PM: &[u8] = b"PredictionMarket";
+    const PM: &[u8] = b"PredictionMarkets";
```
c410-f3r, line 20: Runtime uses `PredictionMarkets` instead of `PredictionMarket`, thus the reason why migrations weren't working.
Capture d’écran de 2021-07-01 11-11-55

c410-f3r:
Still need to figure out why `PredictionMarkets` storage don't show up on the `Chain state` tab and actually create an automated integration test using `tools`

lsaether (approved):

PR #210, opened 2021-06-29
Use raw specs

c410-f3r:
The current parachain file specifications are using non-raw data, which leads to the already experienced non-deterministic WASM errors.


lsaether:
Can we remove `bp_parachain.json`?

c410-f3r:
> Can we remove `bp_parachain.json`?

Removed.

On node initialization, the application will panic telling the user that battery park shouldn't be used for parachains

lsaether (approved):

PR #211, opened 2021-06-30
Fix PM benchmarks

### LINKED ISSUE #195: Rearrange benchmarks taking into account the new `Court` pallet

c410-f3r:
Fixes #195


zrml/simple-disputes/src/lib.rs
```diff
@@ -302,14 +419,16 @@ mod pallet {
             }
 
             Self::deposit_event(Event::MarketDisputed(market_id, outcome));
-            Self::calculate_actual_weight(|_| 0, num_disputes as u32, max_disputes as u32)
+            calculate_actual_weight(|_| 0, num_disputes as u32, max_disputes as u32)
```
sea212, line 422: I am confused about this return value. It doesn't seem to be used anywhere, in fact the whole `on_dispute` function is only called during the setup of benchmarks in the PM pallet, in fuzz and in tests. Will this return value ever be used anywhere? If it will, it must be changed because currently it always implies a weight correction to 0 weight when `num_disputes != max_disputes`

sea212 (approved):

zrml/simple-disputes/src/lib.rs
c410-f3r: Hum... Totally missed this one, thanks for spotting it.

`on_dispute` was a PM call, thus the reason of `calculate_actual_weight`.

zrml/prediction-markets/src/lib.rs
```diff
@@ -279,6 +280,16 @@ mod pallet {
...
+        #[pallet::weight(T::WeightInfo::dispute(T::SimpleDisputes::max_disputes()))]
+        pub fn dispute(
+            origin: OriginFor<T>,
+            market_id: MarketIdOf<T>,
+            outcome: OutcomeReport,
+        ) -> DispatchResultWithPostInfo {
+            let [weight, max_weight] = T::SimpleDisputes::on_dispute(origin, market_id, outcome)?;
+            Self::calculate_actual_weight(&T::WeightInfo::dispute, weight, max_weight)
+        }
+
```
c410-f3r, line 292: @lsaether Re-adding this call until the actual court pallet is implemented

lsaether: I actually think we should keep this like this, then pattern match on the `DisputeMechanism` for the `market_id` and route the call to either `SimpleDisputes` or `Court` based on what's configured.

c410-f3r: Got it!

lsaether (approved):

PR #214, opened 2021-07-05
Update issue template

sea212:
Replace Polkadot-specific instructions with instructions for Zeitgeist.


sea212:
Split from wrong branch.

PR #218, opened 2021-07-06
[Rikiddo] Implement Rikiddo trait and tests - part 1

### LINKED ISSUE #160: [Rikiddo] Implement Rikiddo trait and tests for it
**TODOS**

- [ ] #218 
- [ ] #242

sea212:
This PR contains a partial implementation of the Rikiddo core functionallity and tests for it. Since the PR became too big already, it was split into two parts. 
This part contains the implementation of the `RikiddoMV` trait, the trait that exposes functions to add market data and clear it. It also contains the implementation of the Rikiddo cost function (which is exposed through the LMSR trait). The cost functions uses two strategies to calculate the result: One strategy simply calculates the cost function based on the unmodified cost formula (function `default_cost_strategy`). It requires some less operations than the other variant, but it can't calculate the result in many cases (due to overflow of intermediate exponential function results). If the Fixed point number type used to store the intermediate results of the exponential functions within the cost function does not offer enough bits to store the intermediate result, an alternative function (`optimized_cost_strategy`) is used that is able to calculate the result in any (relevant) case.
The next part of this split PR will offer an implementation of the price function.

### DIFF
--- a/zrml/rikiddo/src/types.rs
+++ b/zrml/rikiddo/src/types.rs
@@ -1,5 +1,5 @@
 use frame_support::dispatch::{fmt::Debug, Decode, Encode};
-use substrate_fixed::traits::Fixed;
+use substrate_fixed::traits::{Fixed, ToFixed};
 
 mod ema_market_volume;
 mod rikiddo_sigmoid_mv;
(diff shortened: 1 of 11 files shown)


zrml/rikiddo/src/types.rs
```diff
@@ -29,21 +33,67 @@ pub enum Timespan {
...
+    // We can safely cast because until here we know that the msb is not set.
+    let integer_part: TO = convert_common(num)?;
+    let fractional_part: FixedU128<U128> = num.frac().to_fixed();
+
+    if let Some(res) = integer_part.checked_add(fractional_part.lossy_into()) {
+        Ok(res)
+    } else {
+        // This error should be impossible to reach.
+        Err("Something went wrong during FixedSigned to FixedUnsigned type conversion")
+    }
```
sea212, line 98: Those can be potentially implemented in a more stylish fashion for all adequate Fixed types by using a blanked implementation. I guess it is fine like this, and the blanket implementation could be a low priority enhancement.

zrml/rikiddo/src/types/ema_market_volume.rs
```diff
@@ -198,32 +218,72 @@ impl<F: FixedUnsigned + From<u32>> MarketAverage<F> for EmaMarketVolume<F> {
...
+                        self.volumes_per_period = if let Some(res) =
+                            self.volumes_per_period.checked_mul(estimate_ratio_fu)
+                        {
+                            res.saturating_ceil()
+                        } else {
+                            return Err("[EmaMarketVolume] Overflow during estimation of \
+                                        transactions per period");
+                        }
+                    }
+
```
sea212, line 265: This is an addition I added after some discussion with Numa about what should happen if the short period ema value is already present, but the long ema period still needs to gather data. One possible option, which you see implemented here, is that we simply extrapolate the transaction count for the long period based on the data that's available. For example, if we have a short period ema of 1 hour, we can configure the long period ema struct to collect 6 hours of data, but extrapolate the transaction count per period after 1 hour of collecting data. This way, the ema ratio `r = ema_short / ema_long` is already available after the short ema period and can be used to calculate the sigmoid fee, instead of using the initial fee.

zrml/rikiddo/src/types/rikiddo_sigmoid_mv.rs
```diff
@@ -0,0 +1,368 @@
...
+            };
+
+        let required_bits_minus_one: u128 =
+            if let Some(res) = log2e_times_biggest_exp_plus_log2_num_assets.checked_ceil() {
+                res.to_num()
+            } else {
+                // Highly unlikely
+                return Err("[RikiddoSigmoidMV] Overflow during calculation: ceil(biggest_exp * \
+                            log2(e) + log2(num_assets))");
+            };
```
sea212, line 292: The code contains a couple of checks like this that are either impossible (from my perspective) or so incredibly unlikely, that one has to spend a great bunch of brain-power time to figure out a way to trigger them. Once the code ran for a longer period and we want to keep it like that, those can be removed in a PR that encompasses the optimization of this code.

c410-f3r (approved):

zrml/rikiddo/src/types.rs
lsaether: agreed

lsaether (approved):

PR #219, opened 2021-07-06
Update to Polkadot v0.9.6

c410-f3r:


primitives/src/constants.rs
```diff
@@ -14,23 +14,29 @@ pub const BLOCKS_PER_HOUR: BlockNumber = BLOCKS_PER_MINUTE * 60;
...
+parameter_types! {
+    pub const ExistentialDeposit: u128 = CENTS;
+    pub const MaxLocks: u32 = 50;
+    pub const MaxReserves: u32 = 50;
+}
 
 // Prediction Market parameters
 parameter_types! {
-    pub const AdvisoryBond: Balance = 25 * DOLLARS;
+    pub const AdvisoryBond: Balance = 25 * BASE;
```
lsaether, line 29: Why was this changed? It should stay low for the testnet.

primitives/src/constants.rs
```diff
@@ -14,23 +14,29 @@ pub const BLOCKS_PER_HOUR: BlockNumber = BLOCKS_PER_MINUTE * 60;
...
     pub const DisputeBond: Balance = 5 * BASE;
     pub const DisputeFactor: Balance = 2 * BASE;
     pub const DisputePeriod: BlockNumber = BLOCKS_PER_DAY;
     pub const MaxCategories: u16 = 10;
     pub const MaxDisputes: u16 = 6;
     pub const MinCategories: u16 = 2;
-    pub const OracleBond: Balance = 50 * DOLLARS;
+    pub const OracleBond: Balance = 50 * BASE;
```
lsaether, line 36: Same as above

primitives/src/constants.rs
```diff
@@ -14,23 +14,29 @@ pub const BLOCKS_PER_HOUR: BlockNumber = BLOCKS_PER_MINUTE * 60;
...
     pub const PmPalletId: PalletId = PalletId(*b"zge/pred");
     pub const ReportingPeriod: BlockNumber = BLOCKS_PER_DAY;
-    pub const ValidityBond: Balance = 50 * DOLLARS;
+    pub const ValidityBond: Balance = 50 * BASE;
```
lsaether, line 39: Same

lsaether:
Needs to bump the spec version

primitives/src/constants.rs
c410-f3r: My Bad. Fixed

lsaether:
Since `RangeInclusive` has changed, it needs to be change in `types.json` as well. https://doc.rust-lang.org/nightly/src/core/ops/range.rs.html#341-357

c410-f3r:
> Since `RangeInclusive` has changed, it needs to be added/ changed in `types.json` as well. https://doc.rust-lang.org/nightly/src/core/ops/range.rs.html#341-357
> 
> Also, since we expose `RangeInclusive` as an extrinsic argument, any idea how this works on Polkadot-JS Apps or with the JS layer? I guess we use an object like `{ start: 0, end: 10, exhausted: false }` ? Or can we create an array `[0, ... 10]` ?

Support for `RangeInclusive` was added to `parity-scale-codec` but `apps` still doesn't know about `RangeInclusive`. I am reverting back to `(u128, u128)` for now because changing to `RangeInclusive` also implicates a migration

lsaether:
Ok, makes sense. Could you make an issue for this change though? since it seems like something that would be reasonable to do in the future but is lower priority.

c410-f3r:
> Ok, makes sense. Could you make an issue for this change though? since it seems like something that would be reasonable to do in the future but is lower priority.

Done -> https://github.com/zeitgeistpm/zeitgeist/issues/220

lsaether:
@sea212 Can you do a quick review on this?

sea212:
> @sea212 Can you do a quick review on this?

Sure, I'll take a look at it.

scripts/parachain/local.sh
```diff
@@ -6,9 +6,9 @@
 
 set -euxo pipefail
 
-PARACHAIN_CHAIN=battery_park_staging
+PARACHAIN_CHAIN=battery_station_staging
 PARACHAIN_ID=9123
-POLKADOT_BRANCH=release-v0.9.4
+POLKADOT_BRANCH=release-v0.9.7
```
sea212, line 11: Should this be v0.9.7?

node/src/service/service_parachain.rs
```diff
@@ -65,7 +65,7 @@ pub fn new_partial(
         config.transaction_pool.clone(),
         config.role.is_authority().into(),
         config.prometheus_registry(),
-        task_manager.spawn_handle(),
+        task_manager.spawn_essential_handle(),
```
sea212, line 68: Does this prevent a client from further operating a tx pool when an error occurs, i.e. the tx pool is shut down (and therefore the client stops operating, at least this service)?

sea212 (approved):
Looks good to me. 
The currency adjustments are also appropriate and the `ExistentialDeposit` was also adjusted accordingly, therefore nothing really but the names for the magnitudes of currency should change.

node/src/service/service_parachain.rs
c410-f3r: I don't know the difference between an ordinary and an essential task. This code was simply copied and pasted from the `0.9.6` cumulus  branch.
https://core.tetcoin.org/rustdocs/v2.0.1/sc_service/struct.TaskManager.html does not state the difference and https://github.com/paritytech/cumulus/pull/496 does not tell why `spawn_handle` was replaced.
A spawned task is a separated unit of computation (`T: Send`) that lives with its own context and I don't know a failed task will prevent the client from further operating a tx pool. If you see https://github.com/paritytech/substrate/blob/4d9f03d3d6d0ceec0708b1a23978feee12f5b0bb/client/service/src/task_manager/mod.rs#L149-L157, then it might be possible that `spawn_essential_handle` will improve the robustness of the node.

scripts/parachain/local.sh
c410-f3r: `v0.9.7` because Rococo is probably using this version

sea212: But this is the local environment, isn't it? If I understand correctly, we launch the Rococo chain locally "rococo-local". This way or another, I don't mind if it works properly.

c410-f3r: The branch version is mirroring a production environment the same way `PARACHAIN_CHAIN` is `battery_station_staging` instead of `dev`.

Changed back to 0.9.6

sea212: Thanks for the clarification.

Issue #200, opened 2021-06-25
Court Pallet


lsaether:
To live alongside `SimpleDisputes`, there is the `Court` pallet.

## Core Logic

- `join_court` - Allows any ZTG holder to stake funds to join the court. The amount of funds to stake is a factor of the number of court participants that have already joined.
- `exit_court` - Allows a ZTG holder who has staked to join the court to leave and get their funds back. 
- `on_dispute` - When a dispute occurs a subset of all court participants is chosen at random and assigned to a "court case". They then have `CourtCaseDuration` amount of time to cast a vote for which `Outcome` is correct. The majority of votes becomes the new reported outcome. The dispute process can occur multiple times, each requiring a larger bond, and taking a larger subset of court participants at random.
- `on_resolution` - Similar to `SimpleDisputes` resolution, however with a different payout structure. All of the parties who are sided on the winning side get their bonds back, and the parties on the losing side are slashed 50% of their bonds, which get split among all parties on the winning side. For Court participants who voted on the losing side in a winning majority or failed to vote within the `CourtCaseDuration` will not receive any rewards, however neither will they be slashed, they will however be placed as `Tardy`. If a `Tardy` court participant fails to participate in a court case for a second time, they are removed from the court and 20% of their staked funds are sent to the on-chain treasury.

## Storage

Similar storage to what's already in simple disputes, with the addition of...

`Jurors` - a `vec` of all staked jurors.

### Juror

```rust
struct Juror {
  bonded: u128,
  status: JurorStatus,
};

enum JurorStatus {
  Ok,
  Tardy,
};
```

Issue #227, opened 2021-07-12
[PM] Attach the dispute mechanism to the market


lsaether:
Markets should have 3 selectable dispute mechanisms:

- Authorized(AccountId) - The market is centrally decided upon.
- SImpleDisputes - The market uses the `SimpleDisputes` pallet.
- Court - The market uses the `Court` pallet.

PR #231, opened 2021-07-15
Fourth iteration of the Court pallet

c410-f3r:
CC #200
Depends on #225


zrml/simple-disputes/src/tests.rs
```diff
@@ -1,128 +0,0 @@
-#![cfg(test)]
```
c410-f3r, line 1: After some thought, it was clear that `simple-disputes` tests are a subset of the `PM` markets and the same would apply to `court`. Therefore, all individual pallet tests were transferred to the `PM` pallet.
Since `simple-disputes` is currently the only used "back-end", no dispute mechanism switch was added (https://github.com/zeitgeistpm/zeitgeist/issues/227).

zrml/prediction-markets/src/lib.rs
```diff
@@ -717,6 +735,14 @@ mod pallet {
 
         type ApprovalOrigin: EnsureOrigin<<Self as frame_system::Config>::Origin>;
 
+        /// The base amount of currency that must be bonded in order to create a dispute.
+        type DisputeBond: Get<BalanceOf<Self>>;
+
+        /// The additional amount of currency that must be bonded when creating a subsequent
+        /// dispute.
+        type DisputeFactor: Get<BalanceOf<Self>>;
```
c410-f3r, line 743: Moving back these parameters from `simple-disputes` to `PM` because they are both used by `simple-disputes` and `court` "back-ends". This feature will avoid duplication across different pallets

zrml/court/src/lib.rs
```diff
@@ -20,44 +18,59 @@ mod pallet {
...
     use frame_system::{ensure_signed, pallet_prelude::OriginFor};
-    use sp_runtime::DispatchError;
+    use sp_runtime::{traits::Saturating, ArithmeticError, DispatchError, SaturatedConversion};
+    use zeitgeist_primitives::{
+        traits::DisputeApi,
+        types::{Market, OutcomeReport, ResolutionCounters},
+    };
     use zrml_market_commons::MarketCommonsPalletApi;
 
+    const MAX_RANDOM_JURORS: usize = 5;
```
c410-f3r, line 32: The maximum capped number of random jurors for a dispute. IMHO, it makes sense the set this value as a constant that can be changes through runtime updates.

lsaether: Yep, this makes sense, although I would set it to a larger number like 13 so that we can have a number of escalations before making it become a chain-wide vote. It could start with a group of 5 jurors, then 7, then 9, 11, 13.

zrml/court/src/lib.rs
```diff
@@ -20,44 +18,59 @@ mod pallet {
...
     pub(crate) type BalanceOf<T> =
         <CurrencyOf<T> as Currency<<T as frame_system::Config>::AccountId>>::Balance;
     pub(crate) type CurrencyOf<T> =
         <<T as Config>::MarketCommons as MarketCommonsPalletApi>::Currency;
+    pub(crate) type MarketIdOf<T> =
+        <<T as Config>::MarketCommons as MarketCommonsPalletApi>::MarketId;
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        #[frame_support::transactional]
```
lsaether, line 43: Does this need to be marked `transactional`?

zrml/court/src/lib.rs
```diff
@@ -91,12 +110,69 @@ mod pallet {
...
+        where
+            D: Fn(usize) -> Self::Balance,
+        {
+            CurrencyOf::<T>::reserve(&who, dispute_bond(1))?;
+            let max_random_jurors = Jurors::<T>::iter().count().max(MAX_RANDOM_JURORS);
+            let random_usize = Self::random_usize();
+            let jurors_len_opt = max_random_jurors.checked_rem(random_usize);
+            let fun = || ArithmeticError::Underflow.into();
+            let _jurors_len = jurors_len_opt.ok_or_else::<DispatchError, _>(fun)?;
+            Ok(())
```
lsaether, line 164: We should generate randomness to determine each index of the Juror that is selected. So the actual composition of the jury is random.

zrml/court/src/lib.rs
c410-f3r: Nope :)

zrml/court/src/lib.rs
c410-f3r: Done. Take a look at `random_jurors`

zrml/court/src/lib.rs
c410-f3r: This PR is already too big so I opened https://github.com/zeitgeistpm/zeitgeist/issues/232. Such feature should be resolved in the next PR

sea212:
Review started

zrml/court/src/lib.rs
```diff
@@ -122,20 +164,29 @@ mod pallet {
...
-        ) -> Result<[u32; 2], DispatchError> {
-            todo!()
+            who: Self::AccountId,
+        ) -> DispatchResult
+        where
+            D: Fn(usize) -> Self::Balance,
+        {
+            CurrencyOf::<T>::reserve(&who, dispute_bond(1))?;
+            let jurors: Vec<_> = Jurors::<T>::iter().collect();
+            let _ = Self::random_jurors(&jurors, 3);
```
sea212, line 183: Can we make `3` a constant or value within the pallet's config?

zrml/court/src/lib.rs
```diff
@@ -122,20 +164,29 @@ mod pallet {
...
+            Ok(())
         }
 
-        fn on_resolution<F>(_now: Self::BlockNumber, _cb: F) -> DispatchResult
+        fn on_resolution<D, F>(_dispute_bond: D, _now: Self::BlockNumber, _cb: F) -> DispatchResult
         where
+            D: Fn(usize) -> Self::Balance,
             F: FnMut(&Market<Self::AccountId, Self::BlockNumber>, ResolutionCounters),
         {
             todo!()
```
sea212, line 192: I know the pallet is not integrated yet, but IMO returning an error is favorable in contrast to having the system panic.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1012,35 +1048,87 @@ mod pallet {
...
                 MarketEnd::Timestamp(timestamp) => {
                     let now = T::Timestamp::now();
-                    now < timestamp.saturated_into()
+                    ensure!(now < timestamp.saturated_into(), Error::<T>::EndTimestampTooSoon);
+                }
+            };
+            Ok(())
+        }
+
+        fn ensure_disputes_does_not_exceed_max_disputes(num_disputes: u32) -> DispatchResult {
```
sea212, line 1065: #don't trust the compiler :smile: 
```suggestion
        #[inline]
        fn ensure_disputes_does_not_exceed_max_disputes(num_disputes: u32) -> DispatchResult {
```

sea212 (changes requested):

primitives/src/traits/dispute_api.rs
```diff
@@ -4,19 +4,24 @@ use frame_support::dispatch::{DispatchError, DispatchResult};
...
+        who: Self::AccountId,
+    ) -> Result<(), DispatchError>
+    where
+        D: Fn(usize) -> Self::Balance;
 
     /// Manages markets resolutions moving all reported markets to resolved.
-    fn on_resolution<F>(now: Self::BlockNumber, cb: F) -> DispatchResult
+    fn on_resolution<D, F>(dispute_bound: D, now: Self::BlockNumber, cb: F) -> DispatchResult
     where
+        D: Clone + Fn(usize) -> Self::Balance,
```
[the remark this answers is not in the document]
c410-f3r: Done

zrml/court/src/lib.rs
c410-f3r: This matter is already addressed on https://github.com/zeitgeistpm/zeitgeist/pull/233

zrml/court/src/lib.rs
```diff
@@ -104,6 +112,40 @@ mod pallet {
...
+                // `idx` will always be within the length of `jurors`
+                let (_, juror) = jurors.get(idx).unwrap();
+                let juror_is_not_included = subset.iter().all(|&el| el != juror);
+                if juror_is_not_included {
+                    // `push` will never overflow the internal capacity of `MAX_RANDOM_JURORS` jurors
+                    subset.push(juror);
+                }
+            }
+            subset
+        }
```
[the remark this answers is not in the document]
c410-f3r: Although it is trivial to change the consensus of a standalone node, parachains use the nimbus consensus that is based on Aura.

If `not secure` in this context means a predictable seed then yes, the randomness is not safe as anyone could predict the next jurors of a block. IMO (although I could be wrong), this feature is not a problem in this context.

No-one is using the function and the pallet is not attached to the runtime, therefore, this proof-of-concept implementation can be merged and then more advanced random seeds can be researched.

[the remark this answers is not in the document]
sea212: With `not secure` I was referring to what is stated at https://substrate.dev/rustdocs/v3.0.0/src/pallet_randomness_collective_flip/lib.rs.html#106-108

> Our low-influence function reduces the actual block producer's influence over the randomness, but increases the influence of small colluding groups of recent block producers.

I see that if it is a requirement of a parachain to use Aura, then we have little to no control over this. If you see any way in the future to use Nimbus with Babe or Babe in any other way in a parachain, using it is definitely preferred.

> c410-f3r: No-one is using the function and the pallet is not attached to the runtime, therefore, this proof-of-concept implementation can be merged and then more advanced random seeds can be researched.

I think it is not a matter of how advanced of a random seed it is, the current implementation is simply ignoring the [best practices](https://substrate.dev/recipes/randomness.html#random). It leads to a critical bug and can simply be avoided by using one of the strategies described in the previous reference, of which one is already suggested. I agree with you that it is not really necessary right now and has no impact (since it's not "active"), but from my perspective leaving this kind of a serious bug, where the solution is quite simple and fast done, instead of immediately fixing it is an unnecessary risk and just demands for further tracking. Whenever you decide to fix it, please keep it as a high priority.

zrml/court/src/tests.rs
```diff
@@ -63,3 +67,43 @@ fn join_court_will_not_insert_an_already_stored_juror() {
...
+        ];
+        let random_jurors = Court::random_jurors(&jurors, 2);
+        let mut at_least_one_is_different = false;
+        for n in 0..1000 {
+            setup_blocks(START_BLOCK..START_BLOCK + n);
+            if random_jurors != Court::random_jurors(&jurors, 2) {
+                at_least_one_is_different = true;
+                break;
+            }
+        }
```
sea212, line 93: This for loop is unable to detect whether the two arrays contains different jurors, it can only check if the complete arrays differ. If you run this test with a println!(...) command in the current state, that prints both arrays for every iteration, and additionally comment out the break statement, you will get an output like this:
> [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 1, status: Ok }, Juror { staked: 2, status: Tardy }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }]
[Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }] - [Juror { staked: 2, status: Tardy }, Juror { staked: 1, status: Ok }]

Cleary the test checks whether the arrays differ, but not whether different jurors were selected.
```suggestion
        for n in 0..1000 {
            setup_blocks(START_BLOCK..START_BLOCK + n);
            rand_jurs = Court::random_jurors(&jurors, 2);
            for juror in &random_jurors {
                at_least_one_is_different = rand_jurs.iter().all(|&el| el != *juror);
                if at_least_one_is_different { return };
            };
        }
```

zrml/court/src/lib.rs
```diff
@@ -104,6 +112,40 @@ mod pallet {
...
+                // `actual_len` will never be 0
+                let idx = Into::<usize>::into(byte) % actual_len;
+                // `idx` will always be within the length of `jurors`
+                let (_, juror) = jurors.get(idx).unwrap();
+                let juror_is_not_included = subset.iter().all(|&el| el != juror);
+                if juror_is_not_included {
+                    // `push` will never overflow the internal capacity of `MAX_RANDOM_JURORS` jurors
+                    subset.push(juror);
+                }
+            }
```
sea212, line 145: This for loop will always select between the first `actual_len` of jurors. If you have for example jurors = [A, B, C, D], and `actual_len = 2`, then it will always select jurors `A` and juror `B`, irrespective of how often and when you will run this function. This is due to the circumstance, that `actual_len` is used for the modulo operation. The following suggestion replaces `actual_len` and adds a break at the end when enough jurors were collected:
```suggestion
            for byte in random_hash.as_ref().iter().copied() {
                if subset.len() == MAX_RANDOM_JURORS {
                    break;
                }
                // `actual_len` will never be 0
                let idx = Into::<usize>::into(byte) % jurors.len();
                // `idx` will always be within the length of `jurors`
                let (_, juror) = jurors.get(idx).unwrap();
                let juror_is_not_included = subset.iter().all(|&el| el != juror);
                if juror_is_not_included {
                    // `push` will never overflow the internal capacity of `MAX_RANDOM_JURORS` jurors
                    subset.push(juror);
                }
                if subset.len() == actual_len { break };
            }
```

The solution above does fix the random selection over the whole set of jurors, but it is not complete at all and must be further improved. Three things have to be considered:
1) If `jurors.len() > 255`, then every juror after index 255 has no chance of being selected (since `a % b = b` for every `b >= a`)
2) The chances are unlikely distributed. Example: We have 120 jurors (this works also with less jurors, but it's easy to see with 120 jurors). If the random byte is between 0 and 119, every juror can be selected. If the random byte is between 120 and 239, every juroro can be selected. If the random byte is between 240 and 255, only the first 15 jurors can be selected! This means, the first fifteen jurors have a higher likelyhood to be selected as jurors and will be present in more courts than the other jurors in the long term.
3) If there are more than `byte_count(random_hash)` jurors to be selected, then this function will cap out at max `byte_count(random_hash)` jurors. For example, if we want 260 jurors, max 256 jurors will be selected and 4 will be left out. Even worse: If we want 120 jurors for example, and the loop fails to find 120 distinguished jurors after iterating over every random byte, the subset will contain less than 120 jurors. A solution to that would be to generate new randomness and start another loop to find the remaining jurors.

sea212: One simple approach towards ensuring almost the same likelyhood for every juror to be selected, is to use the whole random number and execute modulo on it. The likelyhood will still be uneven, but the difference is so incredibly small, that practically this effect will not be observable.

c410-f3r:
@sea212 The majority of your concerns were addressed as requested

zrml/court/src/lib.rs
```diff
@@ -104,6 +113,39 @@ mod pallet {
...
+                rslt
+            });
+            let mut rslt: u64 = nonce;
+            // https://github.com/paritytech/substrate/issues/8312
+            let (random_hash, _) = T::Random::random(b"zrml-court");
+            for byte in random_hash.as_ref().iter().copied() {
+                rslt = rslt.wrapping_add(byte.into());
+            }
+            StdRng::seed_from_u64(rslt)
+        }
```
sea212, line 147: Thanks Caio for incorporating all the suggestions. The usage of the randomness and distribution of jurors throughout time should be increased significantly now. I have one last remark.

**TL;DR**
Adding the bytes reduces the entropy significantly. The best way to keep as much entropy as possible is to just pass the raw byte array that is returned from the ``T::Random::random`` function. Alternatively, concatenating the bytes to a number does also maintain entropy, the only drawback being that we don't have an integer type which can contain 32 bytes of data. The following solution maintains the initial entropy (up to 32 bytes, the most common length used for the `Hash` type) and passes it to ``StdRng``:
```suggestion
        pub(crate) fn rng() -> impl RngCore {
            let nonce = <JurorsSelectionNonce<T>>::mutate(|n| {
                let rslt = *n;
                *n = n.wrapping_add(1);
                rslt
            });
            // https://github.com/paritytech/substrate/issues/8312
            let (random_hash, _) = T::Random::random(&nonce.to_le_bytes());
            let mut seed: [u8; 32] = [0; 32];

            for (idx, byte) in random_hash.as_ref().iter().enumerate() {
                if idx == seed.len() {
                    break;
                } 

                seed[idx] = *byte;
            }

            StdRng::from_seed(seed)
        }
```

**Why is entropy reduced during addition?**
To understand this, we will use the definition of Entropy (from information theory) by Claude Shannon:
Entropy

Imagine you have one fair 6-sided dice. The probability of each number to appear after a fair throw of the dice is 1/6. Using this formula, we get the result `H(X) = -log2(1/6) = 2.584962501 Sh, in this case 2.584962501 bits`.

Now imagine you have two fair 6-sided dices. We throw the dices and concatenate the results, i.e. `result = d1 || d2`. What is the entropy? Concatenating leads to 36 different possible outcomes, each with a probability of 1/36, because we can reach every 36 outcomes only by one combination of both dices. This leads to an entropy of:
`H(X) = -log2(1/36) = 5.169925001 bits`.

Now imagine we use the add operation to calculate the result, i.e `result = d1 + d2`. When you play that through, you'll see that there is only one option for the number 2 and 12, but 6 options for the number 7 for example. Obviously the total possible number is lower and also the distribution is uneven.
What you get is this distribution of probabilities for the potential outcomes:
image
Now if we apply Shannons formula for entropy in this scenario, we will get (shortened, because the formula is quite long):
`H(X) ≈ 3.27 bits`.

Clearly we reduced the entropy by using the add operation. Therefore we also reduced the quality of the random numbers the RNGs can calculate and consequently also the distributions of the jurors.

sea212 (commented):
Thank you very much for tackling those issues. Regarding the random selection of jurors, I can now sleep a little better at night. Would be nice if you could incorporate the final suggestion.

zrml/court/src/lib.rs
c410-f3r: Great explanation! Thanks!

sea212: Thank you. Beautiful way to realize the byte copy you implemented there!

sea212 (approved):

PR #234, opened 2021-07-21
Battery station deployment

c410-f3r:
One parachain (bsr.zeitgeist.pm) and two validators were already deployed  but I still need to figure out what else remains to launch our relay chain (probably a light validator node for `apps`).
If more ppl will run parachains, then they will also need to deploy validators in conjunction (at least two).


c410-f3r:
On the Polkadot side, `rococo` was renamed to `battery-station-relay`, i.e., the runtime is the same but only the name, protocol and id of the chain specification were modified along side new custom address.
Capture d’écran de 2021-07-25 09-28-11
Even though parachains aren't working, local validators using the `battery-station-relay` specification are at least producing blocks.

In short, I still need to figure out what the missing pieces to deploy a relay chain and make it work with our parachain

c410-f3r:
In a local network both parachain and relay chain are now producing blocks. The only thing remaining is the construction of the first block of the relay chain in a live network, which probably means inserting all or most session keys of Rococo.

```rust
impl_opaque_keys! {
	pub struct SessionKeys {
		pub grandpa: Grandpa,
		pub babe: Babe,
		pub im_online: ImOnline,
		pub para_validator: Initializer,
		pub para_assignment: SessionInfo,
		pub authority_discovery: AuthorityDiscovery,
		pub beefy: Beefy,
	}
}
```

Battery station relay chain can be accessed at https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fbsr.zeitgeist.pm

c410-f3r:
The relay chain is producing blocks in a live environment and the Zeitgeist parachain was successfully included. Now the only thing remaining is wait for the `0.9.9` release of Polkadot to fix the following `Collation wasn't advertised to any validator` issue (https://github.com/paritytech/polkadot/issues/3139) to produce the genesis parachain block.

Capture d’écran de 2021-08-06 11-53-51

* I already tried the `--no-beefy` approach to no avail.

c410-f3r:
OK, both relay chain and parachain are now producing blocks. In the end, the parachain genesis block stuff wasn't a thing related to https://github.com/paritytech/polkadot/issues/3139.

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fbsr.zeitgeist.pm#/explorer

c410-f3r:
All the changes of this PR are already deployed so I will merge

c410-f3r:
After some hours, a new epoch started and the parachain was automatically downgraded to a parathread.

Capture d’écran de 2021-08-14 16-57-17

Probably related to some Rococo runtime parameter, I guess....

PR #237, opened 2021-07-22
[Runtime] Add utility pallet

### LINKED ISSUE #229: [Runtime] Add the `Utility` pallet from FRAME
We need the Susbtrate FRAME standard `Utility` pallet to make batch transactions.

sea212:
closes #229

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -53,7 +53,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("zeitgeist"),
     impl_name: create_runtime_str!("zeitgeist"),
     authoring_version: 1,
-    spec_version: 19,
+    // major_minor_patch_spec-version
+    spec_version: 20,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 3,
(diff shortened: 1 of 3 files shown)


runtime/src/lib.rs
```diff
@@ -552,6 +560,7 @@ impl_runtime_apis! {
             add_benchmark!(params, batches, frame_system, SystemBench::<Runtime>);
             add_benchmark!(params, batches, pallet_balances, Balances);
             add_benchmark!(params, batches, pallet_timestamp, Timestamp);
+            add_benchmark!(params, batches, pallet_utility, Utility);
```
sea212, line 563: I wonder if this is really needed. How can we benchmark the utility pallet on our reference machine, without having to keep a copy of the pallet and update it manually or even fork the whole substrate repository and benchmark all necessary pallets regularly? AFAIK, we have to replace the `weights.rs` info file that contains the `WeightInfo` struct, and I don't see how this is possible without maintaining a copy of the pallet or substrate. In that case, is it really necessary to include the benchmark and bloat the runtime (that's built with *--feature runtime-benchmarks*) with code that never will be executed?

lsaether: Maybe the easiest way to handle this issue is to make our standard hardware aligned with Polkadot's so that we can use the Substrate benchmarks and weights. Otherwise it's inescapable to re-benchmark them with our standard machine afaict.

PR #238, opened 2021-07-22
[PM] Allow ApprovalOrigin to report market

### LINKED ISSUE #182: [PM] Allow Root origin to report an outcome within the first 24 hours
Currently only the assigned oracle account can report a result to a market within the first 24 hours. We should allow Root origin to make a report during this time as well.

sea212:
closes #182


lsaether (approved):
Thanks!

PR #242, opened 2021-07-23
[Rikiddo] Implement Rikiddo trait and tests - part 2

### LINKED ISSUE #160: [Rikiddo] Implement Rikiddo trait and tests for it

sea212:
This PR brings some more fine granular structure into the tests and implements the Rikiddo `price` and `all_prices` functions and tests for it. 
The `cost` function now calls a helper function, which can be configured to additionally store intermediate results and also is able to leave out a final multiplication. This is due to the circumstance, that the cost function is also a part of the price function, minus the additional final multiplication. In fact, the cost function contains almost every intermediate result that is needed to calculate the price function, effectively eliminating redundant calculation in case of reuse.
The intermediate results are stored within the `RikiddoFormulaComponents` struct.  The effect is even more amplified, when the `all_prices` function is used, instead of sequentially calling `price` for every asset. For example, in addition to the reuse of intermediate results, conversions between types also only have to be executed once.

Some of the intermediate results are stored using `hashbrown::HashMap`. [Using a hashmap is disencouraged by Parity](https://substrate.dev/docs/en/knowledgebase/advanced/no-hash-collections), but in this specific scenario, some properties are given which legitimate a secure usage of this function: The HashMap is temporary (per calculation of the price), the amount of entries limited and insensitive to user inputs.

closes #160


sea212:
Depends on merge of #218

lsaether:
@sea212 #218 has been merged so we'll need to fix conflicts here

lsaether (approved):
LGTM

PR #245, opened 2021-07-29
[Rikiddo] Implement Rikiddo pallet

### LINKED ISSUE #216: [Rikiddo] Implement Rikiddo pallet

sea212:
The Rikiddo pallet is solely a book keeper of `Rikiddo` instances per pool and a translator between the `Balance` type used within the pallets and the `Fixed` type. It exposes all the functions the Rikiddo core also offers in addition to a instantiation function `create`.
The basic procedure for most of the the pallet functions looks like this: Try to find the `Rikiddo` instance associated with the `poolid` that was passed, translate any `Balance` type to the configured `Fixed` type, execute the requested Rikiddo core function and convert the result back from the configured `Fixed` type to the `Balance` type. The pallet does not offer any dispatchable functions - it is ought to be used within another pallet to update the market data, fetch the fee and calculate the prices. In our case the swaps pallet will use the Rikiddo pallet for the previously mentioned functions.
The pallet followed the generic and modular approach of the Rikiddo core: Within it's `Config`, the maintainers can specify which variant of Rikiddo exactly should be used: How should the fee be calculated? How should the market volume be processed? Which exact implemention of Rikiddo should be used? All those properties can be configured within the pallet's `Config` trait. The pallet is instantiable, which enables a simultaneous usage of numerous Rikiddo implementations. Some example use-cases for this could be:

1. Introducing an improved Rikiddo variant and still being able to maintain old markets with the previous Rikiddo variant, while using the new Rikiddo variant for markets that are created after the new Rikiddo variant was introduced. Once all old markets that use the deprecated Rikiddo variant are closed, the old Rikiddo variant can be removed in an upcoming runtime upgrade.
2. Simultaneous usage of multiple Rikiddo variants, for example to gather data of multiple variants and compare this data for insights and improvements or to use different variants for different use-cases.

closes #216

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12146,6 +12146,7 @@ dependencies = [
  "pallet-balances",
  "pallet-timestamp",
  "parity-scale-codec",
+ "sp-core",
  "sp-io",
  "sp-runtime",
  "substrate-fixed",
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -15,7 +15,10 @@ use sc_telemetry::TelemetryEndpoints;
 use sp_core::{crypto::UncheckedInto, Pair, Public};
 use sp_runtime::traits::{IdentifyAccount, Verify};
 use zeitgeist_primitives::{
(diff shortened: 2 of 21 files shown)


sea212:
Depends on #242

lsaether (approved):

Issue #243, opened 2021-07-27
[Orderbook-v1] Implement fuzz targets


sea212:
Pallet *orderbook-v1* still needs fuzz targets to test its stability. Fuzz testing with *cargo-fuzz* is described in the [Rust Fuzz Book](https://rust-fuzz.github.io/book/cargo-fuzz.html).
The pallets *swaps* and *prediction-markets* already implement one fuzz target to fuzz test them, which can serve as an example.

Implement at least one fuzz target that covers the possible logical paths as good as possible. For example, create a bid on one side and fulfill it at the other side. As a bonus, move the different logical paths into distinguished fuzz targets.

Add the commands to execute the fuzz targets to *scripts/tests/fuzz.sh*, once all targets are implemented.

zachanon:
Hello and happy Saturday, or maybe Samstag. I would like to work on this issue if possible.

sea212:
Solved in #247

PR #247, opened 2021-08-01
[Orderbook-v1] Implement fuzz targets

zachanon:
PR for [#243](https://github.com/zeitgeistpm/zeitgeist/issues/243)


zachanon:
I've implemented two paths. One makes and then fills orders, another makes and then cancels them. When running I get a panic,

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /home/zach/projects/zeitgeist/zrml/orderbook-v1/src/lib.rs:339:80
stack backtrace:
   0: rust_begin_unwind
             at /rustc/3a5d45f68cadc8fff4fbb557780f92b403b19c19/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/3a5d45f68cadc8fff4fbb557780f92b403b19c19/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/3a5d45f68cadc8fff4fbb557780f92b403b19c19/library/core/src/panicking.rs:50:5
   3: zrml_orderbook_v1::pallet::Pallet<T>::make_order
   4: orderbook_v1_full_workflow::rust_fuzzer_test_input::{{closure}}
   5: environmental::using::{{closure}}
   6: sp_state_machine::testing::TestExternalities<H,N>::execute_with
   7: rust_fuzzer_test_input
   8: __rust_try
   9: LLVMFuzzerTestOneInput
  10: _ZN6fuzzer6Fuzzer15ExecuteCallbackEPKhm
  11: _ZN6fuzzer6Fuzzer6RunOneEPKhmbPNS_9InputInfoEbPb
  12: _ZN6fuzzer6Fuzzer16MutateAndTestOneEv
  13: _ZN6fuzzer6Fuzzer4LoopERSt6vectorINS_9SizedFileENS_16fuzzer_allocatorIS2_EEE
  14: _ZN6fuzzer12FuzzerDriverEPiPPPcPFiPKhmE
  15: main
  16: __libc_start_main
  17: _start
```

I've double checked my logic on the fuzz and it seems correct. I will attempt to reproduce this behavior on a full node next, and then triple check my fuzz targets to be sure.

sea212:
That's awesome, zachanon, you just have found a instability in the pallet.
In */home/zach/projects/zeitgeist/zrml/orderbook-v1/src/lib.rs:333* we unwrap a value that can become None in case an arithmetic overflow occurred. Can you please add a new Error `ArithmeticOverflow`, change the return type of the `cost` function (lib.rs:333) to `DispatchResult` and change the code, such that it returns `ArithmeticOverflow` when the checked arithmetic operations return `None`?

zachanon:
Sure thing.

zachanon:
I opted to take a different route, as the ```impl Order``` did not like being passed the Config. In this case I changed the return to an ```Option<_>``` type and checked for ```None``` values in the calling functions.

sea212:
Thanks for implementing this feature. Once the pipeline successfully executes we can start the review.
You can use `cargo fmt` to format the code and `cargo fmt --all -- --check` to check if it is formatted, execute *scripts/test/clippy.sh* to check if additional code improvements exist (such as the removal of unnecessary return statements) and finally execute the tests for the orderbook-v1 pallet
`cargo test --features std,runtime-benchmarks --manifest-path zrml/orderbook-v1/Cargo.toml --no-default-features` or all other pallets `./scripts/tests/misc.sh` in case the orderbook-v1 pallet is used by another pallet or in case other pallets were changed, which is not the case.

zrml/orderbook-v1/fuzz/orderbook_v1_full_workflow.rs
```diff
@@ -0,0 +1,99 @@
...
+    match module {
+        0 => Asset::CategoricalOutcome(seed0, seed1),
+        1 => {
+            let scalar_position =
+                if seed1 % 2 == 0 { ScalarPosition::Long } else { ScalarPosition::Short };
+            Asset::ScalarOutcome(seed0, scalar_position)
+        }
+        2 => Asset::CombinatorialOutcome,
+        3 => Asset::PoolShare(SerdeWrapper(seed0)),
+        _ => Asset::Ztg,
```
sea212, line 89: The "_" branch will never be reached.

zrml/orderbook-v1/src/lib.rs
```diff
@@ -252,6 +262,7 @@ mod pallet {
 
     #[pallet::error]
     pub enum Error<T> {
+        ArithmaticOverflow,
```
sea212, line 265: Just a small typo: ArithmaticOverflow -> ArithmeticOverflow

sea212 (commented):
Looks fine overall. Would be nice to have the unreachable branch situation improved.

zrml/orderbook-v1/src/lib.rs
```diff
@@ -73,7 +72,11 @@ mod pallet {
 
                 match order_data.side {
                     OrderSide::Bid => {
-                        T::Currency::unreserve(&maker, order_data.cost());
+                        let cost = match order_data.cost() {
+                            Some(cost) => cost,
+                            None => return Err(Error::<T>::ArithmaticOverflow.into()),
+                        };
+                        T::Currency::unreserve(&maker, cost);
```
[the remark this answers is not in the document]
sea212: My bad about the inclusion of `Error::<T>::ArithmaticOverflow`, overlooked it in the `DispatchError` enum, sorry!

zachanon:
I believe I addressed all concerns.

Is there a better way to error handle when unwrapping Options? The match syntax is okay but I feel like there is probably a more parsimonious syntax to use.

zrml/orderbook-v1/src/lib.rs
```diff
@@ -334,7 +335,13 @@ pub struct Order<AccountId, Balance, MarketId> {
...
 impl<AccountId, Balance: CheckedSub + CheckedMul, MarketId> Order<AccountId, Balance, MarketId> {
-    pub fn cost(&self) -> Balance {
-        self.total.checked_sub(&self.filled).unwrap().checked_mul(&self.price).unwrap()
+    pub fn cost(&self) -> Result<Balance, DispatchError> {
+        match self.total.checked_sub(&self.filled) {
+            Some(subtotal) => match subtotal.checked_mul(&self.price) {
+                Some(cost) => Ok(cost),
+                _ => Err(DispatchError::Arithmetic(ArithmeticError::Overflow)),
+            },
+            _ => Err(DispatchError::Arithmetic(ArithmeticError::Overflow)),
```
sea212, line 344: (pedantic) Would be nice to use the correct `ArithmeticError::Underflow` here
```suggestion
            _ => Err(DispatchError::Arithmetic(ArithmeticError::Underflow)),
```

sea212 (approved):
Thanks for putting the effort into realizing this feature!
We are almost set to merge this into main, you'd only have to incorporate the latest changes of the main branch by rebasing main onto this branch.

sea212:
> Is there a better way to error handle when unwrapping Options? The match syntax is okay but I feel like there is probably a more parsimonious syntax to use.

The only way I know that I personally find a little more readable and not that "cramped" is using `if let` instead. The following example:
```rust
match self.total.checked_sub(&self.filled) {
    Some(subtotal) => match subtotal.checked_mul(&self.price) {
        Some(cost) => Ok(cost),
        _ => Err(DispatchError::Arithmetic(ArithmeticError::Overflow)),
    },
    _ => Err(DispatchError::Arithmetic(ArithmeticError::Underflow)),
}
```

could also be expressed as

```rust
if let Some(subtotal) = self.total.checked_sub(&self.filled) {
    if let Some(cost) = subtotal.checked_mul(&self.price) {
        return Ok(cost);
    }
    return Err(DispatchError::Arithmetic(ArithmeticError::Overflow));
}
Err(DispatchError::Arithmetic(ArithmeticError::Underflow))
```

@c410-f3r has significantly more Rust experience than me, maybe he knows an even better way to express this.

c410-f3r:
Me? I am but a peasant.

Jokes aside, I usually see ppl do `option.ok_or(SomeError)?;` but it is also possible to compose with `and_then` or use let chains when I finish implementing it. Nevertheless, I don't really have a strong reasoning on which option is better.

zachanon:
Now, everything should be in order to merge

PR #255, opened 2021-08-09
Update orml repository

c410-f3r:
Changes back to `https://github.com/zeitgeistpm/open-runtime-module-library`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4815,7 +4815,7 @@ dependencies = [
 [[package]]
 name = "orml-currencies"
 version = "0.4.1-dev"
-source = "git+https://github.com/zeitgeistpm/open-runtime-module-library?branch=0.9.8#2ac4f350da005781a3192efaa4ea7c094782abd4"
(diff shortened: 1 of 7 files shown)


lsaether:
I assume we're using `master` branch because they haven't yet tagged a release? Is this only intended to be temporary and we'll change back to pinning the dependency once we have the deployment issues resolved?

c410-f3r:
The last `crates.io` ORML release was around ~6 months ago when Substrate 2.x/3.x was a thing and the ORML ppl aren't even bothering creating separated branches. I guess it is because of the constant and tiresome Polkadot `0.9.x` releases.

Once they publish a new version, then I will certainty modify the code to use pinned dependencies.

c410-f3r:
> I assume we're using `master` branch because they haven't yet tagged a release? Is this only intended to be temporary and we'll change back to pinning the dependency once we have the deployment issues resolved?

Oh, you mean pinned commits. Well, this can surely can done after the deployment stuff

Issue #251, opened 2021-08-01
[`LiquidityMining`] Figure out how to deal with non time-based market ending


c410-f3r:
cc https://github.com/zeitgeistpm/zeitgeist/pull/122#discussion_r646220423

The Liquidity mining pallet only works with block-based times but `PM` also works with time-based market endings. Therefore, it is necessary to either remove `MarketEnd::Time` or extend `MarketEnd` to the liquidity mining pallet (not sure if it is possible).

PR #259, opened 2021-08-11
Improvements

### LINKED ISSUE #220: Change Scalar range type from `(u128, u128)` to `RangeInclusive`
https://github.com/zeitgeistpm/zeitgeist/blob/8e3387c734eb984726e07a369bbb2d54de68ae6f/primitives/src/market.rs#L146-L147

Needs `apps` support and a new migration

c410-f3r:
Fixes #220 and #251

Taking into consideration that the majority of the block construction have a stable and fixed duration, it is then possible to replace `MarketEnd::Period` with an ending block. For example, in a network with ~6 seconds block construction, ~30 minutes in the future means 300 additional blocks.
This `MarketEnd::Period` removal was motivated due to the fact that the liquidity mining pallet currently only works with block-based times to calculate rewards and the inclusion of actual time evaluation would take more time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12024,6 +12024,7 @@ dependencies = [
  "frame-system",
  "pallet-balances",
  "pallet-randomness-collective-flip",
+ "pallet-timestamp",
  "parity-scale-codec",
  "rand 0.8.4",
  "sp-io",
@@ -12040,11 +12041,13 @@ dependencies = [
  "frame-support",
  "frame-system",
  "pallet-balances",
+ "pallet-timestamp",
  "parity-scale-codec",
  "serde",
  "sp-io",
  "sp-runtime",
  "zeitgeist-primitives",
+ "zrml-market-commons",
 ]
 
 [[package]]
@@ -12174,12 +12177,14 @@ dependencies = [
  "orml-tokens",
  "orml-traits",
  "pallet-balances",
+ "pallet-timestamp",
  "parity-scale-codec",
  "sp-api",
  "sp-io",
  "sp-runtime",
  "zeitgeist-primitives",
  "zrml-liquidity-mining",
+ "zrml-market-commons",
  "zrml-swaps",
  "zrml-swaps-runtime-api",
 ]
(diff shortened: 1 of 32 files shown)


c410-f3r:
The current block producing on https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fbsr.zeitgeist.pm#/explorer is taking `12s` to `24s` so I guess the liquidity mining pallet should add support for time based dead lines to prevent the removal of `MarketEnd`.

sea212:
I am about to start the review, but I hesitate because I think estimating durations from block count is the wrong approach for the following reasons.
Generally it is disadvised to estimate a duration based on block times, because the inaccuracy leads to a cumulative error, whose absolute error becomes bigger the longer the duration, as stated in the documentation of [pallet_timestamp](https://substrate.dev/rustdocs/latest/pallet_timestamp/index.html):

> NOTE: The Timestamp pallet is the recommended way to query the on-chain time instead of using an approach based on block numbers. The block number based time measurement can cause issues because of cumulative calculation errors and hence should be avoided.

In this specific application, prediction markets, we are often reliant on an accurate duration. To provide a simple example, let's assume we have a prediction market that ends when a football match should start. We start the prediction market two weeks before the football match starts and let the prediction market end when the match should start. We can't assume a perfectly benchmarked system and we do also not know, how much overtime the `on_finalize` hooks need. Let's assume we have an average inaccuracy of 1% (which I think is quite optimistic). 1% inaccuracy during two week means we are 3.36 hours off. This potentially could lead to the scenario, that the prediction market closes after the football match is over and the results are already known.

Why was the decision made to remove time based deadlines? Are there some serious problems in implementing it for the liquidity mining pallet?

c410-f3r:
I am already working on adding time support for the liquidity mining pallet and the current block-based system exists because this accuracy situation wasn't a thing on https://github.com/zeitgeistpm/zeitgeist/pull/122.

> Why was the decision made to remove time based deadlines?

There are no incentives for time-based markets
https://github.com/zeitgeistpm/zeitgeist/blob/77cf5a1d1ba6a8eab08048367fdedc351ed0433c/zrml/prediction-markets/src/lib.rs#L397-L402

Let me finish adding time support before review, it is almost done.

c410-f3r:
Should be good now

c410-f3r:
This PR fixes the technical debt of not rewarding markets based on timestamps.
To deal with the non-optimal way of calculating blocks of a time interval, I will open another issue to unblock merging and fix any possible posterior concerns

PR #260, opened 2021-08-12
Remove unnecessary usage of `sp-std`

c410-f3r:
`sp-std` is just a shortcut for the `alloc` or `core` crates and is not necessary at all.


sea212 (approved):

c410-f3r:
This is a trivial change so I will merge

PR #267, opened 2021-08-16
Add Events for Liquidity Mining

### LINKED ISSUE #197: [LiquidityMining] Add events for `distribute_market_incentives`
If `distribute_market_incentives` fails, then it should emit an event `LiquidityRewardsDistributedFailed` while a successful distribution would emit `LiquidityRewardsDistributedSuccess`

c410-f3r:
Fixes #197

### DIFF
--- a/primitives/src/types.rs
+++ b/primitives/src/types.rs
@@ -71,6 +71,11 @@ pub type DigestItem = generic::DigestItem<Hash>;
 /// The market identifier type.
 pub type MarketId = u128;
 
+/// The biggest pointer size taking into consideration all known targeting machine architectures.
+///
+/// Useful to cast `usize` into `MaxUsize` with the guarantee that no truncation will occur.
+pub type MaxUsize = u64;
+
 /// Time
 pub type Moment = u64;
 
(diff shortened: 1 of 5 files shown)


sea212 (commented):
Looks good. I am just not quite sure if we can externally know, if incentives were successfully distributed or if it failed. Since `distribute_market_incentives` is currently only called in `on_initialize` and since it does not emit an event when it's successful, from my current understanding we have no way to query from outside how many incentives were distributed per market, is this correct? We can see how much incentives were added or removed per block in total though.
What happens when an error is returned in `on_initialize` (which in this case returns `Weight`)? Is there a way to notice this without checking the node for logs?

c410-f3r:
> Looks good. I am just not quite sure if we can externally know, if incentives were successfully distributed or if it failed. Since `distribute_market_incentives` is currently only called in `on_initialize` and since it does not emit an event when it's successful, from my current understanding we have no way to query from outside how many incentives were distributed per market, is this correct? We can see how much incentives were added or removed per block in total though.
> What happens when an error is returned in `on_initialize` (which in this case returns `Weight`)? Is there a way to notice this without checking the node for logs?

One event was added to inform the amount of distributed rewards when a market is resolved and another more useful for validators was created to inform that block logic initialization went wrong

sea212 (approved):

PR #270, opened 2021-08-18
[PM] Single extrinsic for creating and deploying a pool

### LINKED ISSUE #269: Single extrinsic for creating and deploying a pool
Currently, while deploying a liquidity pool when creating a market on proto web application there is a need to sign and send multiple extrinsics to:
1 - create a market
2 - buy a complete set of assets needed for liquidity pool
3 - deploy a liquidity pool
4 - add additional assets if needed (e.g in a binary market 120 assets for "Yes" and 110 assets for "No")

Some of the extrinsics can be batched into single transaction, but in worst case, one extrinsic for market creation and two for pool deployment are needed - three transactions.

There is already call for market creation, but would be good to have additional call to be able to create market and deploy liquidity pool in one transaction.
For liquidity pool deployment parameters are `weights` for assets and `asset amounts` to put in the pool.

sea212:
closes #269 

It modified the Swaps API a bit, such that prediction market can call the relevant function to deploy additional assets. An alternative would be to create a new pallet whose `Config` trait is tightly coupled to our prediction market and swaps pallet.

The `pool_join_additional_assets` parameter uses `Asset` as the first element in the 3-tuple. `Asset` contains a market id, but this argument will be replaced by the market id of the market that was created previously during the processing of `create_market_and_deploy_assets`. It is not the most transparent way, but it makes use of existing types, which doesn't further complicate the `types.json`. It is also documented. If preferred, I can replace this by a new type instead.

### DIFF
--- a/primitives/src/traits/swaps.rs
+++ b/primitives/src/traits/swaps.rs
@@ -1,6 +1,6 @@
 use crate::types::{Asset, MarketType, OutcomeReport, Pool, PoolId};
 use alloc::vec::Vec;
-use frame_support::dispatch::{DispatchError, DispatchResult};
+use frame_support::dispatch::{DispatchError, DispatchResult, Weight};
 
 pub trait Swaps<AccountId> {
     type Balance;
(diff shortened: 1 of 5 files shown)


lsaether (approved):
LGTM

PR #272, opened 2021-08-20
Prepare release v0.1.3

sea212:
Handles tasks from #236 that should be done before the release-candidate branch is created


c410-f3r:
It is not possible to review drafts but this PR LGTM.

As stated before, it might be worth to review all runtime parameters before main-net.

PR #276, opened 2021-08-25
Integrate Court into the Runtime

### LINKED ISSUE #200: Court Pallet

c410-f3r:
Depends on #248 
Fixes #200 #281


c410-f3r:
Should be good now

* Added benchamarks
* In regards to https://github.com/zeitgeistpm/zeitgeist/pull/248#discussion_r695740251, punished jurors are simply removed from all markets
* Created #291 and #292 to track the remaining work

lsaether (approved):
Looks good. I think we may want to make some future enhancements to how Binary markets (2 outcome Categorical) handles slashing the losing side, and reducing slashing to 1/2 or some fixed percentage of staked, but I think for now this is good.

Also, slashed jurors should be removed from the Juror pool, requiring them to re-stake if they want to be a juror again. I didn't see this logic in `on_resolution`, can you point it out to me if it exists? Otherwise, let's create an issue for that and go ahead merge this in.

c410-f3r:
> Looks good. I think we may want to make some future enhancements to how Binary markets (2 outcome Categorical) handles slashing the losing side, and reducing slashing to 1/2 or some fixed percentage of staked, but I think for now this is good.
> 
> Also, slashed jurors should be removed from the Juror pool, requiring them to re-stake if they want to be a juror again. I didn't see this logic in `on_resolution`, can you point it out to me if it exists? Otherwise, let's create an issue for that and go ahead merge this in.

https://github.com/zeitgeistpm/zeitgeist/pull/276/files#diff-98309a49dd69a9828c5d52abdaceaa49ea571cfd6e3c98f340484172605d9943R252-R267

This closure is called everytime an already tardy juror does not vote or votes after the maximum allowed block number

c410-f3r:
I will then merge it now, thanks

Issue #205, opened 2021-06-28
Use Substrate's `Collective` for the Advisory Committee


lsaether:
The Advisory Committee is elected by the on-chain council and is tasked as the main curatorial body of Zeitgeist (keeping the markets high quality). On the front-end we will display all markets that have been approved by the Advisory Committee as a good market, while the markets that were created using the `Permissionless` flag will be displayed behind a warning. 

After adding the `Collective` pallet to the runtime, the Advisory Committee should be an allowed origin to the `approve_market` and `reject_market` extrinsics in the `PredictionMarkets` pallet.

sea212:
This could be a potential path of solving this issue:

- [ ] Include *pallet_collective* into the runtime
- [ ] Add the collective instance for the Advisory Committee to the `PredictionMarkets` pallet
- [ ] Add the respective origin as an allowed origin to `approve_market` and `reject_market` extrinsics within the `PredictionMarkets` pallet

at this point, the root can call `set_members` to set the members of the Advisory Committee.
Once the court is realized (maybe create a new issue for that):
- [ ] Implement the `ChangeMembers` trait to allow the court to modify the members of the Advisory Committee.

zachanon:
If appropriate I would like to be assigned to this issue.

sea212:
> If appropriate I would like to be assigned to this issue.

Awesome, thank you for tackling this issue zachanon!
Please let us know if something about the task is not 100% clear.

PR #280, opened 2021-08-28
Implement the ```Collective``` Advisory Committee

zachanon:
Satisfies PR [#205](https://github.com/zeitgeistpm/zeitgeist/issues/205)


zachanon:
In order to implement the pallet into the runtime, we need to decide on a few variables

```
Proposal // Contents of a proposal
MotionDuration // Motions undecided past this are delegated to a 'prime' voter or closed
MaxProposals
MaxMembers
DefaultVote
```

I can continue to implement the pallet with some 'default' features for now, but I would definitely like input as to what these should end up as.

sea212:
You can use this for now:
```rust
parameter_types! {
	pub const AdvisoryCommitteeMotionDuration: BlockNumber = 7 * BLOCKS_PER_DAY;
	pub const AdvisoryCommitteeMaxProposals: u32 = 100;
	pub const AdvisoryCommitteeMaxMembers: u32 = 100;
}

pub type AdvisoryCommittee = pallet_collective::Instance1;
impl pallet_collective::Config<AdvisoryCommittee> for Runtime {
	type Origin = Origin;
	type Proposal = Call;
	type Event = Event;
	type MotionDuration = AdvisoryCommitteeMotionDuration;
	type MaxProposals = AdvisoryCommitteeMaxProposals;
	type MaxMembers = AdvisoryCommitteeMaxMembers;
	type DefaultVote = pallet_collective::PrimeDefaultVote;
	type WeightInfo = weights::pallet_collective::WeightInfo<Runtime>;
}
```

Regarding the new origin, please ensure that either more than half of the members must have voted or the origin can also be root, following [this example from Polkadot runtime](https://github.com/paritytech/polkadot/blob/56a8ebb5ae0ca5980d6861be9d064acc66a374b3/runtime/polkadot/src/lib.rs#L156-L160).

zachanon:
Origin has been added. All current tests are passing. Next step is to add tests to make sure the ```MoreThanHalfCouncil``` origin works as intended.

zachanon:
> Origin has been added. All current tests are passing. Next step is to add tests to make sure the `MoreThanHalfCouncil` origin works as intended.

Well this is embarrassing. It looks like all tests are *not* passing, it's having an issue building the WASM runtime ...

sea212:
The error occurs because when the prediction market pallet is build in `no-std` environment with the `runtime-benchmarks` feature enabled, it tries to use `std` features of the `pallet-collective` (since they were not disabled -> `default-features = false`).
Some other notes: Currently you include the `pallet-collective` as a `dev-dependency` when the `mock` feature is enabled within the prediction market pallet's `Cargo.toml`. This is not what we want, because we want to use the `pallet-collective` also in a non-test environment. In addition, you included the `pallet-collective`'s runtime-benchmarks into the prediction market's `runtime-benchmark` feature. It does not belong there, because the `runtime-benchmark` feature should only contain dependencies that are required to benchmark the prediction market pallet. Putting the `pallet-collective`'s runtime-benchmarks into the `runtime`'s `runtime-benchmark` feature is correct though, because this way the runtime-benchmarks of the `pallet-collective` can be included into the runtime and therefore can be used to benchmark the `pallet-collective`.

**TODOs**
Within `zrml-prediction-market` Cargo.toml:
1) Move `pallet-collective` into default dependencies (non-optional) with `default-features = false`.
2) Remove `pallet-collective` from the `mock` feature.
3) Remove `pallet-collective` from the `runtime-benchmark` feature.
4) Include `pallet-collective/std` in `std` feature

zachanon:
So that was indeed the problem. Thank you for pointing it out. As of the last push ```all-sequential.sh``` runs fine on my machine.

From what I can tell, ```prediction-markets``` does not need ```pallet-collective``` to build. By adding the ```MoreThanHalfCouncil``` type to the Config, the pallet should be able to run as it does already without changes to the logic. I still would like to add the ```pallet-collective``` to the mock, so that I can implement a test that ensures the ```AdvisoryCommittee``` is running properly.

zachanon:
Having trouble resolving an error when implementing a test.

The [changes](https://github.com/zachanon/zeitgeist/commit/4fbdc158af0c414513f52adc0f861c2142cea72b). The [specific line](https://github.com/zachanon/zeitgeist/blob/4fbdc158af0c414513f52adc0f861c2142cea72b/zrml/prediction-markets/src/tests.rs#L144).

On running ```cargo test --features std,runtime-benchmarks --manifest-path zrml/prediction-markets/Cargo.toml --no-default-features``` I get an error

```
error[E0624]: associated function `set_members` is private
   --> zrml/prediction-markets/src/tests.rs:144:30
    |
144 |           assert_ok!(Advisory::set_members(
    |                                ^^^^^^^^^^^ private associated function
```

That I cannot seem to track down the origin of. That method is part of the ```decl_module!{...}``` for pallet-collective, so I should be able to call it in the same way that, e.g., ```MarketCommons::market(...)``` is called.

sea212:
The `set_members` function is private (no `pub` visibility in the `Call` enum's `set_members` function), hence it is not callable outside of the pallet module or it's descendants. You'll have to create an instance of the `Call` enum for the function in question and call one of the functions of the implementation of the [`UnfilteredDispatchable`](https://releases.parity.io/substrate-rustdoc/frame_support/dispatch/trait.UnfilteredDispatchable.html) trait for the `Call` enum. For `set_members`, the code could look like this:
```rust
assert_ok!(
    pallet_collective::Call::<Runtime, pallet_collective::Instance1>::set_members(
        vec![ALICE, BOB, CHARLIE],
        Some(ALICE),
        0u32,
    )
    .dispatch_bypass_filter(Origin::root())
);
```

If you'd like to see how the macros are expanded (for example to see what happens with the `Call` enum), you can use the [*cargo-expand*](https://github.com/dtolnay/cargo-expand) crate. In case of collective, we can see in the expansion that `UnfilteredDispatchable` is implemented for `Call`:
```rust
impl<T: Config<I>, I: Instance> ::frame_support::traits::UnfilteredDispatchable for Call<T, I> {
    type Origin = <T as frame_system::Config>::Origin;
    fn dispatch_bypass_filter(
        self,
        _origin: Self::Origin,
    ) -> ::frame_support::dispatch::DispatchResultWithPostInfo {
        match self {
            Call::set_members(new_members, prime, old_count) => {
                <Module<T, I>>::set_members(_origin, new_members, prime, old_count)
                    .map(Into::into)
                    .map_err(Into::into)
            }

           // -- snip --
        }
    }
}
```

c410-f3r:
What's up @zachanon 

We need to add the collective pallet for the next market campaign so I squashed your commits to create #330. Thanks for your work and interest in helping Zeitgeist!

PR #287, opened 2021-09-02
Remove unnecessary create_pool call

### LINKED ISSUE #178: [`Swaps`] Should `create_pool` be removed?
Not sure if makes sense to have a pool without a market

https://github.com/zeitgeistpm/zeitgeist/blob/d2badc863a54c73a2b90551ba1467d8fe8117228/zrml/swaps/src/lib.rs#L71-L82

@lsaether ?

c410-f3r:
Fixes #178

### DIFF
--- a/zrml/swaps/fuzz/swaps_full_workflow.rs
+++ b/zrml/swaps/fuzz/swaps_full_workflow.rs
@@ -7,13 +7,6 @@ use zrml_swaps::mock::{ExtBuilder, Origin, Swaps};
 fuzz_target!(|data: Data| {
     let mut ext = ExtBuilder::default().build();
     let _ = ext.execute_with(|| {
-        let _ = Swaps::create_pool(
-            Origin::signed(data.create_pool_origin.into()),
-            data.create_pool_assets.into_iter().map(asset).collect(),
-            data.create_market_id.into(),
-            data.create_pool_weights,
-        );
-
         let _ = Swaps::pool_join(
             Origin::signed(data.pool_join_origin.into()),
             data.pool_join_pool_id.into(),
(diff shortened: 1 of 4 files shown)


zrml/swaps/fuzz/swaps_full_workflow.rs
```diff
@@ -7,13 +7,6 @@ use zrml_swaps::mock::{ExtBuilder, Origin, Swaps};
...
-        let _ = Swaps::create_pool(
-            Origin::signed(data.create_pool_origin.into()),
-            data.create_pool_assets.into_iter().map(asset).collect(),
-            data.create_market_id.into(),
-            data.create_pool_weights,
-        );
-
         let _ = Swaps::pool_join(
             Origin::signed(data.pool_join_origin.into()),
             data.pool_join_pool_id.into(),
```
sea212, line 12: As far as I can tell all of the fuzz tests test almost nothing, because they immediately return an error that states that the pool does not exist as one of the first checks: `let pool = Pallet::<T>::pool_by_id(pool_id)?;`.
This error was not introduced in this PR, I just recognized it after the removal of the `create_pool` fuzz tests. Will create an extra issue that targets this.
Anyways, not sure if the removal of the `create_pool` fuzz test is the right way to go. Two reasons:
1) Another pallet might not have fuzz tests and therefore potentially experience an error because we did not (completely) fuzz test our implementation of the public `Swaps` interface.
2) Another pallet that does fuzz test the functions that include `create_pool` cannot test `create_pool` to the same extend, because more parameters and functionality are tested.

sea212 (commented):
Looks good. Need to think about including `create_pool` fuzz test again. The `WeightInfoZeitgeist` trait still contains the `create_pool` weight function, but this is fine since it will be automatically removed (or rather not included in the first place) during the next benchmark.

zrml/swaps/fuzz/swaps_full_workflow.rs
c410-f3r: Oopss... My bad. Thanks for spotting this error and for creating the issue

c410-f3r:
At least for now, it doesn't make sense to have a pool without a market. Also, there isn't a reason to allow non-root accounts create standalone pools.

That said, I will merge this PR but feel free to discuss further future modifications

PR #288, opened 2021-09-03
Audit the use of unchecked arithmetic operations

c410-f3r:
Basically removes explicit arithmetic operations (`+`, `/`, `*`, etc...)  where appropriated to avoid overflows and unexpected outcomes.


c410-f3r:
Reviews are taking a considerable amount of time, re-bases are being a PITA and since this PR mostly deals with ordinary perorations, I will merge now but feel free to question any part of the code

PR #294, opened 2021-09-08
[Rikiddo] Integrate into swaps pallet

### LINKED ISSUE #162: [Rikiddo] Integrate completed Rikiddo pallet
Integrate completed Rikiddo pallet into the system, such that either Rikiddo or CPMM can be used.

sea212:
closes #162


sea212:
- [x] Create `ScoringRule` type.
- [x] Include `ScoringRule` into `Pool`, add `ScoringRule` parameter to `create_pool` and adjust logic.
- [x] Adjust all functions to be able to handle the new `Pool` for CPMM scoring rule.
- [ ] Figure out: Include base currency in pool or not?
- [ ] Integrate Rikiddo into the runtime.
- [ ] Add Rikiddo to Swaps Config.
- [ ] Implement Rikiddo variant for create_pool: Create default instance and change ema long and short periods.
- [ ] Figure out a way to switch between CPMM and Rikiddo for every function and implement solution:
	- [ ] `admin_set_pool_as_stale`
	- [ ] `create_pool`
	- [ ] `get_spot_price`
	- [ ] `pool_exit`
	- [ ] `pool_exit_with_exact_asset_amount`
	- [ ] `pool_exit_with_exact_pool_amount`
	- [ ] `pool_join`
	- [ ] `pool_join_with_exact_asset_amount`
	- [ ] `pool_join_with_exact_pool_amount`
	- [ ] `set_pool_as_stale`
	- [ ] `swap_exact_amount_in`
	- [ ] `swap_exact_amount_out`
- [ ] Update tests.
- [ ] Update benchmarks.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1046,20 +1122,30 @@ mod pallet {
...
+            let mut total_weight: Weight =
+                Self::process_subsidy_collecting_markets(now, T::MarketCommons::now());
+
+            let mut do_resolution = || {
+                Self::resolution_manager(now, |market_id, market| {
+                    let weight = Self::on_resolution(market_id, market)?;
+                    total_weight = total_weight.saturating_add(weight);
+                    Ok(())
+                })
+            };
```
c410-f3r, line 1134: Not sure why this method needs to be wrapped inside a closure. You can move the whole `Self::resolution_manager(...)` inside `with_transaction`.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1447,13 +1571,177 @@ mod pallet {
...
+                    return true;
+                };
+
+            let mut weight_basis = 0;
+            <MarketsCollectingSubsidy<T>>::mutate(
+                |e: &mut Vec<SubsidyUntil<T::BlockNumber, MomentOf<T>, MarketIdOf<T>>>| {
+                    weight_basis = T::WeightInfo::process_subsidy_collecting_markets_raw(
+                        e.len().saturated_into(),
+                    );
+                    e.retain(retain_closure);
```
c410-f3r, line 1740: **Nit**

`retain` is a bit tricky. Since errors are omitted, it will be almost impossible to assert the integrity of an operation

zrml/swaps/src/lib.rs
```diff
@@ -770,6 +1341,182 @@ mod pallet {
...
+                    Ok(res) => {
+                        if res.result {
+                            TransactionOutcome::Commit(Ok(res))
+                        } else {
+                            TransactionOutcome::Rollback(Ok(res))
+                        }
+                    }
+                    Err(err) => TransactionOutcome::Rollback(Err(err)),
+                }
+            })
```
c410-f3r, line 1504: on_initialize -> process_subsidy_collecting_markets -> end_subsidy_phase

`on_initialize` already calls `with_transaction`. IMHO, `with_transaction` should only be used on pallet calls or pallet hooks.

c410-f3r (approved):
You can ignore all nits as they are not so important, just a few thoughts from my mind.

As commented before, the changes are big so proper understanding and checks/reviews of possible invariants should come afterwards

zrml/prediction-markets/src/lib.rs
sea212: Will be improved in a future iteration. I didn't want to return with an error, since all other (Rikiddo) markets would not be processed in that case. Instead, the error-producing market will be kept and on every block error message will indicate that something went wrong, which we can fix then during the next update.

zrml/swaps/src/lib.rs
sea212: The problem here is that the code has to introduce state changes to see if there are some inconsistencies, i.e. some other functions has unreserved more than it should and therefore funds reserved for subsidy are not sufficient anymore in total. In this case the function does not fail though - it just reverts all changes and returns that the operation was successful, but no state transition occured because not as much could be unreserved as was reserved / registered before.
I can't think of another solution rn and with the release pressing, I'd like to move this to the next iteration, since it does not affect functionality (it's not a bug, maybe just not the most elegant solution). Let's create an issue for that.

zrml/prediction-markets/src/lib.rs
c410-f3r: Sure

zrml/swaps/src/lib.rs
c410-f3r: An issue looks good.

zrml/swaps/src/lib.rs
```diff
@@ -586,39 +911,233 @@ mod pallet {
...
+            pool: &Pool<BalanceOf<T>, T::MarketId>,
+            pool_id: PoolId,
+            base_asset: Asset<T::MarketId>,
+            winning_asset: Asset<T::MarketId>,
+            winner_payout_account: &T::AccountId,
+        ) -> Weight {
+            // CPMM handling of market profit not supported
+            if pool.scoring_rule == ScoringRule::CPMM {
+                return 1_000_000;
+            }
```
lsaether, line 942: this code should never be reachable, therefore I think we can safely remove it

lsaether (approved):

PR #296, opened 2021-09-10
Improvements

### LINKED ISSUE #253: Remove remaining uses of explicity arithmetic operations and explcity primitive type casts
* Explicit arithmetic operations like `a / b` or ` a - b` can overflow or underflow depending on the input, leading to plain wrong results or runtime panics.
* Explicit primitive casts can lead to unpredictable behaviours so instead of `let a: u32 = 99u8 as _`, use `let a: u32 = 99.u8.into()` or `let a: u32 = 99u8.saturated_into()`.

### LINKED ISSUE #293: [Court] jurors can vote, then leave the court to avoid penalties
If I read https://github.com/zeitgeistpm/zeitgeist/blob/main/zrml/court/src/lib.rs#L79 right, jurors can leave the court and unreserve their stake at any time.

This likely means that a juror could have their vote counted in a dispute and be eligible for incentives - while avoiding any disincentives.

@c410-f3r

c410-f3r:
Fixes #253
Fixes #293

Unifies #289 with a change that removes remaining references of departed jurors


zrml/rikiddo/src/types.rs
```diff
@@ -213,22 +213,25 @@ impl<F: Fixed, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
...
                     fixed_str.retain(|c| c != '.');
                     // Padding to the right side up to `places`. Cannot underflow.
-                    fixed_str += &"0".repeat((places as usize).saturating_sub(frac_string.len()));
+                    fixed_str += &"0".repeat(places_usize.saturating_sub(frac_str.len()));
                 }
                 Ordering::Greater => {
                     // Cutting down to `places` + arithmetic rounding of the last digit
                     let frac_plus_one_digit_str =
-                        &frac_string[0..(places as usize).saturating_add(1)];
+                        frac_str.get(..places_usize.saturating_add(1)).unwrap_or_default();
```
c410-f3r, line 234: @sea212 Are these changes OK?

lsaether (approved):

sea212 (approved):

zrml/rikiddo/src/types.rs
sea212: Yes, looks good.

c410-f3r: If `fixed_frac` is 1 or 2 then `frac_str.len().cmp(&places_usize)` will always fall under the `Ordering::Less` variant but I wasn't sure if that was the intended behavior.

Well, thanks for looking into it

PR #297, opened 2021-09-10
Add Authorized pallet

c410-f3r:
Depends on #296
Fixes #292

The pallet is not yet integrated into the runtime

PR #298, opened 2021-09-10
Revamp PM

c410-f3r:
Depends on #297

Integrates the authorized pallet, uses reserved balances and moves back common logic that was wrongly placed on the simple-disputes pallet


zrml/prediction-markets/src/lib.rs
```diff
@@ -335,7 +345,7 @@ mod pallet {
             ensure!(creator == sender, "Canceller must be market creator.");
             ensure!(status == MarketStatus::Proposed, "Market must be pending approval.");
             // The market is being cancelled, return the deposit.
-            CurrencyOf::<T>::unreserve(&creator, T::AdvisoryBond::get());
+            CurrencyOf::<T>::unreserve_named(&RESERVE_ID, &creator, T::AdvisoryBond::get());
```
sea212, line 348: We need distinguished names/ids for the reservation, otherwise different reservations get mixed and slashed altogether Example: `AdvisoryBond` and `OracleBond` go in the same pot with id `RESERVE_ID`. If one user participates in both roles and fails one, both his reservations get slashed.
Suggestion: Per reservation type, use one unique id that ideally includes the pallet id if possible.

sea212 (commented):
Thanks Caio, except the currency reservation issue everything looks good to me.

zrml/prediction-markets/src/lib.rs
c410-f3r: I think this unified behavior is intended. For example, `create_(categorical|scalar)_market` sums up `AdvisoryBond` and `OracleBond`. @lsaether ?

c410-f3r: @sea212 https://github.com/zeitgeistpm/zeitgeist/issues/306 was opened to track this matter. In the meanwhile, I will merge this PR

PR #300, opened 2021-09-14
Add Crowdloan pallet

c410-f3r:
IMHO, this is the best pallet available but feel free to suggest another crowdloan implementation

### DIFF
--- a/node/src/chain_spec/dev.rs
+++ b/node/src/chain_spec/dev.rs
@@ -6,7 +6,7 @@ use crate::chain_spec::{
 };
 use sc_service::ChainType;
 use sp_core::sr25519;
-use zeitgeist_primitives::types::Balance;
+use zeitgeist_primitives::{constants::BASE, types::Balance};
 
 const INITIAL_BALANCE: Balance = Balance::MAX >> 4;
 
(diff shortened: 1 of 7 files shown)


node/src/chain_spec/dev.rs
```diff
@@ -27,6 +27,7 @@ pub fn dev_config(
                         get_from_seed::<nimbus_primitives::NimbusId>("Alice"),
                         crate::chain_spec::DEFAULT_STAKING_AMOUNT,
                     )],
+                    crowdloan_fund_pot: 100u128.saturating_add(BASE),
```
sea212, line 30: I think this is ought to be `saturating_mul`. Currently `crowdloan_fund_pot = 10000000100`.
In case this is true, *node/src/chain_spec/mod.rs* needs to be changed as well.

sea212 (approved):
Thanks, looks fine.

node/src/chain_spec/dev.rs
c410-f3r: Ooppss... My bad.

Thanks @sea212 !

lsaether: I opened #326 so we can take care of this before the PLO.

lsaether (approved):

c410-f3r:
Thanks guys

PR #261, opened 2021-08-12
[Rikiddo] Implement fuzz

sea212:
closes #163

PR #301, opened 2021-09-14
Clippy

c410-f3r:
Makes master green again

cc @sea212


c410-f3r (commented):
Took from #261. The point of the `integer_arithmetic` lint is audit the usage of arithmetic operators so any explicit operator should be fine with a documentation that holds true invariants and tell external people the intended behavior

Issue #249, opened 2021-08-01
Group remaining common constants/types scattered across the code-base


c410-f3r:
`zeitgeist-primitives` already groups many common structures but there are many scattered left-overs that could be de-duplicated across crates and runtime definitions.

PR #303, opened 2021-09-16
Remove some redundant parameters

c410-f3r:
Part of #249

### DIFF
--- a/zrml/rikiddo/fuzz/rikiddo_pallet.rs
+++ b/zrml/rikiddo/fuzz/rikiddo_pallet.rs
@@ -11,6 +11,7 @@ use frame_support::traits::{OnFinalize, OnInitialize};
 use frame_system::RawOrigin;
 use libfuzzer_sys::fuzz_target;
 use substrate_fixed::{types::extra::U33, FixedI128, FixedU128};
+use zeitgeist_primitives::constants::MinimumPeriod;
 use zrml_rikiddo::{
     mock::*,
     traits::RikiddoSigmoidMVPallet,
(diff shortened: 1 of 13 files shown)


zrml/rikiddo/fuzz/rikiddo_pallet.rs
```diff
@@ -56,8 +57,11 @@ fuzz_target!(|data: Data| {
             if idx % 2 == 1 {
                 current_block += 1;
                 run_to_block(current_block);
-                let _ = <Runtime as Config>::Timestamp::set(RawOrigin::None.into(), current_block)
-                    .unwrap();
+                let _ = <Runtime as Config>::Timestamp::set(
+                    RawOrigin::None.into(),
+                    MinimumPeriod::get() * current_block,
+                )
+                .unwrap();
```
c410-f3r, line 64: @sea212 I don't think this modification changes any testing logic but can you confirm?

c410-f3r:
Merging as changes are trivial

zrml/rikiddo/fuzz/rikiddo_pallet.rs
sea212: It does change the testing logic. Since in the first case where `Timestamp::set` is called the number of seconds is greater than 1, both ema instance (long and short) skip both in the last state and afterwards represent the same value. This leads to the situation that the fee parameter will always be 1.
Why did you add this in the first place? The fuzz test does not have to be logically consistent with our parameters, it must be able to test as many code execution paths with test data sets. Can you please revert the changes for this file and remove zeitgeist-primitives again?

c410-f3r: > Why did you add this in the first place?

`MinimumPeriod` is not 0 on production environments, CI fails without this change (https://github.com/zeitgeistpm/zeitgeist/runs/3635598436?check_suite_focus=true) and parameters are being de-duplicated  as previously stated.

> Can you please revert the changes for this file and remove zeitgeist-primitives again?

Yes, I will revert

Issue #291, opened 2021-09-07
[Court] `JurorsStatus::Tardy` juror will never be changed back to `JurorStatus::Ok`


c410-f3r:
Once a juror is marked as `JurorStatus::Tardy`, he will always be tardy and I am not sure if this behaviour is desired.

ckamm:
It looks to me like this causes issues. In a subsequent dispute a tardy juror will neither receive an incentive when voting for the best option, nor will their stake be slashed when voting for something other than the best two options. See `slash_losers_to_award_winners` use of `non_tardy_jurors_with_votes`.

c410-f3r:
Yes, already tardy jurors that voted within the `CourtCaseDuration` duration are valid and should be eligible to receive rewards or be slashed, therefore, `non_tardy_jurors_with_votes`  could be renamed to something like `valid_jurors_with_votes` and also yield valid tardy jurors.

Thanks again for spotting another problem

PR #305, opened 2021-09-17
Valid tardy jurors should also be slashed to reward winners

c410-f3r:
Fix https://github.com/zeitgeistpm/zeitgeist/issues/291#issuecomment-915181046


zrml/court/src/lib.rs
```diff
@@ -310,54 +343,16 @@ mod pallet {
...
 
-            for (jai, _, _, (_, outcome)) in
-                Self::non_tardy_jurors_with_votes(requested_jurors, votes)
-            {
-                if outcome == winner_outcome {
+            for (jai, outcome) in valid_winners_and_losers {
+                if outcome == &winner_outcome {
                     total_winners = total_winners.saturating_add(BalanceOf::<T>::from(1u8));
                 } else {
                     let all_reserved = CurrencyOf::<T>::reserved_balance_named(&RESERVE_ID, jai);
```
sea212, line 358: In regards to line 360: It is not part of the current changes, but it would be nice to have constants like slash percentages as pallet parameters (50% is currently hard-coded).

sea212 (approved):
Thanks, LGTM!

zrml/court/src/lib.rs
c410-f3r: Created https://github.com/zeitgeistpm/zeitgeist/issues/309 to tackle this matter. Personally, runtime parameters are useful for different relay chains (Kusama and/or Polkadot) but third-party usage is questionable due to the specific and particular nature of this pallet.

Issue #250, opened 2021-08-01
What to do with valid permissionless markets?


c410-f3r:
https://github.com/zeitgeistpm/zeitgeist/blob/97f233d51ed6f9bef9450e965a52810fc1ed9f50/zrml/simple-disputes/src/lib.rs#L220-L230

cc @lsaether

c410-f3r:
@lsaether Can you give more context?

c410-f3r:
Sup guys

It seems that all my issues in this repository are stalled for a period long time and since I am cleaning up my `https://github.com/issues` list, I will close everything due to inactiveness. Hope you guys don't mind, maybe even some issues are no longer relevant? ¯\_(ツ)_/¯

In any case, here is a list for possible future references or for the creation of new material/issues.

* https://github.com/zeitgeistpm/zeitgeist/issues/130
* https://github.com/zeitgeistpm/zeitgeist/issues/141
* https://github.com/zeitgeistpm/zeitgeist/issues/145
* https://github.com/zeitgeistpm/zeitgeist/issues/165
* https://github.com/zeitgeistpm/zeitgeist/issues/196
* https://github.com/zeitgeistpm/zeitgeist/issues/250
* https://github.com/zeitgeistpm/zeitgeist/issues/256
* https://github.com/zeitgeistpm/zeitgeist/issues/258
* https://github.com/zeitgeistpm/zeitgeist/issues/283
* https://github.com/zeitgeistpm/zeitgeist/issues/306
* https://github.com/zeitgeistpm/zeitgeist/issues/309
* https://github.com/zeitgeistpm/zeitgeist/issues/313

Cheers

PR #311, opened 2021-09-22
Remove commented code

c410-f3r:
This snippet has always been commented and does not contain valid logic. For example, `report.outcome` is not a number or `unreserve` already uses `T::ValidityBond::get()` regardless.

If permission-less markets need special handling, then future development on this area should be discussed on https://github.com/zeitgeistpm/zeitgeist/issues/250 or another issue. If not, https://github.com/zeitgeistpm/zeitgeist/issues/250 can be closed.


sea212 (approved):

c410-f3r:
@lsaether I will merge this PR since the commented code is invalid but feel free discuss permission-less markest on #250 or another issue

PR #312, opened 2021-09-22
Group most runtime parameters

c410-f3r:
Fix https://github.com/zeitgeistpm/zeitgeist/issues/249


c410-f3r:
Merging as changes are trivial but feel free to discuss possible following divergences

Issue #313, opened 2021-09-22
[Runtime] Fill `benchmark_metadata` with actual relevant pallet information


c410-f3r:
On https://github.com/zeitgeistpm/zeitgeist/pull/310, Substrate introduced a new benchmark method called `benchmark_metadata` (https://github.com/zeitgeistpm/zeitgeist/pull/310/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR510-R522) and this method needs to be filled with relevant pallet information for Zeitgeist.

For example, take a look at what Acala did -> https://github.com/AcalaNetwork/Acala/blob/8dbdde4517f4af36d51a0b32d1ead41d4d3eba77/runtime/karura/src/lib.rs#L2094-L2136

c410-f3r:
https://github.com/zeitgeistpm/zeitgeist/pull/329 added some data but it would be nice the audit all necessary pallets. For example, `orml-tokens` isn't currently taken into consideration.

c410-f3r:
Sup guys

It seems that all my issues in this repository are stalled for a period long time and since I am cleaning up my `https://github.com/issues` list, I will close everything due to inactiveness. Hope you guys don't mind, maybe even some issues are no longer relevant? ¯\_(ツ)_/¯

In any case, here is a list for possible future references or for the creation of new material/issues.

* https://github.com/zeitgeistpm/zeitgeist/issues/130
* https://github.com/zeitgeistpm/zeitgeist/issues/141
* https://github.com/zeitgeistpm/zeitgeist/issues/145
* https://github.com/zeitgeistpm/zeitgeist/issues/165
* https://github.com/zeitgeistpm/zeitgeist/issues/196
* https://github.com/zeitgeistpm/zeitgeist/issues/250
* https://github.com/zeitgeistpm/zeitgeist/issues/256
* https://github.com/zeitgeistpm/zeitgeist/issues/258
* https://github.com/zeitgeistpm/zeitgeist/issues/283
* https://github.com/zeitgeistpm/zeitgeist/issues/306
* https://github.com/zeitgeistpm/zeitgeist/issues/309
* https://github.com/zeitgeistpm/zeitgeist/issues/313

Cheers

PR #329, opened 2021-09-30
Fill `benchmark_metadata`

c410-f3r:
Turns out that the newly introduced benchmark method must be filled in order to actually run benchmarks

cc @sea212 
cc https://github.com/zeitgeistpm/zeitgeist/issues/313


sea212:
Much appreciated @c410-f3r, I owe you one!

PR #334, opened 2021-10-03
[Pallets] Prepare events for usage in Subsquid

### LINKED ISSUE #275: [Pallets] Evaluate completeness of events
To be able to effectively use Subsquid, we have to output the information that should be processed (and is not already part of the call itself) within events.
It is possible to query the state of blocks in the past by sending a RPC to an archival node. However, this is not the desired approach, since it introduces additional problems, such as timeout handling, knowledge of low-level node information outside the node (how did the storage structure look at every point in time), etc.

**This must be done before the upcoming launch of the next testnet**

**Update meeting 1st Sep. 2021**
Most of the information can be gathered via events. Whenever a new structure is stored, such as a `Market` or a `Pool`, an event is emitted that contains the id, the structure itself and the account id. All following events which introduce changes to those data structures must contain the id of the specific instance of the data structure and, if available, the account id that instructed the changes.
By using this approach, Subsquid is able to collect the complete data set whenever a new instance of a data structure is created (such as a `Market`) and trace all subsequent changes to it (since the changes can be joined using the corresponding ids). In addition, any additional information such as the timestamp, block number and account id, can be used to execute complex queries over many different datasets to gain additional insights.
A corner-case are the spot prices: They are calculated on demand for a pair, therefore we cannot trace every spot price through events. In this case, the SDK will use a [RPC](https://github.com/zeitgeistpm/tools/blob/10b5810cb257400c8c9c557a16af42837e9d528d/packages/sdk/src/models/swaps.ts#L74-L88) to calculate the prices. This allows us to gather data about the price history in addition to the quantity (and therefore implicitly the volume) history. This data in turn can be used in our research to evaluate our market scoring rule and in response the behavior of the users. Querying Subsquid prices in contrast to using a RPC to fetch the price history is by magnitudes faster.

**TODOs**

- [x] #284
- [ ] zrml-swaps: Include `Pool` that was created to the `PoolCreate` event and add pool ids to the events wherever possible.
- [x] zrml-court: Emit events that contain the juror that joined or left the court.
- [ ] zrml-orderbook-v1: Include `OrderData` that was created to the `OrderMade` event.

sea212:
Closes #275 
Depends on #294

### DIFF
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -871,37 +871,40 @@ mod pallet {
     {
         /// Share holder rewards were distributed. \[pool_id, num_accounts_rewarded, amount\]
         DistributeShareHolderRewards(PoolId, u64, BalanceOf<T>),
-        /// A new pool has been created. \[account\]
-        PoolCreate(CommonPoolEventParams<<T as frame_system::Config>::AccountId>),
-        /// Someone has exited a pool. \[account, amount\]
+        /// A new pool has been created. \[CommonPoolEventParams, pool\]
+        PoolCreate(
+            CommonPoolEventParams<<T as frame_system::Config>::AccountId>,
+            Pool<BalanceOf<T>, T::MarketId>,
+        ),
+        /// Someone has exited a pool. \[PoolAssetsEvent\]
         PoolExit(PoolAssetsEvent<<T as frame_system::Config>::AccountId, BalanceOf<T>>),
         /// Someone has (partially) exited a pool by removing subsidy. \[account, amount\]
         PoolExitSubsidy(PoolAssetEvent<<T as frame_system::Config>::AccountId, BalanceOf<T>>),
-        /// Exits a pool given an exact amount of an asset. \[account, amount\]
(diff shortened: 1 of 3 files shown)


lsaether (approved):

zrml/swaps/src/lib.rs
```diff
@@ -1256,7 +1256,6 @@ mod pallet {
         /// * `scoring_rule`: The scoring rule that's used to determine the asset prices.
         /// * `swap_fee`: The fee applied to each swap (mandatory if scoring rule is CPMM).
         /// * `weights`: These are the raw/denormalized weights (mandatory if scoring rule is CPMM).
-        #[frame_support::transactional]
```
lsaether, line 1256: Why remove this? If we call this from an extrinsic in the prediction markets pallet, then we should make it so it reverts the entire logic

sea212: Good catch, tbh. I have no idea how I managed to remove it, it was not on purpose.

c410-f3r (approved):

PR #339, opened 2021-10-05
Remove standalone node

c410-f3r:
Now that 0.2.0 was issue, feel free to merge this PR once you guys feel confident enough with the Battery Station Relay Chain


lsaether:
How does this work when using the `--dev` flag? Will it still allow for the solo developer node for quick testing?

c410-f3r:
IIRC, Moonbeam created a built-in way to run a development instance without setting up a relay chain and it should be possible to do the same for Zeitgeist.
But as things are now, it is necessary to call `./scripts/parachain/local.sh` in order to do local tests.

c410-f3r:
> > IIRC, Moonbeam created a built-in way to run a development instance without setting up a relay chain and it should be possible to do the same for Zeitgeist.
> 
> Ok, I think until we have a strategy for this we should keep around the solo node. The front end team sometimes tests using a local development node. We currently facilitate this using a single line docker command `docker run zeitgeistpm/zeitgeist-node --dev`. Even with the local script, asking for someone to build both the polkadot binary and zeitgeist binary instead is much less ideal imo.

Seems like a real problem, indeed

PR #340, opened 2021-10-05
Re-name OutcomeReport to Outcome

c410-f3r:
`OutcomeReport` was originally named `Outcome` and its renaming was a over-sighted mistake. Changing back to `Outcome`.

cc @saboonikhil

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -6,7 +6,7 @@ mod asset;
 pub mod constants;
 mod market;
 mod max_runtime_usize;
-mod outcome_report;
+mod outcome;
 mod pool;
 mod pool_status;
 mod serde_wrapper;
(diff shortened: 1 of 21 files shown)


lsaether:
We renamed our `Outcome` type to avoid a metadata conflict with [Polkadot's `Outcome` type](https://github.com/paritytech/polkadot/blob/cc24fb87b702db6510d08c689e442b6be384d798/xcm/src/v1/traits.rs#L108-L117) which was added for XCM. 

I remember this causing an issue when using Polkadot-JS at the time, however it could be something we can solve instead with a type alias.

c410-f3r:
Really? I don't remember. Closing this issue then

cc https://github.com/zeitgeistpm/zeitgeist/pull/140/files

lsaether:
Hmm, I remember having hit this problem at one point, although the renaming could have just been a lucky coincidence then.

PR #373, opened 2021-11-22
Add transaction filter

### LINKED ISSUE #363: Add a transaction filter
When we first deploy the parachain, we will want to have a transaction filter so that some of the extrinsics are not allowed to be called. Sudo, and other governance origins, should be able to bypass this filter.

sea212:
closes #363

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12267,6 +12267,7 @@ dependencies = [
  "frame-system-benchmarking",
  "frame-system-rpc-runtime-api",
  "hex-literal 0.3.3",
+ "log",
  "nimbus-primitives",
  "orml-currencies",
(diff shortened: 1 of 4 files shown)


lsaether (approved):
Looks good overall, will build and test it later.

Note: We need to be careful when we build the production runtime to use the transaction filtering feature.

sea212:
> Note: We need to be careful when we build the production runtime to use the transaction filtering feature.

Yup, agreed.
I tested the txfilter on the parachain-build today and it works as expected. The parachain produces blocks. Every tx that calls non-essential extrinsics is blocked. Sudo works and can be used to either execute as root or as another user (by doing so, every extrinsic can be successfully executed).

PR #383, opened 2021-11-26
Register remaining benchmarks

sea212:
Preparation for #377 

Before we are able to benchmark every pallet (including external pallets), we have to register their benchmarking logic in our runtime.
The frame pallets work out-of-the-box, the orml pallet unfortunately require a custom benchmarking environment and benchmarking logic.

PR #389, opened 2021-11-29
Export pallet constants

### LINKED ISSUE #387: [Pallets] Export configuration into storage
Currently some of our pallets only use constant in the code without exporting them into the storage, which renders reading configuration values in the frontend impossible. One example is court.

**Solution:**
Add `#[pallet::constant]` above the declaration of a type with the trait `Get` within the pallet `Config` trait.

sea212:
depends on #383 
closes #387


lsaether (approved):
Thanks!

PR #391, opened 2021-11-29
Add assets in question to swap events

### LINKED ISSUE #388: [Swaps] Events do not contain assets in questions
When liquidity is added/removed or a swap is executed, the event that gets emitted does not contain the assets in question.

sea212:
closes #388

Tested every event manually.


sea212:
cc @saboonikhil 

Most of the events changed: All join/exit liquidity/subsidy events and the swap events now contain the assets in question.
This will be active on Zeitgeist mainnet. For now, staging keeps the same. I'll let you know when we the new events are used on staging.

lsaether (approved):

PR #401, opened 2021-12-08
Bump pallet-author-mapping version

### LINKED ISSUE #404: [Pallets] pallet-author-mapping contains deprecated migration code
Executing a runtime upgrade with the current version of `pallet-author-mapping` leads to the execution of a deprecated migration that deletes all authors from storage.

sea212:
closes #404 

The previous version had an unchecked migration, which lead to a removal of all authoring accounts during a runtime upgrade. Consequently, the runtime upgrade with the previous version will stall the parachain.
This PR is required to create a valid runtime for Battery Station.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4739,7 +4739,7 @@ dependencies = [
 [[package]]
 name = "nimbus-primitives"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


sea212:
Fuzz workflow fails currently because `cargo-fuzz` can't be installed. The installation fails since the dependency of `cargo-fuzz` `proc-macro2` was updated from v1.0.32 to v1.0.33. Probably needs to use stable branch on rustc to install `cargo-fuzz`.

sea212:
Depends on #405

PR #415, opened 2021-12-22
Add state changes to PM events

### LINKED ISSUE #396: Add or modify event params
Modify param `Currency` to `CurrencyId` (to resolve type conflicts) for the following `currency` events: 
1. Transferred
2. Deposited
3. Withdrawn

_(Note: Deferring the above point to continue with the manual workaround since this would require making changes in the currency pallet)_

Add modified market fields along with `MarketIdOf` (to ensure that the logic remains at one place) for the following `predictionMarkets` events:
1. MarketApproved
2. MarketStartedWithSubsidy
3. MarketInsufficientSubsidy
4. MarketReported
5. MarketDisputed
6. MarketRejected
7. MarketCancelled

sea212:
closes #396 

This PR adds the new state to PM events that are emitted after the market state changed.
Found one event that as never emitted: `MarketResolved`


sea212:
Any event that is emitted after a change state includes the new state now.

@saboonikhil Are those changes sufficient?

saboonikhil:
> Any event that is emitted after a change state includes the new state now.
> 
> @saboonikhil Are those changes sufficient?

Looks good - if only state is changing for all PM events.

saboonikhil (approved):

PR #423, opened 2021-12-26
Fix/implement bpow_approx for `base >= 2`

maltekliemann:
It seems like `bpow_approx` is broken.

For `base` larger or equal to `2 * BASE`, the `while` loop in `bpow_approx` never ends. This is because you're using the binomial series to calculate the power `base^exp`, but the binomial series' convergence radius is `1` (for any exponent), so as you shift by `1` (or, rather, `BASE`) the method used only works if `0 < base < 2 * BASE`. It seems to me that this cannot be guaranteed in functions like `calc_in_given_out`, where `bpow` is used.

The idea of the fix is that if you want to calculate `base^exp` with `base > 1` using the binomial series, then you set `x = - (base - 1)/base`, calculate `y = (1 + x)^exp` using the binomial series and then return `1/y` (this works as `1/(1 + x)` is equal to `base`). I'm not sure if there's a prettier way to do this, but using the two `if` statements seemed straightforward to me.

I took the values for the test `bpow_works_with_some_non_integer_values` from Python's floating point arithmetic, but had to change some of the less significant digits. First five digits are the same as Python's for all eight tests. No idea if this is a good measure of precision.

### DIFF
--- a/zrml/swaps/src/fixed.rs
+++ b/zrml/swaps/src/fixed.rs
@@ -61,7 +61,9 @@ pub fn bpow(base: u128, exp: u128) -> Result<u128, DispatchError> {
 
 pub fn bpow_approx(base: u128, exp: u128, precision: u128) -> Result<u128, DispatchError> {
     let a = exp;
-    let (x, xneg) = bsub_sign(base, BASE)?;
+    // Reduce the general case `base^exp` to a computation of the binomial series `(1 + x)^exp`.
+    let (x, xneg) =
+        if base < 2 * BASE { bsub_sign(base, BASE)? } else { (bdiv(base - BASE, base)?, true) };
     let mut term = BASE;
     let mut sum = term;
     let mut negative = false;
@@ -95,7 +97,12 @@ pub fn bpow_approx(base: u128, exp: u128, precision: u128) -> Result<u128, Dispa
         i = i.check_add_rslt(&1)?;
     }
 
-    Ok(sum)
+    // Deduce the correct result from the result of the binomial series.
+    if base < 2 * BASE {
+        Ok(sum)
+    } else {
+        Ok(bdiv(BASE, sum)?) // Reciprocal value
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


maltekliemann:
There's another problem. If you use `bpow_approx` with small `base`, say `BASE / 100`, and default `BPOW_PRECISION`, then you eventually get an `Err(Other("Arithmetic overflow"))`. This seems like a numerical problem: `x` is very close to `BASE`, and so `c` and `big_k` run extremely close, meaning that these two operations leave `term` unchanged:

```rust
term = bmul(term, bmul(c, x)?)?;
term = bdiv(term, big_k)?;
```

So the `sum` keeps piling up until overflow.

Switching from computing `bpow(base, exp)`, where `base` is close to zero, to computing first `y = bpow(bdiv(BASE, base), exp)` and then returning `bdiv(BASE, y)` might be a solution (so we get a third branch for the two `if` statements). But I'd like to hear how you feel about this. Maybe there's a better way?

sea212:
Thanks for the report maltekliemann, very interesting observation. I'll have a look next week when I am back in office.

maltekliemann:
I think I may have figured this out... So the problem is that `bpow` doesn't work for a base larger or equal to `2` (no convergence for the binomial series), and is numerically unstable for values close to zero.

Balancer asserts a minimum and maximum base of 1 Wei and 2 BAL for `bpow`: https://github.com/balancer-labs/balancer-core/blob/f4ed5d65362a8d6cec21662fb6eae233b0babc1f/contracts/BNum.sol#L112-L113 I don't think these assertions are numerically safe; they do not occur in the Zeitgeist code base.

But that doesn't really matter, I think. The `bpow` function is called from the `calc_*` functions in `math.rs`. Each of these makes an assumption about the ratio of two of its parameters.

For example, let take `swap_exact_amount_out`. This is the only method that calls `calc_in_given_out`. It expects that `asset_amount_out` is at most one third (`MaxOutRatio`) of `balance_out`. Thus (see the source code of `calc_in_given_out`). Thus `diff >= (2/3) * balance_out` and `diff <= asset_balance_out`. This gives `1 <= y <= 3/2`. So `y` is well within the set of "safe" parameters for `bpow`. In fact, I guess this is why the `MaxInRatio`/`MaxOutRatio` checks exist?

You can do a similar computation for `swap_exact_amount_in`, and probably for all the pool entry/exit function as well (I haven't checked these). So you don't even need anything from this pull request.

So, if like, I could work out some sound numeric limits for `bpow` and raise an error if they are violated. Or you can just leave everything the way it is.

maltekliemann:
Closed in favor of #474 and #475.

PR #424, opened 2022-01-04
Ensure that first dispute has different outcome

maltekliemann:
This is already ensured for the second and later disputes, and should probably be checked when making the first dispute, as well.

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -2041,7 +2041,7 @@ mod pallet {
             Self::ensure_can_not_dispute_the_same_outcome(
                 disputes,
                 (&market.report.as_ref()).ok_or(Error::<T>::MarketNotReported)?,
-                outcome
+                outcome,
             )?;
             Self::ensure_disputes_does_not_exceed_max_disputes(num_disputes)?;
             Ok(())
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1477,10 +1477,13 @@ mod pallet {
 
         fn ensure_can_not_dispute_the_same_outcome(
             disputes: &[MarketDispute<T::AccountId, T::BlockNumber>],
+            report: &Report<T::AccountId, T::BlockNumber>,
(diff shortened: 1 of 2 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -2035,7 +2038,11 @@ mod pallet {
         ) -> DispatchResult {
             ensure!(market.report.is_some(), Error::<T>::MarketNotReported);
             Self::ensure_outcome_matches_market_type(market, outcome)?;
-            Self::ensure_can_not_dispute_the_same_outcome(disputes, outcome)?;
+            Self::ensure_can_not_dispute_the_same_outcome(
+                disputes,
+                &market.report.as_ref().unwrap(),
```
sea212, line 2043: We generally avoid code that can panic.
```suggestion
                (&market.report.as_ref()).ok_or(Error::<T>::MarketIsNotReported)?
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -2035,7 +2038,11 @@ mod pallet {
...
+                outcome
+            )?;
             Self::ensure_disputes_does_not_exceed_max_disputes(num_disputes)?;
             Ok(())
```
sea212, line 2047: Please increment `spec_version` in `runtime/src/lib.rs` by one to signal changes in the runtime logic.

sea212 (changes requested):
Thanks for pointing this out and fixing it.

zrml/prediction-markets/src/lib.rs
maltekliemann: Understood. Doesn't `ensure!(market.report.is_some(), Error::<T>::MarketNotReported);` guarantee that `unwrap()` doesn't fail, though?

sea212: Yes, it does. It's just a philosophy to avoid code that can potentially panic, even if it is clear that it can not occur in the current state of the code.

maltekliemann: Ok, got it!

sea212 (approved):

PR #432, opened 2022-01-17
Ensure that there's no activity on pending proposed markets

maltekliemann:
Without this change, you can deploy a pool for a proposed market and buy/sell complete sets. There major problem with this is that if the market then gets rejected, there's an orphaned pool on the chain.

So the idea is that it's now no longer allowed to do either of these things. I've added a test which checks that pending advised markets don't allow buying/selling complete sets. The test that deploying pools on pending advised markets isn't allowed is slightly awkward, since we're not allowed to even obtain the required liqiudity. So this test fails for multiple reasons and may throw an unexpected error in the future, depending on implementation details of `deploy_swap_pool_for_market`. I'm not sure if that test is a good idea. Feel free to remove it.

I'm not sure if this is going in the right direction. If so, then before merging, I would like to remove the `creation` parameters of `create_cpmm_market_and_deploy_assets`, since with `Advised`, the function will now always raise an error.


sea212:
Thanks for the PR. I think those changes make sense. We don't want that orphaned pool. In addition, I don't think that anybody wants to provide liquidity that is not being used and does not return profits.

@TvrtkoM How difficult is it to integrate this distinction (disable combined market creation + pool deployment for `Advised` markets, but allow for `Permissionless` markets) into the frontend?

sea212 (approved):
LGTM. Feel free to remove the `creation` parameter for the combined market creation and pool deployment function.

TvrtkoM:
@sea212 It shouldn't be too much effort to update this in the application

maltekliemann:
@sea212 @TvrtkoM I'll remove the `creation` parameter from market creation with pool in the near future and add that to this PR.

TvrtkoM:
ok @maltekliemann, that'll require small sdk change as well as some updates on web-app. but not too many changes will be necessary imo.

sea212 (approved):

PR #435, opened 2022-01-21
Update docker spin-up script

sea212:
*Oopsie, wrong branch name*

Removes the relay chain spec file since it is included in the binary now, optimizes the script and updates the relay chain docker image.


sea212:
Not required.

PR #468, opened 2022-02-23
Add corner-case tests for fixed type conversion

maltekliemann:
We also change the way that tests are arranged. This is a bit of a half-measure (macros or using a testing framework would be a more idomatic long-term solution), but it serves one particular purpose in this case: Previously, if the vector with the correct results was too short, the test would be ignored. This is now impossible.


sea212 (approved):
Thanks, LGTM

PR #469, opened 2022-02-24
Use BaseCallFilter to filter extrinsics

sea212:


maltekliemann (approved):
Can't comment on the selection of allowed pallets in detail, but it seems that everything except basics, XCM and consensus is disabled, which checks out.

Looks goods to me!

PR #470, opened 2022-02-24
Remove panickers from Rikiddo code

### LINKED ISSUE #338: [Rikiddo] Replace unchecked variants
Replace all unchecked variants of `to_num` and `to_fixed` with their checked counter-parts.

maltekliemann:
Removed all instances of `to_num`, `from_num`, `to_fixed` and `from_fixed`. Some repeated code is now abstracted into utility functions.

To remove panickers from implementations of the `Default` trait, we've added a `From<u32>` or `From<i32>` trait bounds to some Rikiddo structs. These do not seem to be serious restrictions.

Closes #338.

### DIFF
--- a/zrml/rikiddo/src/types.rs
+++ b/zrml/rikiddo/src/types.rs
@@ -127,18 +127,27 @@ impl Timespan {
 /// Returns integer part of `FROM`, if the msb is not set and and num does it into `FROM`.
 fn convert_common<FROM: Fixed, TO: Fixed>(num: FROM) -> Result<TO, &'static str> {
     // Check if number is negatie
-    if num < FROM::from_num(0u8) {
+    let zero = FROM::checked_from_num(0u8)
+        .ok_or("Unexpectedly failed to convert 0u8 to fixed point number")?;
+    if num < zero {
         return Err("Cannot convert negative signed number into unsigned number");
     }
 
     // PartialOrd is bugged, therefore the workaround
     // https://github.com/encointer/substrate-fixed/issues/9
-    let num_u128: u128 = num.int().to_num();
-    if TO::max_value().int().to_num::<u128>() < num_u128 {
+    let num_u128: u128 = num.int().checked_to_num().ok_or("Conversion from FROM to u128 failed")?;
+    let max_value: u128 = TO::max_value()
+        .int()
+        .checked_to_num()
+        .ok_or("Converting fixed point numerical limit to u128 failed uexpectedly")?;
+    if max_value < num_u128 {
         return Err("Fixed point conversion failed: FROM type does not fit in TO type");
     }
 
-    Ok(num_u128.to_fixed())
+    let fixed = num_u128
+        .checked_to_fixed()
+        .ok_or("Integer part of fixed point number does not fit into u128")?;
+    Ok(fixed)
 }
 
 /// Converts an unsigned fixed point number into a signed fixed point number (fallible).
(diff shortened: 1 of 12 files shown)


maltekliemann:
The exact choice of type bound is certainly up for debate. I've simply tried to stick to the original material as close as possible. Maybe `u8` is a better, less constricting choice?

sea212:
@maltekliemann I think we should use the lowest representation possible to allow usage with as many types as possible. If it is possible to use `From<u8>` and `From<i8>` instead `From<u32>` and `From<i32>`, we should that trait bound imo.

maltekliemann:
@sea212 Done!

zrml/rikiddo/src/types.rs
```diff
@@ -206,17 +219,26 @@ impl<F: Fixed, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
...
+                        } else {
+                            return Err("Unexpected overflow when dividing one by two");
                         }
+                    } else {
+                        return Err("Unexpectedly failed to convert 1 to fixed point number");
                     }
+                } else {
+                    return Err("Unexpectedly failed to convert 2 to fixed point number");
                 }
             }
```
sea212, line 243: [nitpicking] We could formulate that a little more concise. With so many else branches, it requires some effort to assign them to the operation that caused it

```rust
if F::frac_nbits() > 0 {
    let one = F::checked_from_num(1)
        .ok_or("Unexpectedly failed to convert 1 to fixed point number")?;
    let two = F::checked_from_num(2)
        .ok_or("Unexpectedly failed to convert 2 to fixed point number")?;
    let half =
        one.checked_div(two).ok_or("Unexpected overflow when dividing one by two")?;

    if fixed.frac() >= half {
        result = result.saturating_add(1);
    }
}
```

sea212: I remember that you have asked a question related to that based on that code example. I hope my answer was not misleading. Both ways are fine imo, it's just when the code becomes lengthy and all branches are used, it can become a bit hard to read.

zrml/rikiddo/src/utils.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/// Return the maximum value of FixedType as u128.
+pub fn max_value_u128<FixedType: Fixed>() -> Result<u128, &'static str> {
+    if let Some(res) = FixedType::max_value().int().checked_to_num() {
+        Ok(res)
+    } else {
+        Err("Unexpectedly failed to convert max_value of fixed point type to u128")
+    }
+}
```
sea212, line 22: Since you created a new module for utility functions, I think we should also move the following function into that module: https://github.com/zeitgeistpm/zeitgeist/blob/cd55ba8a6af316993b288b319436e4b55cddd585/zrml/rikiddo/src/types.rs#L127-L172

sea212: Not a 100% sure about the following, but I tend towards yes: https://github.com/zeitgeistpm/zeitgeist/blob/cd55ba8a6af316993b288b319436e4b55cddd585/zrml/rikiddo/src/types.rs#L174-L299

sea212 (approved):
Thanks for removing that evil code. Besides some nitpicking, everything is fine from my perspective.

zrml/rikiddo/src/types.rs
maltekliemann: No, the answer wasn't misleading at all. :) I used this approach to avoid changing the style, but I agree that the alternative is more readable.

zrml/rikiddo/src/utils.rs
maltekliemann: Agreed. Thinking about this in terms of encapsulation, I would consider these functions to be implementation details of `types.rs`, so putting them in the utilities seems correct.

sea212 (approved):

PR #474, opened 2022-02-28
[Swaps] Clarify and implement numerical limits for `bpow` and `bpow_approx`

maltekliemann:
Major changes to `bpow` and `bpow_approx`, which are used to calculate spot prices and pool join/exit swaps:

- Ensure that `bpow_approx` always terminates
- Define API for `bpow` and `bpow_approx`
- Add checks for numerical limits
- Test `bpow` and `bpow_approx`

This "solves" part of the problems addressed in #423 in the sense that, instead of extending the range of `bpow`, we instead enforce its restriction.


zrml/swaps/src/fixed.rs
```diff
@@ -180,14 +232,158 @@ mod tests {
...
+            assert_le!(diff, precision);
+        }
+    }
+
+    #[test]
+    fn bpow_returns_error_when_parameters_are_outside_of_specified_limits() {
+        let test_vector: Vec<(u128, u128)> =
+            vec![(BASE / 10, 3 * BASE / 2), (2 * BASE - BASE / 10, 3 * BASE / 2)];
+        for (base, exp) in test_vector.iter() {
+            assert!(bpow(*base, *exp).is_err());
```
sea212, line 285: I think this approach of testing the result is sufficient, but a more pedantic test could incorporate the exact errors that are excepted. I think in that case it is not necessary, but generally some other change might introduce another error in one of the test cases that is thrown before the excepted errors.
The check could look like this:

```rust
let errors = vec![
    Err(DispatchError::Other("[bpow_approx]: expected base >= BASE / 4")),
    Err(DispatchError::Other("[bpow_approx]: expected base <= 7 * BASE / 4"))
];
let test_vector: Vec<(u128, u128)> =
    vec![(BASE / 10, 3 * BASE / 2), (2 * BASE - BASE / 10, 3 * BASE / 2)];
for (base, exp) in test_vector.iter() {
    assert!(errors.contains(&bpow(*base, *exp)));
}
```

An even cleaner approach would be to open up two different test cases for the respective expected errors.

zrml/swaps/src/fixed.rs
```diff
@@ -180,14 +232,158 @@ mod tests {
...
+    fn bpow_approx_returns_error_when_parameters_are_outside_of_specified_limits() {
+        let test_vector: Vec<(u128, u128)> = vec![
+            (BASE / 10, 3 * BASE / 2),
+            (2 * BASE - BASE / 10, 3 * BASE / 2),
+            (BASE, 11 * BASE / 10),
+        ];
+        for (base, exp) in test_vector.iter() {
+            assert!(bpow_approx(*base, *exp).is_err());
+        }
     }
```
sea212, line 388: Just for the sake of completeness, a test that verifies that the max iteration count check works would be great.

zrml/swaps/src/lib.rs
```diff
@@ -7,6 +7,9 @@
 
 extern crate alloc;
 
+#[cfg(test)]
+#[macro_use]
+extern crate more_asserts;
```
sea212, line 12: [nitpick] [`#[macro_use]`](https://doc.rust-lang.org/1.29.0/book/first-edition/macros.html#scoping-and-macro-importexport) leads to a global macro import, which can be useful when having tests in multiple different modules that use those macros. Otherwise the import is too broad and it should be used only within the relevant module (by using the `use` statement), like the fixed module in our case. 
This is just my opinion. For example, by using this approach executing specific tests (e.g. `cargo test fixed`) will import the macros, whereas other tests won't.

zrml/swaps/src/fixed.rs
```diff
@@ -70,8 +112,18 @@ pub fn bpow_approx(base: u128, exp: u128, precision: u128) -> Result<u128, Dispa
...
+            break;
+        }
+
+        if i == BPOW_APPROX_MAX_ITERATIONS {
+            // Unreachable with the current limits.
+            return Err(DispatchError::Other(
+                "[bpow_approx] Maximum number of iterations exceeded",
+            ));
+        }
+
```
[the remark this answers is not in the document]
maltekliemann: Yeah, much better that way 👍

zrml/swaps/src/fixed.rs
maltekliemann: Would have loved to do that. Unfortunately, I don't think that's possible. I've profiled the function like this:

```rust
for base in (BASE / 4..7 * BASE / 4).step_by((BASE / 1_000) as usize) {
    for exp in (0..BASE).step_by((BASE / 1_000) as usize) {
        bpow_approx(base, exp).unwrap();
    }
}
```

Maximum number of iterations was 50, and I've got an estimate here that says it should never take more than 70 iterations. So the loop is purely a defense against human error. To me, this is similar to the idea of not having panickers in the code, even if you can argue that they will never trigger.

If you think it's worth it, I can do the following: Put the `precision` parameter back in and then test that with sufficiently small `precision` the errors gets triggered.

sea212: Thanks for the explanation. We can leave it like that.

zrml/swaps/src/lib.rs
maltekliemann: Done!

zrml/swaps/src/fixed.rs
```diff
@@ -55,11 +71,37 @@ pub fn bpow(base: u128, exp: u128) -> Result<u128, DispatchError> {
...
+/// # Errors
+///
+/// If this function encounters an arithmetic over/underflow, or if the numerical limits
+/// for `base` or `exp` (specified above) are violated, a `DispatchError::Other` is
+/// returned.
+pub fn bpow_approx(base: u128, exp: u128) -> Result<u128, DispatchError> {
+    // We use the binomial power series for this calculation. We stop adding terms to
+    // the result as soon as one term is smaller than `BPOW_PRECISION`. (Thanks to the
+    // limits on `base` and `exp`, this means that the total error should not exceed
+    // 4*BPOW_PRECISION`.)
```
sea212, line 94:
```suggestion
    // the result as soon as one term is smaller than `BPOW_PRECISION`. Due to the
    // limits on `base` and `exp`, the total error should not exceed `4*BPOW_PRECISION`.
```

zrml/swaps/src/fixed.rs
```diff
@@ -180,14 +232,166 @@ mod tests {
...
+                DispatchError::Other("[bpow_approx]: expected base >= BASE / 4"),
+            ),
+            (
+                2 * BASE - BASE / 10,
+                1 * BASE / 2,
+                DispatchError::Other("[bpow_approx]: expected base <= 7 * BASE / 4"),
+            ),
+        ];
+        for (base, exp, err) in test_vector.iter() {
+            assert_eq!(bpow_approx(*base, *exp).unwrap_err(), *err);
```
sea212, line 394: The `frame_support` crate offers a macro for that specific case: `assert_err!(result, excepted_error)`
```suggestion
            frame_support::assert_err!(bpow_approx(*base, *exp), *err);
```

sea212 (approved):

PR #475, opened 2022-03-02
[Swaps] Add `Max*Ratio` checks

maltekliemann:


maltekliemann:
Related to #474, where we restrict the values that `bpow` and `bpow_approx` will accept.

- I'm setting `MaxInRatio` and `MaxOutRatio` to the same value for clarity. It shouldn't matter for practical cases.
- The new test case outlines a very particular bug and should stand on its own.
- Added two `Max*Ratio` checks. When combined with #474, these slightly restrict the possible values that the checked extrinsics are called with. The purpose of these checks is to provide a proper error message for the case that too much of a pool is removed or added.
- The added checks restrict liquidity providers. In practical terms, this means that a provider who owns a large part of the pool cannot remove their entire liquidity at once, and must instead proceed in increasingly smaller increments. However, a similar restriction is already in place in form of other `Max*Ratio` checks. However, the important checks are those that make sure that the numerical functions aren't called with any values that they can't handle.

What bugs me about the last point: How do the liquidity providers get their tokens back when the market is closed? Then _everybody_ will want to remove their tokens, and those who are last will have to do it in increments and will be forced to leave some of their tokens behind, unless there is another mechanism.

zrml/swaps/src/math.rs
```diff
@@ -101,6 +173,21 @@ pub fn calc_single_in_given_pool_out(
     asset_amount_in
 }
 
+/// Calculate the amount of a single asset received when exiting the pool with a specified amount of
```
sea212, line 176:
```suggestion
/// Calculate the amount of tokens received when exiting the pool with a specified amount of
```

zrml/swaps/src/math.rs
```diff
@@ -126,6 +213,21 @@ pub fn calc_single_out_given_pool_in(
     asset_amount_out
 }
 
+/// Calculate the amount of pool tokens required to obtain the specified amount of a single asset
```
sea212, line 216:
```suggestion
/// Calculate the amount of pool tokens required to obtain the specified number of tokens for a single asset
```

zrml/swaps/src/math.rs
```diff
@@ -126,6 +213,21 @@ pub fn calc_single_out_given_pool_in(
...
+/// when exiting the pool.
+///
+/// See _Single-Asset Deposit/Withdrawal_ for details.
+///
+/// * `asset_balance_in` - The pool balance of the ingoing asset
+/// * `asset_weight_in` - The weight of the ingoing asset
+/// * `pool_supply` - The total supply of pool tokens
+/// * `total_weight` - The total weight of the pool (sum of all weights)
+/// * `asset_amount_out` - The expected amount of the outgoing asset to receive
```
sea212, line 225:
```suggestion
/// * `asset_amount_out` - The expected amount outgoing asset tokens to receive
```

zrml/swaps/src/lib.rs
```diff
@@ -480,6 +484,12 @@ mod pallet {
...
                 asset_amount: |asset_balance: BalanceOf<T>, total_supply: BalanceOf<T>| {
+                    let mul: BalanceOf<T> = bmul(
+                        total_supply.saturated_into(),
+                        T::MaxOutRatio::get().saturated_into(),
+                    )?
+                    .saturated_into();
+                    ensure!(pool_amount <= mul, Error::<T>::MaxOutRatio);
                     let asset_amount: BalanceOf<T> = crate::math::calc_single_in_given_pool_out(
                         asset_balance.saturated_into(),
                         Self::pool_weight_rslt(&pool, &asset)?,
```
sea212, line 495: I think those constraints also have to be implemented for `pool_exit_with_exact_asset_amount` and `pool_join_with_exact_asset_amount`

sea212 (commented):
The changes look fine to me. The only concern I have is that some functions are still not covered yet.

In regards to the problem of being constrained when adding or removing liquidity and having to call it multiple times: This will be handled in a later PR that covers #479

zrml/swaps/src/lib.rs
```diff
@@ -292,6 +292,10 @@ mod pallet {
             let params = PoolExitWithExactAmountParams {
                 asset,
                 asset_amount: |asset_balance: BalanceOf<T>, total_supply: BalanceOf<T>| {
+                    let mul: BalanceOf<T> =
+                        bmul(total_supply.saturated_into(), T::MaxInRatio::get().saturated_into())?
+                            .saturated_into();
+                    ensure!(pool_amount <= mul, Error::<T>::MaxInRatio);
                     let asset_amount: BalanceOf<T> = crate::math::calc_single_out_given_pool_in(
                         asset_balance.saturated_into(),
                         Self::pool_weight_rslt(&pool, &asset)?,
```
sea212, line 301: A little below after the caculation this part can be found:
```rust
ensure!(
    asset_amount
        <= bmul(
            asset_balance.saturated_into(),
            T::MaxOutRatio::get().saturated_into()
        )?
        .saturated_into(),
    Error::<T>::MaxOutRatio
);
```

Is this still necessary? (is also related to `pool_join_with_exact_pool_amount`)

maltekliemann: To summarize, all the `pool_join_with_*` and `pool_exit_with_*` extrinsics check that asset amounts swapped in or out are within the `MaxInRatio` or `MaxOutRatio` relative to the asset balance.

Furthermore, the `*_exact_asset_amount` extrinsics perform this check _before_ calling any of the methods from `math.rs`, which is useful, as they make sure that the functions from `math.rs` are not called with unallowed values. On the other hand, the `*_exact_pool_amount` extrinsics only perform this check _after_ the `math.rs` functions are called, which, in my opinion, draws their use into question.

My suggestion is to leave this in until we have completely audited the Balancer implementation. I'd say this PR is more about putting the important checks in rather than removing unnecessary stuff.

Issue #493, opened 2022-03-18
[Runtime] Create one runtime configuration per network


sea212:
To be able to use different configurations on mainnet and testnet, such as different collator, staking or XCM configurations, we'll have to create one runtime per network which is loaded whenever the respective chain is passed to the nodes `--chain` option.

PR #494, opened 2022-03-18
Add democracy

### LINKED ISSUE #453: [Runtime] Switch mainnet from sudo to community governance
This is an overarching issue, that encompasses multiple issues. Currently the mainnet uses a superuser. The superuser must be replaced by community governance solutions. Some tasks required to tackle for this are:

- Split the runtime into Battery Station and Zeitgeist (mainnet), the former will continue to use a superuser, while the latter will use community governance exclusively
- We need a council for system-overaching decisions, for example runtime proposal approval
- We need a technical committee to maintain the chain, for example swap to maintenance mode, execute runtime upgrades and verify stability
- We need a mechanism to launch proposals

sea212:
closes #453 

This PR does:

- [ ] Add collective and membership instances for the general council and the technical committee
- [ ] Add the democracy pallet
- [ ] Create multiple origins per collective, percentage of votes. The root origin is part of all of those (for now)
- [ ] Assign power to collectives as specified in the [democracy document](https://docs.google.com/spreadsheets/d/1gHThrMm4aD5gOI4iv_yliGpPnxC8IVDJ/edit?usp=sharing&ouid=113507151362969450355&rtpof=true&sd=true)

This PR does not:
   - Separate combined runtime into one runtime per testnet. Postponed within #493


runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type Event = Event;
+    type MaxMembers = TechnicalCommitteeMaxMembers;
+    type MaxProposals = TechnicalCommitteeMaxProposals;
+    type MotionDuration = TechnicalCommitteeMotionDuration;
+    type Origin = Origin;
+    type Proposal = Call;
+    type WeightInfo = weights::pallet_collective::WeightInfo<Runtime>;
+}
+
+// TODO configure origins + Add custom weight
```
maltekliemann, line 592:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type CancelProposalOrigin = EnsureRootOrAllTechnicalCommittee;
+    type VetoOrigin = pallet_collective::EnsureMember<AccountId, TechnicalCommitteeInstance>;
+    type CooloffPeriod = CooloffPeriod;
+    type PreimageByteDeposit = PreimageByteDeposit;
+    type OperationalPreimageOrigin = pallet_collective::EnsureMember<AccountId, CouncilInstance>;
+    type Slash = Treasury;
+    type Scheduler = Scheduler;
+    type PalletsOrigin = OriginCaller;
+    type MaxVotes = MaxVotes;
+    // TODO Add custom weight
```
maltekliemann, line 620:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
     type MaxRegistrars = MaxRegistrars;
     type MaxSubAccounts = MaxSubAccounts;
-    type RegistrarOrigin = EnsureRoot<AccountId>;
+    type RegistrarOrigin = EnsureRootOrHalfCouncil;
     type Slashed = Treasury;
     type SubAccountDeposit = SubAccountDeposit;
     type WeightInfo = weights::pallet_identity::WeightInfo<Runtime>;
 }
 
+// TODO add origins
```
maltekliemann, line 640:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type MembershipChanged = AdvisoryCommittee;
+    type MembershipInitialized = AdvisoryCommittee;
+    type PrimeOrigin = EnsureRootOrTwoThirdsCouncil;
+    type RemoveOrigin = EnsureRootOrTwoThirdsCouncil;
+    type ResetOrigin = EnsureRootOrTwoThirdsCouncil;
+    type SwapOrigin = EnsureRootOrTwoThirdsCouncil;
     type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
 }
 
+// TODO add config values
```
maltekliemann, line 654:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type MembershipChanged = AdvisoryCommittee;
+    type MembershipInitialized = AdvisoryCommittee;
+    type PrimeOrigin = EnsureRootOrThreeFourthsCouncil;
+    type RemoveOrigin = EnsureRootOrThreeFourthsCouncil;
+    type ResetOrigin = EnsureRootOrThreeFourthsCouncil;
+    type SwapOrigin = EnsureRootOrThreeFourthsCouncil;
+    type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
+}
+
+// TODO add config values
```
maltekliemann, line 668:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type MembershipChanged = TechnicalCommittee;
+    type MembershipInitialized = TechnicalCommittee;
+    type PrimeOrigin = EnsureRootOrTwoThirdsCouncil;
+    type RemoveOrigin = EnsureRootOrTwoThirdsCouncil;
+    type ResetOrigin = EnsureRootOrTwoThirdsCouncil;
+    type SwapOrigin = EnsureRootOrTwoThirdsCouncil;
+    type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
+}
+
+// Todo set weightinfo
```
maltekliemann, line 682:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -491,36 +567,146 @@ impl pallet_collective::Config<AdvisoryCommitteeCollectiveInstance> for Runtime
...
+    type Currency = Balances;
+    type ManagerOrigin = EnsureRoot<AccountId>;
+    type MaxSize = PreimageMaxSize;
+    type BaseDeposit = PreimageBaseDeposit;
+    type ByteDeposit = PreimageByteDeposit;
+}
+
 impl pallet_randomness_collective_flip::Config for Runtime {}
 
+// Todo set weightinfo
```
maltekliemann, line 695:
```suggestion
```

maltekliemann (commented):
Looks good to me, all comments are either nitpicks or benign question.

One thing, though... Of the powers mentioned in the democracy document, two seem to be missing (or I'm missing them):

- Resume or suspend XCMP queue
- Maintenance mode

sea212:
Thanks for the review. In regards to missing permissions:
- Resume or suspend XCMP queue -> This config value is only available since `polkadot-v0.9.17`, so the origin will be assigned within the next release but one.
- Maintenance mode -> The pallet that handles this is not yet integrated

runtime/src/weights/pallet_democracy.rs
```diff
@@ -0,0 +1,283 @@
...
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
+use core::marker::PhantomData;
+
+/// Weights for pallet_democracy using the Substrate node and recommended hardware.
+pub struct WeightInfo<T>(PhantomData<T>);
+impl<T: frame_system::Config> pallet_democracy::weights::WeightInfo for WeightInfo<T> {
+	// Storage: Democracy PublicPropCount (r:1 w:1)
```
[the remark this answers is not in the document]
sea212: I'll remember that for when taking care of https://github.com/zeitgeistpm/zeitgeist/issues/500

primitives/src/constants.rs
```diff
@@ -46,46 +49,18 @@ parameter_types! {
...
-    pub ExistentialDeposits: |currency_id: CurrencyId| -> Balance {
-        match currency_id {
-            Asset::Ztg => ExistentialDeposit::get(),
-            _ => 0
-        }
-    };
-}
-parameter_types! {
-    pub const GetNativeCurrencyId: CurrencyId = Asset::Ztg;
-    pub DustAccount: AccountId = PalletId(*b"orml/dst").into_account();
```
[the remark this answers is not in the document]
sea212: They both exist but I have moved the `DustAccount` into the runtime `parameters` module and left `GetNativeCurrencyId` in the primitives create. The reasoning behind this is the following: Ideally we would have the configuration for the pallets in the runtime crate from my perspective, where they are applied. It happens though that some of the values are needed in tests. Circular dependencies are not allowed in rust, hence we can't have those config values in the runtime crate. That's why they are in the primitives crate.
Not sure if splitting the configuration in two crates is ideal though, maybe a cleaner approach would be an all-or-nothing approach where we have all config values for a pallet in exactly one of those.

runtime/src/parameters.rs
```diff
@@ -63,6 +97,15 @@ parameter_types! {
   pub const OperationalFeeMultiplier: u8 = 5;
   pub const TransactionByteFee: Balance = 100 * MICRO;
 
-  // XCM
-  pub const MaxInstructions: u32 = 100;
+  // Treasury
+  pub const Burn: Permill = Permill::from_percent(50);
+  pub const MaxApprovals: u32 = 100;
+  pub const ProposalBond: Permill = Permill::from_percent(5);
+  pub const ProposalBondMinimum: Balance = 10 * BASE;
```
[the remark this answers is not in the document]
sea212: I think this is fine, since when a proposal is rejected the complete proposal bond will be slashed, even if in theory that proposal made sense. [Moonbeam](https://github.com/PureStake/moonbeam/blob/617f917b236e93b807daba6086e2823d8485893f/runtime/moonbeam/src/lib.rs#L554), [Acala](https://github.com/AcalaNetwork/Acala/blob/5a8f381a95649eb74385fb7db31421d78118a651/runtime/acala/src/lib.rs#L568) and [Kusama](https://github.com/paritytech/polkadot/blob/22967a4128702b12e438bb951aae32927b949dcd/runtime/kusama/src/lib.rs#L737) use 1, 10 and 20 native currency respectively. I suggest we start of with ten and increase it in the future in case it is insufficient.

runtime/src/parameters.rs
```diff
@@ -26,9 +27,28 @@ parameter_types! {
...
+  pub const TechnicalCommitteeMaxMembers: u32 = 100;
+  pub const TechnicalCommitteeMaxProposals: u32 = 64;
+  pub const TechnicalCommitteeMotionDuration: BlockNumber = 7 * BLOCKS_PER_DAY;
+
+  // Democracy
+  pub const LaunchPeriod: BlockNumber = 5 * BLOCKS_PER_DAY;
+    pub const VotingPeriod: BlockNumber = 5 * BLOCKS_PER_DAY;
+    pub const FastTrackVotingPeriod: BlockNumber = 3 * BLOCKS_PER_HOUR;
+    pub const MinimumDeposit: Balance = 100 * BASE;
+    pub const EnactmentPeriod: BlockNumber = 2 * BLOCKS_PER_DAY;
```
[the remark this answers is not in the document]
sea212: I am not 100% certain which stake they exactly mean by that, but as I understand it relates too all the parts of the runtime where staking is possible, such that stakers have the chance to remove their stake until the proposal is enacted.
If that is the case, all constraints should be met since the only places in the runtime where staking is possible are:
- `court` pallet, unstaking is immediately possible without a time constraint
- `parachain_staking` pallet, anybody can bond/unbond stake or leave completely within 2 rounds (currently 4 minutes) - this will most likely never exceed 1 day.
- `pallet_democracy` not exactly sure if it can be called staking, but voters and delegators back their votes with funds. In that case it is obviously not desired that they can pull out their funds during `EnactmentPeriod`

On a side note, the current configuration only allows voting once for a proposal and doesn't leave room for changing opinions.

`EnactmentPeriod = 2` was chosen, because:
1. Other great Kusama parachains use that
2. It allows to move forward relatively fast with accepted changes (which imo. is appropriate for a fast-changing and experimental system like ours)

I think eventually we will have two configurations for testnet and mainnet, whereas testnet will be much faster paced than mainnet, but for now I think the given configuration is appropriate for both networks.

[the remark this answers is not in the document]
maltekliemann: My understanding was that in case of a malignous decision, those who oppose the decision and fear for the economic implications have a chance to unstake and cash out before the change in enacted. Seems like we are well within that timeframe. 👍

primitives/src/constants.rs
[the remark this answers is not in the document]
maltekliemann: Ok, I missed that! Thanks for explaining!

Issue #485, opened 2022-03-16
[swaps] Check Balancer security audit


maltekliemann:
balancer.fi published a security audit for Balancer v1: https://github.com/balancer-labs/balancer-core/blob/master/Trail%20of%20Bits%20Full%20Audit.pdf

We should review which of these findings apply to our implementation of Balancer pools.

- [x] 1. Single-asset liquidity functions allow stealing of assets
- [x] 2. Lack of events in setBLabs is error-prone
- [x] 3. Parameters' order of single-asset functions is confusing
- [x] 4. Assets will be lost in case of token migrations
- [x] 5. Users can silently burn tokens with transfers to 0x0
- [x] 6. Privileged addresses can be transferred without confirmation even to invalid values
- [x] 7. Users can join and exit pools even where there are no tokens
- [x] 8. Single-asset exit functions allow withdrawing of a negligible amount of assets for free
- [x] 9. Assets with low decimals or low liquidity lead to withdraw a negligible amount of assets for free
- [x] 10. Rounding issues in joinPool/exitPool allow for a negligible amount of free pool tokens
- [x] 11. Attackers with large funds can steal the pool's assets
- [x] 12. The normalized sum of the weight is not always equal to 1
- [x] 13. Pools with a large total supply cause SWAP functions to always revert
- [x] 14. Token balance limits are declared but not enforced
- [x] 15. The swap-in and swap-out ratios are not correctly enforced

maltekliemann:
2, 4, 5, 6 seem to be EVM-only problems.

maltekliemann:
3 is resolved in Balancer and Zeitgeist

maltekliemann:
1 is also based on Balancer's "controller" address and non-finalized pools. Note that this vulnerability is not related to #496.

maltekliemann:
Regarding **8**, the report later states (p. 44):

> This seems to have been mitigated by a check that prevents zero division errors from reaching zero.

Not sure what "reaching zero" refers to, or what zero division errors this is refering to, but we have a check that ensures that `calc_pool_in_given_single_out` doesn't return zero.

Furthermore, the report states on the same page:

> While it is still possible to take advantage of rounding errors [...], it incurs and extremely high cost to the attacker.

This checks out. Rounding errors can also be "exploited" to withdraw more liquidity from the pool than you own, but even in pools with extremely large liquidity, these "profit" is well below the required amount of transaction fees.

maltekliemann:
[...]

PR #496, opened 2022-03-21
[swaps] Check that user has enough pool tokens on withdraw

maltekliemann:
This is similar, but not the same as issue 1 of the article linked in #485.

### DIFF
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -1105,10 +1105,8 @@ mod pallet {
         ) -> DispatchResult {
             let shares_id = Self::pool_shares_id(pool_id);
             // Check that the account has at least as many free shares as we wish to burn!
-            ensure!(
-                T::Shares::ensure_can_withdraw(shares_id, from, amount).is_ok(),
-                Error::<T>::InsufficientBalance
-            );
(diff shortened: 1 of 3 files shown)


zrml/swaps/src/lib.rs
```diff
@@ -1104,6 +1104,11 @@ mod pallet {
             amount: BalanceOf<T>,
         ) -> DispatchResult {
             let shares_id = Self::pool_shares_id(pool_id);
+            // Check that the account has at least as many free shares as we wish to burn!
+            ensure!(
+                T::Shares::ensure_can_withdraw(shares_id, from, amount).is_ok(),
+                Error::<T>::InsufficientBalance
+            );
```
sea212, line 1111: [nitpick] A little more rustacean way:
```suggestion
            T::Shares::ensure_can_withdraw(shares_id, from, amount)
                .map_err(|_| Error::<T>::InsufficientBalance)?;
```

sea212 (approved):
Nice!

zrml/swaps/src/lib.rs
maltekliemann: Perfect, thanks! 👍

PR #498, opened 2022-03-23
Add tests for `TooManyAssets`/`TooFewAssets`

maltekliemann:
Add two tests which verify that `create_pool` checks for too many/too few assets.

### DIFF
--- a/zrml/swaps/src/tests.rs
+++ b/zrml/swaps/src/tests.rs
@@ -1033,6 +1033,51 @@ fn swap_exact_amount_out_exchanges_correct_values_with_rikiddo() {
     });
 }
 
+#[test]
+fn create_pool_fails_on_too_many_assets() {
+    ExtBuilder::default().build().execute_with(|| {
+        let length = 12;
+        let assets: Vec<Asset<MarketId>> =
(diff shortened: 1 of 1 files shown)


maltekliemann:
Not quite sure why the workflows don't trigger for this one, the tests should run for any PR into main. Can't start manually with `workflow_dispatch`.

zrml/swaps/src/tests.rs
```diff
@@ -1033,6 +1033,51 @@ fn swap_exact_amount_out_exchanges_correct_values_with_rikiddo() {
     });
 }
 
+#[test]
+fn create_pool_fails_on_too_many_assets() {
+    ExtBuilder::default().build().execute_with(|| {
+        let length = 12;
+        let assets: Vec<Asset<MarketId>> =
+            (0..length).map(|x| Asset::CategoricalOutcome(0, x)).collect::<Vec<_>>();
```
sea212, line 1041: We should use the crate configuration value that's specified within `mock.rs` for that
```suggestion
        let length = <Runtime as crate::Config>::MaxAssets::get();
        let assets: Vec<Asset<MarketId>> =
            (0..=length).map(|x| Asset::CategoricalOutcome(0, x)).collect::<Vec<_>>();
```

sea212 (approved):

PR #499, opened 2022-03-23
Bump crate and RuntimeVersion versions

sea212:


runtime/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_version: 34,
```
maltekliemann, line 62: I believe we also need to bump the `spec_version`, right?

sea212: Yes we have to do that between two releases, ideally when the first change in the runtime logic is introduced after a release. The previous release had a `spec_version` that's 1 version lower, so we're good to go :)

maltekliemann: Ok, I see! Thanks!

maltekliemann (approved):

PR #503, opened 2022-03-24
Use BoundedVec instead of WeakBoundedVec

sea212:
We initially used `WeakBoundedVec` instead of `BoundedVec` because of pre-existing vector size concerns. `BoundedVec` is preferred over `WeakBoundedVec` though (due to stricter checks).
Since those concern only target the battery station testnet and in reality are very very unlikely to occur, plus that it is a breaking change later down the line, I decided to fast-track that PR (before mainnet launch)


maltekliemann (approved):
Looks good to go! 👍

PR #504, opened 2022-03-25
Remove `keep_outcome_assets` parameter

maltekliemann:
(Based on an internal discussion) The effects of _not_ keeping excess outcome tokens seem to be rather unclear. With the changes in this PR, the excess outcome tokens are always kept.


zrml/prediction-markets/src/tests.rs
```diff
@@ -661,68 +653,24 @@ fn create_market_and_deploy_assets_is_identical_to_sequential_calls() {
             amount_base_asset.clone(),
             amounts.clone(),
             weights.clone(),
```
sea212, line 655: We can remove all `.clone()` calls within the function parameters of `create_cpmm_market_and_deploy_assets`

sea212 (approved):
Looks good, thanks.

zrml/prediction-markets/src/tests.rs
maltekliemann: Thanks! 👍

Issue #511, opened 2022-03-30
Scalar market doesn't resolve after sudo close


maltekliemann:
Market 384 (scalar market) on Battery Station was created, closed using sudo before its end-of-life, reported, disputed four times, but did not resolve.

PR #513, opened 2022-03-31
Fix `set_pool_as_stale` for scalar markets

### LINKED ISSUE #512: `set_pool_as_stale` returns `WinningAssetNotFound` for scalar market
Most likely related to #511.

maltekliemann:
Fixes #512 and possibly #511.

The function `set_pool_as_stale` does a couple of things:

- Sets pool status
- Destroys losing outcomes
- Distributes Rikiddo winnings

It is, among other things, used in the `resolution_manager`. So it's important that this works correctly with scalar markets.

Unfortunately, scalar markets are not considered, so the default error message `WinningAssetNotFound` is always returned. This has probably prevented scalar markets from resolving correctly (see #511).

I've restructured the function call. No assets are removed from scalar markets (this seems compatible with the way that scalar outcomes are redeemed).

~~Blocked until #510 is merged.~~


maltekliemann:
Note: I haven't yet done the benchmarking. Now that the `set_pool_as_stale` function is split, the common part needs to be benchmarked.

zrml/swaps/src/lib.rs
```diff
@@ -1198,6 +1198,70 @@ mod pallet {
...
+            Self::mutate_pool(pool_id, |pool| {
+                if pool.pool_status == PoolStatus::Stale {
+                    return Ok(());
+                }
+                ensure!(pool.pool_status == PoolStatus::Active, Error::<T>::InvalidStateTransition);
+                pool.pool_status = PoolStatus::Stale;
+                Ok(())
+            })?;
+            Ok(0) // TODO Needs benchmark!
+        }
```
sea212, line 1211: We can get a quite precise estimation by relying on the biggest chunk of weight in this function: DB reads and writes.

```suggestion
        fn set_pool_as_stale_common(pool_id: PoolId) -> Result<Weight, DispatchError> {
            let mut total_weight = T::DbWeight::get().reads(1);

            Self::mutate_pool(pool_id, |pool| {
                if pool.pool_status == PoolStatus::Stale {
                    return Ok(());
                }
                ensure!(pool.pool_status == PoolStatus::Active, Error::<T>::InvalidStateTransition);
                pool.pool_status = PoolStatus::Stale;
                total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
                Ok(())
            })?;

            Ok(total_weight)
        }
```

I am also thinking about allowing the function only to be called on `Active` state to enforce only valid state transitions. This should uncover errors - the market should not try to set the same pool to stale twice, if this happens something unplanned happened

```suggestion
        fn set_pool_as_stale_common(pool_id: PoolId) -> Result<Weight, DispatchError> {
            Self::mutate_pool(pool_id, |pool| {
                ensure!(pool.pool_status == PoolStatus::Active, Error::<T>::InvalidStateTransition);
                pool.pool_status = PoolStatus::Stale;
                Ok(())
            })?;

            Ok(T::DbWeight::get().reads_writes(1, 1))
        }
```

sea212 (approved):
Looks good to me.
It also uncovered that Rikiddo does not correctly handle scalar market, a PR will be created for that.

Issue #514, opened 2022-03-31
`admin_move_market_to_resolved` returns `authorized::UnknownOrigin`


maltekliemann:
When moving a market with MDM `authorized` to resolved, `authorized::UnknownOrigin` is thrown.

sea212:
Solved in #516

PR #516, opened 2022-03-31
Remove resolving undisputed markets from `DisputeApi`

maltekliemann:
Motivation for this PR is #514.

The implementations of `DisputeApi::on_resolution` are inconsistent in how they deal with undisputed markets. `simple-disputes` resolves them, while `authorized` throws a rather obscure `UnknownOutcome` error.

The `prediction-markets` pallet seems to expect the `DisputeApi` to handle undisputed markets (see https://github.com/zeitgeistpm/zeitgeist/blob/05c67d136b87167be2396977ace2aad7d48f6b9f/zrml/prediction-markets/src/lib.rs#L1675-L1685). This means that every implementation of `DisputeApi::on_resolution` needs to check if the market is disputed and, if not, return the reported outcome. This leads to code-duplication or errors like #514.

This PR removes that responsibility from `DisputeApi` and instead handles this case in `prediction-markets`. The only disadvantage to this is that you cannot implement a dispute mechanism which actively prevent a market from resolving without further interaction. No idea if this is a feature that we want to implement somewhere down the line. But right now, I see some gain in preventing interactions between `DisputeApi` and undisputed markets.


sea212 (approved):
Great, that's much better now!

PR #517, opened 2022-03-31
Add missing parachain-staking migration

sea212:
Unfortunately the [`SplitCandidateStateToDecreasePoV` migration](https://github.com/PureStake/moonbeam/pull/1117) was not listed in the overview document, consequently it was not regarded for inclusion. The `SplitCandidateStateToDecreasePoV` migration is mandatory, leaving it out will brick the parachain because it will be unable to select a collator.

This PR also increases the version of `parachain-staking` to avoid a critical bug that requires a future migration and messes up the inflation expectancy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4943,7 +4943,7 @@ dependencies = [
 [[package]]
 name = "pallet-author-mapping"
 version = "2.0.5"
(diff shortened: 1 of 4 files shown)


maltekliemann (approved):
Looks good to me!

Issue #481, opened 2022-03-12
[PM] Audit liquidity functions for CPMM markets


sea212:
The liquidity functions that are used to add and remove liquidity for a pool are in their current state already a part of the runtime since Battery Park testnet. Nevertheless their application was not as thoroughly tested on the test networks as the other operations.
Before mainnet launch, those functions should be audited carefully.

sea212:
Some things I noticed during a fast scan-over, please verify:

- Charges fee twice (swap and exit fee) in specific asset exit
- Single asset exit functions do nothing with the exit fee and all assets exit functions transfer them to the pool. In both cases, they should be burned, otherwise they will be unusable but still increase the total supply.
- Two `ExitFee` definitions that are not coupled: in `consts` module and in pallet `Config`
- `pool_exit` does not regard `ExitFee` when calculating the amount of asset tokens retrieved.

maltekliemann:
Exit fees have been implemented in #518. In particular, we remove the constant `EXIT_FEE`.

Per default, exit fees are substracted from the amount of tokens that the LP receives when withdrawing from the pool. This means the remaining LPs get richer for staying in the pool. In order to prevent exploits, this does not apply when the pool is `Closed` or `Clean`.

The single-asset operations must take trading fees in addition to swaps fees, as they are implemented using pool join/exit operations followed by a series of swaps. To prevent exploits like "trading" using single-asset liquidity operations, these operations are now disabled on `Closed` pools.

PR #518, opened 2022-04-01
Implement and configure exit fees for CPMM

maltekliemann:
This will resolve some of the issues in #481.


maltekliemann:
The exit fee behavior of `pool_exit` required some fixes. Previously, `pool_exit` behaved as follows:

- Disregard exit fees when calculating the amount of assets received
- Transfer the exit fee to the pool account
- Burn the rest of the pool tokens

For example, if a pool contains 100 A, 100 B and 100 ZTG, and Alice own 100 pool shares (with a 100 total issuance), and Alice exits 10 of her pool shares and the exit fee is 10%, then she receives 10 A, 10 B and 10 ZTG (bad, as she is not really charged a fee if she receives 10% of the pool for trading in 10% of the pool shares), 9 pool shares are burnt and 1 is sent to the pool account (why?).

The new behavior is:

- Take exit fee from the amount of assets received
- Burn all pool tokens

For example, if a pool contains 100 A, 100 B and 100 ZTG, and Alice own 100 pool shares (with a 100 total issuance), and Alice exits 10 of her pool shares and the exit fee is 10%, then she receives 9 A, 9 B and 9 ZTG (90% of the 10% share that she is withdrawing) and 10 pool shares are burnt.

Note that with this implementation, the fee is essentially passed to the remaining LPs. This poses an interesting problem: When the market is over and no trading happens anymore, the last LP to withdraw will receive the largest part of the exit fees. With particularly greedy LPs, this will result in a game of chicken over who leaves the pool first. The obvious solution for this is to remove exit fees for stale pools, which seems fine to me.

maltekliemann:
cf3e6ea solves #535 for now. Long-term, I think auto-resolving the winning asset into the pool is the way to go.

zrml/swaps/src/lib.rs
```diff
@@ -1192,6 +1197,17 @@ mod pallet {
                 .cloned()
                 .ok_or(Error::<T>::AssetNotBound)
         }
+
+        /// Calculate the exit fee percentage for `pool`.
+        fn _calc_exit_fee(pool: &Pool<BalanceOf<T>, T::MarketId>) -> BalanceOf<T> {
```
sea212, line 1202: Just wondering what the underscore before the function is for? I know in python this can be used to hint that the function is private, in rust a non-pub function is inherently private though.

zrml/swaps/src/math.rs
```diff
@@ -199,10 +198,11 @@ pub fn calc_single_out_given_pool_in(
     total_weight: u128,
     pool_amount_in: u128,
     swap_fee: u128,
+    exit_fee: u128,
 ) -> Result<u128, DispatchError> {
```
sea212, line 202: If we remove the swap fee here, can this be abused to get an advantage by using pool join / exits as an alternative to using the swap functionality for trade?

sea212: I am also confused why an exit fee is used here, since this is function is only used for joining operations.

zrml/swaps/src/math.rs
```diff
@@ -239,6 +239,7 @@ pub fn calc_pool_in_given_single_out(
     total_weight: u128,
     asset_amount_out: u128,
     swap_fee: u128,
+    exit_fee: u128,
 ) -> Result<u128, DispatchError> {
```
sea212, line 243: Same question in regards to swap fee as above.

zrml/swaps/src/lib.rs
maltekliemann: Yeah, I was inspired by [this](https://docs.substrate.io/how-to-guides/v3/basics/helper-functions/) and my background in Python. But honestly, I don't see the point anymore. I'll remove the underscore (it's also inconsistent wiht the other helper functions).

zrml/swaps/src/math.rs
maltekliemann: > If we remove the swap fee here, can this be abused to get an advantage by using pool join / exits as an alternative to using the swap functionality for trade?

Haven't done the exact testing (I'll do that when swap fees are implemented), but I see this approach. Suppose you have x ZTG and want to buy asset A from the pool, and suppose the pool has the standard weights (1/n for outcome tokens, 1 for ZTG; currently, all pools created via the frontend have this weight). Then you do the following:

- Buy x/2 complete sets
- `pool_join` with exactly x/2 of each asset
- `pool_exit_with_exact_pool_amount` with the pool amount received in the second step and A as the asset

This will cost you transaction fees (constant) and exit fees. I expect you get approximately the same amount as when doing `swap_exact_amount_in` with `x` as the amount, but instead of paying swap fees, you just pay the exit fees.

> I am also confused why an exit fee is used here, since this is function is only used for joining operations.

It's actually used in `pool_exit_with_exact_pool_amount`. The function names are just super confusing here. ;)

sea212: Oh my, indeed!
Okay, I assume the best approach would be to leave the swap fee for single asset exits in until we have more information in regards to possible attacks without it.

sea212 (approved):

PR #534, opened 2022-04-03
Add events for minting/burning pool shares

maltekliemann:
(This is required for subsquid tracking of pool tokens.)

Whenever pool tokens are minted/burned `swaps::Event::PoolTokensMinted`/`swaps::Event::PoolTokensBurned` is emitted.


maltekliemann:
> Whenever pool shares are minted (or burned), the [(orml) currencies pallet will emit an event](https://github.com/open-web3-stack/open-runtime-module-library/blob/7e2f985d8ed49fe0d243e7ec8eb864cd9cc91d2a/currencies/src/lib.rs#L129-L139): [...]

We burn tokens by calling `slash`, which doesn't seem to emit an event. So we at least need the event for burning pool tokens. I guess the event for minting pool tokens is there for symmetry, but I'm fine with removing that.

I created a new event for burning tokens instead of adding the value to an existing event for the sake of abstraction. By letting the helper function for burning tokens emit the event, we basically guarantee that we always emit the necessary event. It's impossible to forget to emit this event. It also slightly protects subsquid from being disturbed by any API changes of `orml_currencies`.

But I had another look at `PoolAssetEvent`. We _could_ put the pool amount in there. What bothers me about that event is that it's not even clear if it's a join or an exit event, right? (Thinking about subsquid again.)

sea212:
Oh yes, I see. `slash` does neither emit an event in `currencies` nor in `tokens` pallet. I think it would be better to include this information in the existing events, since they already exist and contain all other relevant information in regards to a join/exit operation that happened.

maltekliemann:
Blocked until #518 is merged.

sea212:
I just realized that my previous comment was a bit unambiguous. With existing events I was talking about the `PoolJoin*` and `PoolExit` events. They contain all the information in regards to a join and exit operation, except how many `PoolShares` were exchanged for that operation. I think we should add the `PoolShares` amount information into those events, instead of creating new events (or deriving that info from currency events).

https://github.com/zeitgeistpm/zeitgeist/blob/9014b203f41dd7a0ecb2a19ba6ebb86a3663bc60/zrml/swaps/src/events.rs#L12-L26

maltekliemann:
@sea212 Yeah, that makes perfect sense, thanks for clarifying again. I've been working on that. I'm just currently trying to figure out what `pool_join/exit_subsidy` does in order to provide the correct amounts to the event.

sea212:
In a nutshell the providers just reserve an amount of `base_asset` (join) or unreserve an amount (exit) until the market periods start time is reached. When that start time is reached, `end_subsidy_phase` is called (from `process_subsidy_collecting_markets` in prediction markets), which checks if `MinSubsidy` is reached and if yes calculates the initial set of outcome assets that can be minted and locked away in the pool account (remember, outcome assets in Rikiddo are printed and burned and the price impact of a fixed size order is determined by the number of existing outcome assets). The number of `PoolShares` is equal to the provided `base_asset` at that point and it will be distributed during `end_subsidy_phase` if all conditions are met. If they conditions are not met, the `base_asset` that was provided as subsidy is simply unreserved.

maltekliemann:
Thanks for giving this some context. In that case, it doesn't really make sense to me to put any amount of pool shares into the event emitted by the Rikiddo liquidity functions, because the pool shares are only distributed later.

zrml/swaps/src/lib.rs
```diff
@@ -1146,7 +1143,8 @@ mod pallet {
             amount: BalanceOf<T>,
         ) -> DispatchResult {
             let shares_id = Self::pool_shares_id(pool_id);
-            T::Shares::deposit(shares_id, to, amount)
+            T::Shares::deposit(shares_id, to, amount)?;
+            Ok(())
```
sea212, line 1147: This change is unnecessary, since it does the same as `T::Shares::deposit(shares_id, to, amount)` (return `Ok(())` or the error)

zrml/swaps/src/lib.rs
```diff
@@ -1377,6 +1375,7 @@ mod pallet {
                     let mut account_created = false;
                     let mut total_balance = <BalanceOf<T>>::zero();
                     total_assets = pool.assets.len();
+                    let mut providers_and_pool_shares = vec![];
```
sea212, line 1378: Not sure if that is true but we probably can save some time that would be used for reallocations by specifying the capacity of the vector `with_capacity` based on the `decode_len` of the `<SubsidyProviders<T>>::decode_len(pool_id)`.

My assumption is that for small values it's slower, but for big values it faster, but where is the border? We definitely lose about 25µs for the extra storage read to get the length beforehand.

zrml/swaps/src/lib.rs
```diff
@@ -1392,9 +1391,10 @@ mod pallet {
                                 &pool_account,
                                 subsidy,
                             )?;
-                            total_balance = subsidy;
+                            total_balance = total_balance.saturating_add(subsidy);
```
sea212, line 1394: This changes nothing, because for the first account adding and setting are equivalent.

sea212 (commented):
The changes look fine overall.
Please add information about the changes and new events and when exactly `PoolShares` are generated to [docs/changelog_for_devs.md](https://github.com/zeitgeistpm/zeitgeist/blob/ef0a9a199588e88bf231e555ac2625190e99699e/docs/changelog_for_devs.md).

zrml/swaps/src/lib.rs
maltekliemann: Good catch, thanks!

zrml/swaps/src/lib.rs
maltekliemann: For a while there was a `deposit_event` call between those, and I forgot to put things back to the way they were 🙈

zrml/swaps/src/lib.rs
maltekliemann: Hmm... When there's time, I guess we could benchmark this.

Ok, maybe another thing that's related here: I was thinking if there should be a maximum number of subsidy providers, let's say 100. In that case, we could just go with that number as capacity.

sea212: That idea is actually not that bad. We'd need a far bigger number to avoid users creating multiple accounts and clogging the list with minimal amounts.
I think we can keep that in mind for the future.

maltekliemann:
Thanks for the review so far! :thumbsup: Just realized that there are two cases missing:

- Amount of pool shares issued to the market creator
- Amount of unreserved subsidy on insufficient subsidy

I won't finish that until tomorrow, I'm afraid.

maltekliemann:
See the CHANGELOG for details on what changes I've made.

Not super happy with this. It seems like there are multiple events emitted when the subsidy phase ends, some emitted from `prediction-markets`, others from `swaps`.

Altogether, I don't get the substrate event system. So many low-level functions (like `Currencies::slash`) throw events without much semantic meaning, which forces us to throw new events explaining the other events (explain that the _slash_ measn that pool tokens were burned). Feels like unnecessary clutter. Also, I see potential dangers of double-booking with subsquid. For example, let's say there's a slash event and an event explaining that pool tokens are burned. How does subsquid know that this refers to the slash event? Couldn't the slash event be completely unrelated to the burn event?

sea212:
I agree, this is very confusing and increases the probability to make errors. I have the following idea:
1. Those events are purely informative. They will be logged in SubSquid, but do not trigger balance changes
2. Since [completing the token events and avoiding event duplication was confirmed by ORML maintainers](https://github.com/open-web3-stack/open-runtime-module-library/issues/726), we can agree on that only the events from the Tokens pallet and the Balances pallet will be taken into account to calculate balance changes. The events from us are purely informal and should just log what happened in the pallet.
3. We create a fork until https://github.com/open-web3-stack/open-runtime-module-library/issues/726 is resolved.

zrml/swaps/src/lib.rs
sea212: In addition to that, the number must be reasonably low such that it won't cause overweight blocks that halt the chain.

maltekliemann: Yes. I guess we could combine it with a minimum amount of subsidy per provider.

maltekliemann:
Sure, I can create the fork. One thing, though:

> The original idea is that events are emitted by the extrinsic pallet (currencies), not the underlying implementation (tokens) so that people can decide what event to emit. [https://github.com/open-web3-stack/open-runtime-module-library/issues/726#issuecomment-1097416750]

So we could just call the underlying implementation and prevent the "low-level" events from being emitted entirely?

sea212:
Yes, but it is a bit inconsistent. For example, you could call `transfer` from the `Tokens` pallet directly, it would do the transfer and not emit an event. Then with `reserve`, it would emit the event.
I think the ideal state would be (as discussed in ORML) that every pallet that changes balances (such as Tokens or Balances), excluding pallets that do the routing (such as Currencies), emits the respective event when doing so. This way we will always have a complete picture of the balance changes. 
Our events that contain balances (such the join/exit events) should be purely informative in that regard in my opinion. The real balance changes that the external block processors apply occur only when the pallet that is responsible to maintain the balances signals a change by emitting an event.

zrml/swaps/src/lib.rs
```diff
@@ -1308,10 +1311,19 @@ mod pallet {
...
+                    pool,
+                    amount,
+                ));
+            } else {
+                Self::deposit_event(Event::PoolCreate(
+                    CommonPoolEventParams { pool_id: next_pool_id, who },
+                    pool,
+                    0u32.saturated_into(),
+                ));
+            }
```
sea212, line 1326: We could use a ternary operator here to improve style. I am fine to keep it like that, that PR already demanded enough effort from us.

```suggestion
            Self::deposit_event(Event::PoolCreate(
                CommonPoolEventParams { pool_id: next_pool_id, who },
                pool,
                if scoring_rule == ScoringRule::CPMM { amount } else { 0u32 },
            ));
```

sea212 (approved):

PR #538, opened 2022-04-05
Remove nimbus patch and parachain-staking migrations

### LINKED ISSUE #459: Remove patch instructions for nimbus-consensus and nimbus-primitives
Within the v0.3.0 release we use a hotfix that allows our mainnet nodes to use the old nimbus api that is stored within the latest deployed runtime. Once v0.3.0 was released and the runtime is deployed, we should remove the patched projects and replace them by the originals.

The code that has to be removed can be found here: https://github.com/zeitgeistpm/zeitgeist/blob/fcaa54446db07c1b90b879166a1f7385049643a6/Cargo.toml#L25-L27

sea212:
closes #459 

Also removes parachain-staking migrations. The patch will be removed after dependency upgrade to v0.9.18 (due to usage of `parity-scale-codec v3`
In addition, a document is added that contains all the changes that affect developers who interact with our node directly. Internally this will help SDK, subsquid and frontend devs to figure out how to adjust for the next release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4595,7 +4595,7 @@ dependencies = [
 [[package]]
 name = "nimbus-consensus"
 version = "0.9.0"
-source = "git+https://github.com/zeitgeistpm/nimbus?branch=moonbeam-polkadot-v0.9.16#24f7c50ed1e8652f7f68342c9cb12f52df985079"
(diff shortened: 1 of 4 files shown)


maltekliemann (approved):
Looks good to me 👍

PR #545, opened 2022-04-07
Add market state diagram

sea212:


saboonikhil:
@sea212, amazing stuff will be useful for subsquid 💥

PR #553, opened 2022-04-08
Check that report matches market type and range

### LINKED ISSUE #552: [prediction_markets] Users can report outcomes not present in the market
For example, scalar outcomes can be reported to a categorical market. The following cases must be checked:

- Market type
- Market ID
- (For categorical markets) is the outcome available in the market

maltekliemann:
Fixes #552.

Note that we remove the check that scalar markets are within range. Any value outside of the range is already treated as if saturated into the range, and it makes more sense to allow users to report the actual number instead of having to do the saturation themselves.

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1562,15 +1562,11 @@ mod pallet {
                     return Err(Error::<T>::OutcomeMismatch.into());
                 }
             }
-            if let OutcomeReport::Scalar(ref inner) = outcome {
-                if let MarketType::Scalar(ref outcome_range) = market.market_type {
-                    ensure!(
-                        inner >= outcome_range.start() && inner <= outcome_range.end(),
-                        Error::<T>::OutcomeOutOfRange
-                    );
-                } else {
-                    return Err(Error::<T>::OutcomeMismatch.into());
-                }
+            if let OutcomeReport::Scalar(_) = outcome {
+                ensure!(
+                    matches!(&market.market_type, MarketType::Scalar(_)),
+                    Error::<T>::OutcomeMismatch
+                );
             }
             Ok(())
         }
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -939,10 +939,8 @@ mod pallet {
             let market_report = Report { at: current_block, by: sender.clone(), outcome };
 
             T::MarketCommons::mutate_market(&market_id, |market| {
-                // TODO make this a conditional check
-                // ensure!(outcome <= market.outcomes(), Error::<T>::OutcomeOutOfRange);
                 ensure!(market.report.is_none(), Error::<T>::MarketAlreadyReported);
-
+                Self::ensure_outcome_matches_market_type(market, &market_report.outcome)?;
                 Self::ensure_market_is_closed(&market.period)?;
 
                 let mut should_check_origin = false;


maltekliemann:
Remaining tasks:

- Add a test that verfies that scalars outside of the outcome range evaluate correctly
- Add tests that verify that incorrect outcomes can not longer be reported

zrml/prediction-markets/src/tests.rs
```diff
@@ -871,6 +948,100 @@ fn full_scalar_market_lifecycle() {
...
+        let assets = PredictionMarkets::outcome_assets(0, &MarketCommons::market(&0).unwrap());
+        for asset in assets.iter() {
+            assert_eq!(Tokens::free_balance(*asset, &CHARLIE), 0);
+            assert_eq!(Tokens::free_balance(*asset, &EVE), 0);
+        }
+        assert_eq!(Balances::free_balance(&CHARLIE), 1000 * BASE);
+        assert_eq!(Balances::free_balance(&EVE), 1000 * BASE);
+    })
+}
+
```
sea212, line 1044: Nothing wrong here, just a lot of common code that could be put into a common function

zrml/prediction-markets/src/lib.rs
```diff
@@ -1564,15 +1562,11 @@ mod pallet {
...
-                        Error::<T>::OutcomeOutOfRange
-                    );
-                } else {
-                    return Err(Error::<T>::OutcomeMismatch.into());
-                }
+            if let OutcomeReport::Scalar(_) = outcome {
+                ensure!(
+                    matches!(&market.market_type, MarketType::Scalar(_)),
+                    Error::<T>::OutcomeMismatch
+                );
```
sea212, line 1569: Based on the tests I see that the reported outcome is saturated into the min/max value. However, this is an assumption that relies on another part of the code and I have the hunch that it would be better to enforce the saturation within the `report` function.

sea212 (approved):

zrml/prediction-markets/src/lib.rs
sea212: Not by throwing an error, but by saturating it within the code

maltekliemann: The reported value is not saturated into the market range. And that's fine by me, because (for transparency) the reported outcome should not depend on the market range. Instead, `redeem` saturated the value into the market range when calculating winnings. So I don't think there's any code that depends on the outcome report being within range.

That's why I removed the check completely.

sea212: I see the argument. From a tech perspective I'd like to saturate it immediately to avoid errors later down the line. From a practical perspective it is important to not modify the reported to not distort the truth (reported by the oracle). I think we can leave it like this and just assume that `redeem` will always respect the saturation.

zrml/prediction-markets/src/tests.rs
maltekliemann: :thumbsup: Yeah, resolved!

zrml/prediction-markets/src/lib.rs
maltekliemann: Right now, the tests that verify this behavior aren't really unit tests, but rather integration tests that verify the correct interaction between `report` and `redeem`. Still, I feel fairly safe right now.

When we make the entire approach a bit more object-oriented, maybe we can find a better middle ground.

PR #557, opened 2022-04-11
Add Hadolint and Trivy Checks

samuelarogbonlo:


sea212 (commented):
This is just a fast scan-over opinion, proper review coming soon.

The major thing I see is that we do the checks when we want to publish the docker images. I assume an lint/security issue will stop the publishing of the image, won't it?
I think it would be better to check those before a PR is merged into main. In regards to that, it would be nice to triggers those checks only when relevant changes are made.

samuelarogbonlo:
Hello @sea212 can you please take a look at the changes I've made so far? And possibly take a look at the results of the job in https://github.com/zeitgeistpm/zeitgeist/actions/runs/4145477882/jobs/7169887429

sea212 (approved):

.github/workflows/docker-hub.yml
```diff
@@ -5,9 +5,21 @@ on:
   push:
     branches: [ main ]
     tags: [ '*' ]
+  pull_request:
+    branches: [ main ]
```
sea212, line 9:
```suggestion
```

sea212 (changes requested):

sea212 (approved):

Issue #500, opened 2022-03-24
[Benchmark] malformed weight files


sea212:
The output files generated by the benchmark subcommand generate malformed weight files. Semantically they are correct, syntactically they are a disaster.
Take the resulting weight file for pallet_balances as an example:
```rust
  GNU nano 5.6.1                                                                                                                   pallet_balances.rs                                                                                                                             
//! Autogenerated weights for pallet_balances
//!
//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
//! DATE: 2022-03-23, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
//! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("dev"), DB CACHE: 1024

// Executed Command:// ./target/release/zeitgeist// benchmark// --chain=dev// --steps=50// --repeat=20// --pallet=pallet_balances// --extrinsic=*// --execution=wasm// --wasm-execution=compiled// --heap-pages=4096// --template=./misc/frame_weight_template.hbs// --output=./r>
#![cfg_attr(rustfmt, rustfmt_skip)]
#![allow(unused_parens)]
#![allow(unused_imports)]

use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
use core::marker::PhantomData;

/// Weight functions for pallet_balances (automatically generated)
pub struct WeightInfo<T>(PhantomData<T>);
impl<T: frame_system::Config> pallet_balances::weights::WeightInfo for WeightInfo<T> {  // Storage: System Account (r:1 w:1)    fn transfer() -> Weight {
                (105_740_000 as Weight)                 .saturating_add(T::DbWeight::get().reads(1 as Weight))                  .saturating_add(T::DbWeight::get().writes(1 as Weight)) }       // Storage: System Account (r:1 w:1)    fn transfer_keep_alive() -> Weight {
                (81_100_000 as Weight)                  .saturating_add(T::DbWeight::get().reads(1 as Weight))                  .saturating_add(T::DbWeight::get().writes(1 as Weight)) }       // Storage: System Account (r:1 w:1)    fn set_balance_creating() -> Weight {
[...]

PR #564, opened 2022-04-14
Fix weight formatting

sea212:
closes #500


maltekliemann (approved):
Yeah, looks much better this way. All good!

PR #565, opened 2022-04-14
Disable Rikiddo and Court market creation

### LINKED ISSUE #451: [PM] Disable Rikiddo on mainnet
Rikiddo was not tested enough and consequently the dangers of using it with real value are too high.

Implement a solution that is flexible enough to enable the creation of Rikiddo markets for the Battery Station testnet but disable them on the mainnet.

### LINKED ISSUE #452: [PM] Disable court dispute mechanism on mainnet
A couple of bugs have been found that render the court dispute mechanism too unreliable to be used on mainnet.

Implement a solution that is flexible enough to enable the creation of markets that use the court dispute mechanism for the Battery Station testnet but disables it on the mainnet.

sea212:
closes #451 
closes #452 

**Solution**
A call-filter is implement that is used when the runtime is built without the `txfilter` feature. This call filter rejects market creation where Rikiddo `ScoringRule` or Court `MarketDisputeMechanism` are selected:

Currently this disables this on both the testnet and the mainnet. Since we have only one runtime for both, this can't be controlled without additional features.
During the next release it is planned to separate the testnet and mainnet runtime. At this point, the functionality will be unlocked again for the testnet.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -326,7 +326,7 @@ cfg_if::cfg_if! {
                     | Call::Court(_)
                     | Call::LiquidityMining(_)
                     | Call::Swaps(_)
-                    | Call::PredictionMarkets()
+                    | Call::PredictionMarkets(_)
                     | Call::Vesting(_) => false,
                 }
             }
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -326,7 +326,7 @@ cfg_if::cfg_if! {
                     | Call::Court(_)
                     | Call::LiquidityMining(_)
                     | Call::Swaps(_)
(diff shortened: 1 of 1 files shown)


maltekliemann (approved):
Looks good! :thumbsup:

Stricly speaking, this of course doesn't disable all Rikiddo-related functions. For example, `process_subsidy_collecting_markets` will still run every block. But this has always been the case, right?

sea212:
Yes that's right. Since no Rikiddo markets can be created, it won't even start the first iteration because there are no subsidy collecting markets. We lose a tiny bit of weight for that, but I guess that's OK comparing the alternative methods and the effort and impact to apply those.

PR #566, opened 2022-04-14
Disable liquidity mining

### LINKED ISSUE #467: [LiquidityMining] Disable liquidity mining on mainnet
The feature is too unstable for mainnet, see #370

sea212:
closes #467 

This disables liquidity mining on all networks currently by replacing the liquidity mining pallet by a pallet that implements the liquidity mining pallet api with no-ops.
Once we split the runtimes into testnet and mainnet (probably next release), we'll enable LM on the testnet and keep it disabled on the mainnet.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -366,9 +366,9 @@ cfg_if::cfg_if! {
                 }
             }
         }
-    // Unrestricted (no "txfilter" feature) chains
-    // Currently disables Rikiddo and Court markets, will be relaxed
-    // for testnet once runtimes are separated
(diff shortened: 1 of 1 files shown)


maltekliemann:
I believe `LiquidityMining` also needs to be replaced in the `Swaps` pallet and the benchmarks.

Other than that, this seems good to go :thumbsup:

sea212:
Nice catch, it indeed has to be replaced in the `Swaps` pallet.
Changing the benchmarks is not necessary though, since it only applies to the dispatchable calls in liquidity mining. Thanks for pointing this out, this is a place where the weights are not calculated properly.

PR #571, opened 2022-04-15
Benchmark weights

sea212:


runtime/src/weights/frame_system.rs
```diff
@@ -19,9 +31,8 @@ use frame_support::{
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> frame_system::weights::WeightInfo for WeightInfo<T> {
     fn remark(_b: u32) -> Weight {
-        (4_155_000 as Weight)
+        (0 as Weight)
     }
```
sea212, line 35: Looks strange, but `remark` does basically nothing but checking the origin. [fn remark](https://github.com/paritytech/substrate/blob/22d40c761a985482f93bbbea5ba4199bdba74f8e/frame/system/src/lib.rs#L381-L385):
```rust
pub fn remark(origin: OriginFor<T>, _remark: Vec<u8>) -> DispatchResultWithPostInfo {
	ensure_signed_or_root(origin)?;
	Ok(().into())
}
```
Theoretically this can be abused to stall the change, but it will cost a lot of fees.

runtime/src/weights/frame_system.rs
```diff
@@ -19,9 +31,8 @@ use frame_support::{
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> frame_system::weights::WeightInfo for WeightInfo<T> {
     fn remark(_b: u32) -> Weight {
-        (4_155_000 as Weight)
+        (0 as Weight)
```
sea212, line 34:
```suggestion
        (4_155_000 as Weight)
```

runtime/src/weights/frame_system.rs
```diff
@@ -19,9 +31,8 @@ use frame_support::{
...
     }
-
     fn remark_with_event(b: u32) -> Weight {
         (0 as Weight)
```
sea212, line 37:
```suggestion
        (12_000_000 as Weight)
            .saturating_add(T::DbWeight::get().writes(1 as Weight))
```

PR #573, opened 2022-04-18
Add high-level component diagram

### LINKED ISSUE #490: Add architecture overview
Usually projects have a high-level overview of the architecture in the repository or documentation site. We should have one, too.

sea212:
closes #490

Also updates the state diagram.


sea212:
Closes #490

Chralt98 (commented):
Nice graphics, well done. I checked the `puml` files with https://www.planttext.com/. They are correct. I checked the dependencies for all pallets to verify the `high_level_component_diagram`. `orderbook-v1` is not used anymore, ok. 
I think `liquidity-mining` uses https://github.com/zeitgeistpm/zeitgeist/blob/49ab68a2c488bfad2a3beeae84264931965389c3/zrml/liquidity-mining/src/lib.rs#L96-L100 `MarketAPI` too.

Chralt98 (approved):
Looks good to me. Waiting for you @vivekvpandya

vivekvpandya:
Looks good. I just have a question so for Subsidy base market in case of insufficient subsidy case market is suspended so there is another flow to get it back somehow ?

vivekvpandya (approved):

sea212:
Good question @vivekvpandya . If the market did not collect enough subsidy, it is regarded as "too low interest / quality not sufficient" and it will be discarded. There is no way to get it back (currently it is a storage zombie, we'll have to clean those up eventually). If someone wants to have this market, they'll have to open it again.

PR #574, opened 2022-04-19
Rewrite `authorized` pallet, remove `markets` from `market-commons`

maltekliemann:
Apologies for the long PR. :pray:

- The following changes were made to `DisputeApi`:
  - Clarify the API. In particular, we allow `on_dispute` to not return an error even if the market unexpectedly has the incorrect MDM. I don't think we should make it a responsibility of `DisputeApi` to verify the correctness of the state of the market primitive.
  - Change the return value of `DisputeApi::on_resolution` from `Result<OutcomeReport, DispatchError` to `Result<Option<OutcomeReport>, DispatchError>`; the specification now states that `None` is returned when the MDM was unable to resolve the dispute. This could happen, for example, if the authorized account doesn't report back in time.

- The `prediction-markets` pallet was changed accordingly: If `on_resolution` returns `None`, the original report by the oracle is used.

- The `markets` function of the `market-commons` pallet is removed. It was only used in one place, namely the `authorized` pallet, where its use has been removed (see below). This resolves zeitgeistpm/runtime-audit-1#2 and zeitgeistpm/runtime-audit-1#4.

- The authorized pallet suffered a major rewrite. The use of `markets` has been removed. We store outcomes reported by the authorized account in a `HashMap<MarketIdOf<T>, OutcomeReport>` which maps market ids to the reports submitted for them. Some notes on the functionality:
  - An authorized account can overwrite their previous report within the 24h dispute window as often as they want
  - We ensure that the market has state `Disputed`, and that the reported outcome matches the market type
  - We ensure that the reporter is the authorized account, and that the market actually uses the `authorized` MDM
  The rewrite makes it possible to have the same account as authorized user for multiple markets at the same time. This fixes zeitgeistpm/runtime-audit-2#1. We demonstrate that `authorized` plays well with multiple markets in the test `authorize_market_outcome_allows_using_same_account_on_multiple_markets`. Note that this test also mitigates the concern expressed in the final paragraph of said issue. I've also added an integration test for `authorized` to the tests of the `prediction-markets` pallet.

- In order to be able to verify that the outcome reported by the authorized account matches the market type, it was necessary to use functionality already present in `prediction-markets`, namely `ensure_outcome_matches_market_type`. So I moved this function out of the pallet and into the `primitives` module, where it is now a method of `Market`. We use `test-case` for testing, which seems to be rather useful for these types of unit tests. The only possible downside is that we lose a bit of granularity in our error reporting: The error `OutcomeOutOfRange` is replaced by the already present `OutcomeMismatch`. I'm okay with that.

I will add origin based authorization in another PR. I think this one is large enough.


zrml/authorized/src/lib.rs
```diff
@@ -173,14 +151,9 @@ mod pallet {
...
-        _,
-        Blake2_128Concat,
-        MarketIdOf<T>,
-        Blake2_128Concat,
-        T::AccountId,
-        OutcomeReport,
-    >;
+    #[pallet::getter(fn outcomes)]
+    pub type Outcomes<T: Config> =
+        StorageMap<_, Blake2_128Concat, MarketIdOf<T>, Option<OutcomeReport>, ValueQuery>;
```
sea212, line 156: We can use [`OptionQuery`](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.OptionQuery.html) here.

sea212, line 156: This change demands a migration.

sea212 (changes requested):

zrml/authorized/src/lib.rs
```diff
@@ -181,6 +159,12 @@ mod pallet {
         T::AccountId,
         OutcomeReport,
     >;
+
+    /// Maps the market id to the outcome reported by the authorized account.    
+    #[pallet::storage]
+    #[pallet::getter(fn outcomes)]
+    pub type AuthorizedOutcomeReports<T: Config> =
+        StorageMap<_, Blake2_128Concat, MarketIdOf<T>, OutcomeReport, OptionQuery>;
```
sea212, line 167: We can use the significantly faster and more light-weight hasher [xx64](https://docs.substrate.io/rustdocs/latest/frame_support/struct.Twox64Concat.html). Since the users don't have direct control over the keys, a cryptographic hash function like `Blake2_128` is not necessary.

zrml/authorized/src/migrations.rs
```diff
@@ -0,0 +1,67 @@
...
+    where
+        T: Config,
+    {
+        if StorageVersion::get::<Pallet<T>>() != REQUIRED_STORAGE_VERSION {
+            return 0;
+        }
+        let mut total_weight: Weight = 0;
+        log::info!("Starting storage migration of authorized reports");
+
+        for (market_id, _, outcome_report) in <Outcomes<T>>::iter() {
```
sea212, line 27: We want to empty the previous storage to avoid having a "zombie storage" (A storage that is not exposed anymore but contains data). The function [`drain`](https://docs.substrate.io/rustdocs/latest/frame_support/pallet_prelude/struct.StorageDoubleMap.html#method.drain) can be used for that.

zrml/authorized/src/migrations.rs
```diff
@@ -0,0 +1,67 @@
...
+pub struct MigrateAuthorizedStorage<T>(PhantomData<T>);
+
+// Due to problems with `storage_key_iter`, we temporarily use two maps. The old `Outcomes` will
+// no longer be used and removed at the next opportunity.
+impl<T: Config> OnRuntimeUpgrade for MigrateAuthorizedStorage<T> {
+    fn on_runtime_upgrade() -> Weight
+    where
+        T: Config,
+    {
+        if StorageVersion::get::<Pallet<T>>() != REQUIRED_STORAGE_VERSION {
```
sea212, line 21: I think we have to add weight for one storage read in case of success.

sea212 (changes requested):

zrml/authorized/src/lib.rs
maltekliemann: Fixed in [868f737](https://github.com/zeitgeistpm/zeitgeist/pull/574/commits/868f73725af190be47d052930efac7669fbe91a2)

zrml/authorized/src/migrations.rs
maltekliemann: Fixed in 6745af9

zrml/authorized/src/migrations.rs
maltekliemann: I think we have to add a read in any case, even if we abort, right?

maltekliemann: See aab45b7. One read when checking the storage version. A read and a write for each item in `Outcomes`, then a write for setting the new storage version.

sea212 (approved):

Issue #554, opened 2022-04-10
[prediction_markets] `MarketStatus::Closed` is unused, but appears in tests


maltekliemann:
First of all, it appears as though `MarketStatus::Closed` is unused in the pallet. Instead `ensure_market_is_closed` regulates which actions can be taken when the block number exceeds the market's period. This should be fixed, probably by using `Closed`.

Furthermore, `MarketStatus::Closed` is used in the tests of the pallet. This means that tests rely on storage states that are impossible to achieve.

PR #577, opened 2022-04-20
Ensure that market can only be reported when closed

maltekliemann:
Fixes zeitgeistpm/runtime-audit-2#3. This is also related to #554.


maltekliemann:
We talked about starting to use `MarketStatus::Closed`. I would like to do this in a separate PR. This PR is only for fixing functionality.

sea212 (approved):

PR #578, opened 2022-04-21
Make `MarketIdOf`, `BalanceOf` & `PoolId` compact

maltekliemann:


maltekliemann:
Fixes zeitgeistpm/runtime-audit-1#16

maltekliemann:
Question to the reviewers: It's possible to make some other parameters compact, at the cost of essentially having tuples as parameters: https://substrate.stackexchange.com/questions/2240/can-scale-compact-encoding-be-used-with-optiont-if-so-how-exactly

Should we apply this to some of our parameters?

Chralt98:
I looked into many repositories including polkadot and found no extrinsic call parameter with the use of [`Compact`](https://docs.rs/parity-scale-codec/3.1.2/parity_scale_codec/struct.Compact.html).

>This is more space-efficient but less compute-efficient."

I would rather go for compute-efficiency in extrinsic calls. I mean the `StorageMap, StorageDoubleMap` and `StorageValue` elements are much more important to limit in size, at least rather than the function parameter. Quite apart from the tuple notation and complexity when using `Compact`.

maltekliemann:
> This is more space-efficient but less compute-efficient.

Thanks for raising this point. I saw that earlier but ignored it.

The claim is pretty vague. I guess the idea is that the extrinsic becomes shorter, thus saving the user length fees, but takes longer to compute, resulting in higher weight fees. And hopefully, the gain from the length fees is larger than the loss from the weight fees.

zrml/prediction-markets/src/lib.rs
```diff
@@ -357,7 +357,7 @@ mod pallet {
         #[pallet::weight(T::WeightInfo::cancel_pending_market())]
         pub fn cancel_pending_market(
             origin: OriginFor<T>,
-            market_id: MarketIdOf<T>,
+            #[pallet::compact] market_id: MarketIdOf<T>,
         ) -> DispatchResult {
             let sender = ensure_signed(origin)?;
```
Chralt98, line 363: In `create_categorical_market` parameter `categories` is `u16` and can be compacted too.

runtime/src/lib.rs
```diff
@@ -60,10 +60,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("zeitgeist"),
     impl_name: create_runtime_str!("zeitgeist"),
     authoring_version: 1,
-    spec_version: 35,
+    spec_version: 36,
```
Chralt98, line 63: Why this is required?

runtime/src/lib.rs
```diff
@@ -60,10 +60,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
...
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 12,
+    transaction_version: 13,
```
Chralt98, line 66: Why this is required?

runtime/src/lib.rs
Chralt98: Is it, because this commit can go straight to production (otherwise it wouldn't get accepted as an updated WASM file)?

docs/changelog_for_devs.md
```diff
@@ -1,3 +1,9 @@
+# v0.3.2
+
+- Changed the all parameters of type `BalanceOf`, `MarketId` and `PoolId` in
```
Chralt98, line 3: Is it required to mention the one `#[pallet::compact] categories: u16` parameter of the prediction markets function `create_categorical_market`? (look review comment below)

Chralt98 (commented):
I checked all extrinsic calls which allow to use `#[pallet::compact]` for more compact parameters and found only one additional parameter which can be compacted too.

Good work @maltekliemann!

zrml/prediction-markets/src/lib.rs
maltekliemann: I didn't change `cancel_pending_market`, because that's on its way out the door. But good catch regarding `u16`, I didn't think about that.

runtime/src/lib.rs
maltekliemann: You can find some info here: https://docs.substrate.io/rustdocs/latest/sp_version/struct.RuntimeVersion.html

Basically, we've changed the transactional interface, so we need to update that version. In particular, we need to update the spec version.

You might be wondering why this wasn't done for the PR where `cancel_pending_market` was removed. I should have incremented the versions there, as well. It doesn't matter though, since this is all part of a single update. I've setup a checklist that we usually use to keep track of these things: https://github.com/zeitgeistpm/zeitgeist/issues/597

runtime/src/lib.rs
maltekliemann: See above.

docs/changelog_for_devs.md
maltekliemann: Yeah, we should mention `u16`

docs/changelog_for_devs.md
```diff
@@ -1,5 +1,8 @@
 # v0.3.2
 
+- Changed the all parameters of type `u16`, `BalanceOf`, `MarketId` and `PoolId`
```
Chralt98, line 3:
```suggestion
- Changed all call parameters of type `u16`, `BalanceOf`, `MarketId` and `PoolId`
```
[nitpick]

Chralt98 (approved):

PR #580, opened 2022-04-21
Fix unreserving of validity bond for advised market

maltekliemann:
Fixes zeitgeistpm/runtime-audit-1#35

PR #576, opened 2022-04-20
Make extrinsics `#[transactional]`

maltekliemann:
We make all extrinsics `#[transactional]` unless they are not already explicitly marked as non-transactional. We now use `MARK(non-transactional): description...` as format for marking extrinsics which use verify-first-write-second.

Fixes zeitgeistpm/runtime-audit-1#17, fixes zeitgeistpm/runtime-audit-1#18, fixes zeitgeistpm/runtime-audit-1#24 and fixes zeitgeistpm/runtime-audit-2#7.

PR #586, opened 2022-04-22
Slash bonds when destroying market

maltekliemann:
Again, tests aren't super detailed, but I believe I have covered everything from the table of zeitgeistpm/runtime-audit-1#34. We should also look into how to better check that slashes succeed.

Fixes zeitgeistpm/runtime-audit-1#34 and zeitgeistpm/runtime-audit-1#43.


zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
 
             let mut total_accounts = 0usize;
             let mut share_accounts = 0usize;
             let market = T::MarketCommons::market(&market_id)?;
             let market_status = market.status;
             let outcome_assets = Self::outcome_assets(market_id, &market);
             let outcome_assets_amount = outcome_assets.len();
+
+            // Slash outstanding bonds
+            if market_status == MarketStatus::Proposed {
```
Chralt98, line 159: So the AdvisoryBond needs only to get slashed in the case `no state -> Proposed for advised`

`permissionless` has `N/A` (not available?) so no state which would lead to the proposed market status.

Yeah, then this case would only get called when it is advised. But how do you ensure, that it has "no state" before? Because there is no other possibility for "advised" to get into the `proposed` state?

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
+            }
+            if market_status != MarketStatus::Resolved {
+                if market.creation == MarketCreation::Permissionless {
+                    CurrencyOf::<T>::slash_reserved_named(
+                        &RESERVE_ID,
+                        &market.creator,
+                        T::ValidityBond::get(),
+                    );
+                }
+                CurrencyOf::<T>::slash_reserved_named(
```
Chralt98, line 174: Correct me if I am wrong, but the OracleBond is slashed in every case (excluding `resolved`). 

True, most of the time the OracleBond is still missing to slash, according to the table.

Take for example the `* -> InsufficientSubsidy` state change. The OracleBond is already slashed (`-O` in table) and you slash it again. Two is better, yes, but is this wanted? 

 Why do you excluding the `resolved` market status? I can't read that in the table (maybe `Reported/Disputed -> Resolved` but it makes no sense).

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
+            // Slash outstanding bonds
+            if market_status == MarketStatus::Proposed {
+                CurrencyOf::<T>::slash_reserved_named(
+                    &RESERVE_ID,
+                    &market.creator,
+                    T::AdvisoryBond::get(),
+                );
+            }
+            if market_status != MarketStatus::Resolved {
+                if market.creation == MarketCreation::Permissionless {
```
Chralt98, line 167: Yeah it makes sense to only slash the `ValidityBond` in permissionless mode, but why do you exclude the `resolved` market status, when it is only required to slash from `proposed` to `active` and from `proposed` to `CollectingSubsidy`. 

I would expect, that this code slashes too often.

Chralt98 (commented):
Thank you, @maltekliemann for giving me the opportunity to review your code. 

There are multiple things I don't really get in this proposed code. I could be totally wrong, but is it not required to slash only the created bonds? I mean it seems, that this code would slash multiple times, also when it is not really required (double slash). It would be nice to write tests that actually check the `reserved_balance` before and after the calls.

zrml/prediction-markets/src/lib.rs
```diff
@@ -38,19 +42,26 @@
 //! #### Admin Dispatches
 //!
 //! The administrative dispatches are used to perform admin functions on chain. Currently, the
-//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `DestroyOrigin` and `ResolveOrigin`
+//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `ResolveOrigin` and
```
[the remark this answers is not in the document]
maltekliemann: Ok, if by _admin functions_ we mean the `admin_*` functions, then yeah, the approval comittee doesn't have access to those.

zrml/prediction-markets/src/lib.rs
```diff
@@ -116,10 +127,10 @@ mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
-        /// Allows the `DestroyOrigin` to immediately destroy a market.
+        /// Destroy a market, including its outcome assets.
```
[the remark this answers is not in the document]
maltekliemann: We're not 100% settled on what `destroy_market` should actually do, but the general idea is that it's a last resort and should only be used when a market creator has serious misbehaved (assassination markets come to mind). Therefore, we've first of all decided to slash the market creator's funds.

Other than that, yes, _all_ outcome tokens are slashed. We're not happy with this, because it means that everyone who participated in the market (traded/LPed) loses their outcome tokens. Now that I think about it, I'm not sure what happens with the ZTG of the pool. Probably stuck in the pool account forever. I'll have to look into that.

zrml/prediction-markets/src/lib.rs
```diff
@@ -116,10 +127,10 @@ mod pallet {
...
         ///
-        /// todo: this should check if there's any outstanding funds reserved if it stays
-        /// in for production
+        /// Must be called by `DestroyOrigin`. Bonds (unless already returned) are slashed without
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -38,19 +42,26 @@
...
 //!
 //! The administrative dispatches are used to perform admin functions on chain. Currently, the
-//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `DestroyOrigin` and `ResolveOrigin`
+//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `ResolveOrigin` and
+//! root.
 //!
 //! - `admin_destroy_market` - Destroys a market and all related assets, regardless of its state.
-//! - `admin_move_market_to_closed` - Immediately moves a market that is an `Active` state to closed.
-//! - `admin_move_market_to_resolved` - Immediately moves a market that is `Reported` or `Disputed` to resolved.
+//!   Can only be called by sudo.
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -38,19 +42,26 @@
 //! #### Admin Dispatches
 //!
 //! The administrative dispatches are used to perform admin functions on chain. Currently, the
-//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `DestroyOrigin` and `ResolveOrigin`
+//! admin functions can only be called by the `ApprovalOrigin`, `CloseOrigin`, `ResolveOrigin` and
+//! root.
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
maltekliemann: I'm honestly not quite sure why (in the audit) state _transitions_ are associated with slashes. We need to slash according to _market state_ and _creation_. Depending on the state and the creation type, different bonds are reserved and now need to be slashed.

maltekliemann: See below for an example.

zrml/prediction-markets/src/lib.rs
maltekliemann: The `ValidityBond` is reserved when a permissionless market is created. _Permissionless_ refers to the market creation. If the market moves through its lifecycle as expected, the `ValidityBond` is unreserved when the market is resolved.

If the market is now unexpectedly destroyed mid-usage, the `ValidityBond` must be slashed, but only if it hasn't been unreserved already, i.e. only if the market is not yet resolved.

zrml/prediction-markets/src/lib.rs
maltekliemann: I'm excluding `MarketStatus::Resolved` because the oracle bond is unreserved when the market is resolved, so we shouldn't slash that.

Good catch regarding `InsufficientSubsidy`. The function `process_subsidy_collecting_markets` handles unreserving of bonds, so there's no need to slash them in `destroy_market`. I'll submit a fix and check how this slipped through testing.

zrml/prediction-markets/src/tests.rs
```diff
@@ -1125,6 +1125,163 @@ fn market_resolve_does_not_hold_liquidity_withdraw() {
...
+#[test]
+fn admin_destroy_market_correctly_slashes_permissionless_market_active() {
+    ExtBuilder::default().build().execute_with(|| {
+        simple_create_categorical_market::<Runtime>(
+            MarketCreation::Permissionless,
+            0..1,
+            ScoringRule::CPMM,
+        );
+        let balance_free_before_alice = Balances::free_balance(&ALICE);
+        let balance_free_before_bob = Balances::free_balance(&BOB);
```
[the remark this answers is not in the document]
maltekliemann: Yeah, good catch, we need to use sentinel values for the balances.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -249,9 +249,9 @@ benchmarks! {
             let _ = T::SimpleDisputes::on_dispute(&disputes, &market_id, &market)?;
         }
 
-        let approval_origin = T::ApprovalOrigin::successful_origin();
+        let sudo = RawOrigin::Root.into();
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
-            T::DestroyOrigin::ensure_origin(origin)?;
+            ensure_root(origin)?;
 
             let mut total_accounts = 0usize;
             let mut share_accounts = 0usize;
             let market = T::MarketCommons::market(&market_id)?;
             let market_status = market.status;
             let outcome_assets = Self::outcome_assets(market_id, &market);
             let outcome_assets_amount = outcome_assets.len();
+
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
+            // Slash outstanding bonds
+            if market_status == MarketStatus::Proposed {
+                CurrencyOf::<T>::slash_reserved_named(
+                    &RESERVE_ID,
+                    &market.creator,
+                    T::AdvisoryBond::get(),
+                );
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
+            }
+            if market_status != MarketStatus::Resolved {
+                if market.creation == MarketCreation::Permissionless {
+                    CurrencyOf::<T>::slash_reserved_named(
+                        &RESERVE_ID,
+                        &market.creator,
+                        T::ValidityBond::get(),
+                    );
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
```diff
@@ -135,21 +146,45 @@ mod pallet {
...
+                }
+                CurrencyOf::<T>::slash_reserved_named(
+                    &RESERVE_ID,
+                    &market.creator,
+                    T::OracleBond::get(),
+                );
```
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/tests.rs
```diff
@@ -1125,6 +1125,163 @@ fn market_resolve_does_not_hold_liquidity_withdraw() {
...
+
+#[test]
+fn admin_destroy_market_correctly_slashes_advised_market_active() {
+    ExtBuilder::default().build().execute_with(|| {
+        simple_create_categorical_market::<Runtime>(
+            MarketCreation::Advised,
+            0..1,
+            ScoringRule::CPMM,
+        );
+        assert_ok!(PredictionMarkets::approve_market(Origin::signed(SUDO), 0));
```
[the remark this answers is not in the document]
maltekliemann: We need to sign with a `RawOrigin`. The `ApprovalOrigin` implements `EnsureOrigin`, instead, which is used to verify signatures, and doesn't have a `signed` method.

Chralt98:
What happens to the `OracleBond` when a [market gets rejected](https://github.com/zeitgeistpm/zeitgeist/blob/2ca5d74dcb2aba37554e6ae246791f6b41a009b3/zrml/prediction-markets/src/lib.rs#L948-L966)?

Chralt98:
What happens to the `OracleBond` when a [market gets rejected](https://github.com/zeitgeistpm/zeitgeist/blob/2ca5d74dcb2aba37554e6ae246791f6b41a009b3/zrml/prediction-markets/src/lib.rs#L948-L966)?

maltekliemann:
I've put up a table with how I think unreserving should go: https://github.com/zeitgeistpm/runtime-audit-1/issues/34#issuecomment-1120187097

Also, you suggested that we should verify that all bonds are non-equal to ensure that the tests check that the function doesn't unreserve the incorrect bonds. I agree. My only concern is that this would mean putting a verification of the correctness of the test environment into the tests. This feels wrong. If this were pytest, I might put the verification into a module level fixture. We don't have stuff like this in Rust (or do we?). My best guess right now is to add a comment to the `parameter_types` telling other developers not to mess with these values.

maltekliemann:
> What happens to the `OracleBond` when a [market gets rejected](https://github.com/zeitgeistpm/zeitgeist/blob/2ca5d74dcb2aba37554e6ae246791f6b41a009b3/zrml/prediction-markets/src/lib.rs#L948-L966)?

Hmm, doesn't look like anything happens. I would say the oracle bond gets slashed, as well. I dimly remember fixing that, but maybe I'm wrong. Can you go through the current PRs and check if this has been fixed already? If not, would you mind creating a PR for slashing the oracle, as well (and add the necessary tests)?

zrml/prediction-markets/src/lib.rs
```diff
@@ -139,13 +152,30 @@ mod pallet {
...
+            }
+            if market_status != MarketStatus::Resolved
+                && market_status != MarketStatus::InsufficientSubsidy
+            {
+                if market.creation == MarketCreation::Permissionless {
+                    slash_market_creator(T::ValidityBond::get());
+                }
+                slash_market_creator(T::OracleBond::get());
+            }
+
```
Chralt98, line 171: https://github.com/zeitgeistpm/runtime-audit-1/issues/34#issuecomment-1121268112

maltekliemann:
Sounds good, I'll put that into #580!

zrml/prediction-markets/src/lib.rs
```diff
@@ -139,13 +152,30 @@ mod pallet {
...
+                slash_market_creator(T::AdvisoryBond::get());
+            }
+            if market_status != MarketStatus::Resolved
+                && market_status != MarketStatus::InsufficientSubsidy
+            {
+                if market.creation == MarketCreation::Permissionless {
+                    slash_market_creator(T::ValidityBond::get());
+                }
+                slash_market_creator(T::OracleBond::get());
+            }
```
Chralt98, line 170:
```suggestion
            match market_status {
                // already unreserved or slashed when InsufficientSubsidy or Resolved
                MarketStatus::InsufficientSubsidy | MarketStatus::Resolved => (),
                // reserves left when market is proposed (when proposed always advised then)
                MarketStatus::Proposed if market.creation == MarketCreation::Advised => {
                    slash_market_creator(T::AdvisoryBond::get())
                }
                _ => {
                    // for permissionless and advised markets and
                    // when not InsufficientSubsidy or Resolved then the OracleBond is left
                    slash_market_creator(T::OracleBond::get());
                    // when not InsufficientSubsidy or Resolved for a permissionless market the ValidityBond is left 
                    if market.creation == MarketCreation::Permissionless {
                        slash_market_creator(T::ValidityBond::get());
                    }
                }
            }
```
I agree with your style. The match would be my style. Probably you should comment a little bit in there, because it's hard to understand (when looking at it in two months from now).

Chralt98 (commented):
Maybe comment the slashes.

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
maltekliemann: Not a big fan of this comment style, but I've added a clarifying comment.

zrml/prediction-markets/src/tests.rs
```diff
@@ -1125,6 +1128,350 @@ fn market_resolve_does_not_hold_liquidity_withdraw() {
...
+        let market_id = 0;
+        let pool_id = 0;
+        let pool_account = Swaps::pool_account_id(pool_id);
+        let market_account = PredictionMarkets::market_account(market_id);
+        let alice_balances_before = vec![
+            Currency::free_balance(Asset::CategoricalOutcome(0, 0), &ALICE),
+            Currency::free_balance(Asset::CategoricalOutcome(0, 1), &ALICE),
+            Currency::free_balance(Asset::CategoricalOutcome(0, 2), &ALICE),
+            Currency::free_balance(Asset::Ztg, &ALICE),
+        ];
```
Chralt98, line 1449: Maybe check non-zero free balances for Alice, Pool_Account, Market_Account before?

zrml/prediction-markets/src/tests.rs
```diff
@@ -1125,6 +1128,350 @@ fn market_resolve_does_not_hold_liquidity_withdraw() {
...
+        );
+        assert_eq!(
+            Currency::free_balance(Asset::CategoricalOutcome(0, 1), &ALICE),
+            alice_balances_before[1]
+        );
+        assert_eq!(
+            Currency::free_balance(Asset::CategoricalOutcome(0, 2), &ALICE),
+            alice_balances_before[2]
+        );
+        assert_eq!(Currency::free_balance(Asset::Ztg, &ALICE), alice_balances_before[3]);
```
Chralt98, line 1471: So Alice looses nothing? I thought the assets get slashed :-O ? This checks that Alice free balance of each asset is the same as before. Am I missing something here?

zrml/prediction-markets/src/tests.rs
```diff
@@ -1125,6 +1128,350 @@ fn market_resolve_does_not_hold_liquidity_withdraw() {
...
+        let balance_free_before_alice = Balances::free_balance(&ALICE);
+        assert_ok!(PredictionMarkets::admin_destroy_market(Origin::signed(SUDO), 0));
+        assert_eq!(Balances::reserved_balance_named(&RESERVE_ID, &ALICE), SENTINEL_AMOUNT);
+        let balance_free_after_alice = Balances::free_balance(&ALICE);
+        assert_eq!(balance_free_before_alice, balance_free_after_alice);
+    });
+}
+
+#[test]
+fn admin_destroy_market_correctly_cleans_up_accounts() {
```
Chralt98, line 1426: Maybe create private functions for the `free_balance` checks.

Chralt98 (commented):
This adds functionality for `admin_destroy_market` to remove the pool, slash the market account balance, and deletes the market pool.

zrml/prediction-markets/src/lib.rs
```diff
@@ -139,27 +151,47 @@ mod pallet {
...
-            if let Some(first_asset) = outcome_assets_iter.next() {
-                total_accounts = manage_outcome_asset(first_asset);
+            for asset in outcome_assets.into_iter() {
+                total_accounts = destroy_asset(asset);
             }
-            for asset in outcome_assets_iter {
-                let _ = manage_outcome_asset(asset);
+            CurrencyOf::<T>::slash(&market_account, CurrencyOf::<T>::free_balance(&market_account));
+            if let Ok(pool_id) = T::MarketCommons::market_pool(&market_id) {
+                T::Swaps::destroy_pool(pool_id)?;
```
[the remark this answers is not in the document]
maltekliemann: The `#[transactional]` is added later.

Other than that, the reasoning for checking for errors is the same a the reasoning for not allowing anything that panicks in our code: Maybe someone changes `destroy_pool` in the future so that it can return an error that must be handled by `admin_destroy_market`, but overlooks that `admin_destroy_market` must be changed, as well.

zrml/swaps/src/lib.rs
```diff
@@ -1382,6 +1384,22 @@ mod pallet {
...
+            for asset in pool.assets.into_iter() {
+                let amount = T::Shares::free_balance(asset, &pool_account);
+                T::Shares::slash(asset, &pool_account, amount);
+            }
+            let pool_share_id = Self::pool_shares_id(pool_id);
+            let (_, liquidity_providers) = T::Shares::accounts_by_currency_id(pool_share_id);
+            T::Shares::destroy_all(pool_share_id, liquidity_providers.iter().cloned());
+            Pools::<T>::remove(pool_id);
+            Self::deposit_event(Event::PoolDestroyed(pool_id));
+            Ok(0u64)
```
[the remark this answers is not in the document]
maltekliemann: Yeah, great point!

zrml/swaps/src/tests.rs
```diff
@@ -1508,6 +1508,45 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
     });
 }
 
+#[test]
+fn destroy_pool_fails_if_pool_does_not_exist() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool(ScoringRule::CPMM, true);
```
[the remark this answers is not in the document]
maltekliemann: Yeah, pool ids start at zero, but, yeah, could be more obvious.

zrml/swaps/src/tests.rs
```diff
@@ -1508,6 +1508,45 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
...
+fn destroy_pool_correctly_cleans_up_pool() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool(ScoringRule::CPMM, true);
+        let pool_id = 0;
+        let alice_balance_before = [
+            Currencies::free_balance(ASSET_A, &ALICE),
+            Currencies::free_balance(ASSET_B, &ALICE),
+            Currencies::free_balance(ASSET_C, &ALICE),
+            Currencies::free_balance(ASSET_D, &ALICE),
+        ];
```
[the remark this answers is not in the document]
maltekliemann: I'm now allocating funds using `create_initial_pool_with_funds_for_alice` in order to ensure that Alice has positive funds.

zrml/swaps/src/tests.rs
```diff
@@ -1508,6 +1508,45 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
...
+        assert_ok!(Swaps::destroy_pool(pool_id));
+        assert!(matches!(
+            Swaps::pool_by_id(pool_id).unwrap_err(),
+            crate::Error::<Runtime>::PoolDoesNotExist
+        ));
+        assert_all_parameters(alice_balance_before, 0, [0, 0, 0, 0], 0);
```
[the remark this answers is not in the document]
Chralt98: Sorry about that. I should have read [this](https://github.com/zeitgeistpm/zeitgeist/blob/4ae4d4137a17374a5ce65c04b85cd61e38f0e233/zrml/swaps/src/tests.rs#L1594-L1597).

Chralt98 (approved):
Looks fine to me. Hope we've not forgotten something.

zrml/prediction-markets/src/lib.rs
sea212, line 186: The Rikiddo markets need some additional treatment here:
1. If the market is in `CollectingSubsidy` state, the market has to be removed from the `MarketsCollectingSubsidy` vector, otherwise the pallet will try to process it at a later time
2. In Rikiddo pools the liquidity providers reserve their liquidity in ZTG before the pool becomes active. Once it becomes active, ZTG is converted to `PoolShares`. When a market that is in state `CollectingSubsidy` should be destroyed, those liquidity reserves from the so called "subsidy providers" must be freed (either by unreserving or in that case slashing them). The swaps pallet provides a function called `destroy_pool_in_subsidy_phase`. It should be called instead of `destroy_pool` for markets in `CollectingSubsidy` state. Feel free to add an additional parameter to that function that indicates whether unreserving or slashing should happen.

zrml/swaps/src/lib.rs
```diff
@@ -1384,6 +1386,22 @@ mod pallet {
...
+            let pool = Self::pool_by_id(pool_id)?;
+            let pool_account = Self::pool_account_id(pool_id);
+            for asset in pool.assets.into_iter() {
+                let amount = T::Shares::free_balance(asset, &pool_account);
+                T::Shares::slash(asset, &pool_account, amount);
+            }
+            let pool_share_id = Self::pool_shares_id(pool_id);
+            let (_, liquidity_providers) = T::Shares::accounts_by_currency_id(pool_share_id);
+            T::Shares::destroy_all(pool_share_id, liquidity_providers.iter().cloned());
+            Pools::<T>::remove(pool_id);
```
sea212, line 1399: For Rikiddo pools the Rikiddo pallet must be informed that the Rikiddo data for that pool should be cleaned up: `T::RikiddoSigmoidFeeMarketEma::destroy(pool_id)?`

zrml/swaps/src/tests.rs
```diff
@@ -46,6 +46,46 @@ const _101: u128 = 101 * BASE;
...
+            Currencies::free_balance(ASSET_A, &ALICE),
+            Currencies::free_balance(ASSET_B, &ALICE),
+            Currencies::free_balance(ASSET_C, &ALICE),
+            Currencies::free_balance(ASSET_D, &ALICE),
+        ];
+        assert_ok!(Swaps::destroy_pool(pool_id));
+        assert!(matches!(
+            Swaps::pool_by_id(pool_id),
+            Err(crate::Error::<Runtime>::PoolDoesNotExist)
+        ));
```
sea212, line 72: We should use [`assert_err!`](https://paritytech.github.io/substrate/master/frame_support/macro.assert_err.html) for that.

zrml/swaps/src/lib.rs
```diff
@@ -1384,6 +1386,22 @@ mod pallet {
...
+            Self::deposit_event(Event::PoolDestroyed(pool_id));
+            // TODO(#603): Fix weight calculation.
+            Ok(0u64)
```
sea212, line 1402: We should not use 0, better use a decent part of the available block weight until we properly calculate it.

zrml/prediction-markets/src/tests.rs
```diff
@@ -64,6 +64,338 @@ fn simple_create_scalar_market<T: crate::Config>(
...
+        );
+        assert_ok!(Currency::deposit(Asset::Ztg, &ALICE, 2 * SENTINEL_AMOUNT));
+        assert_ok!(Balances::reserve_named(&RESERVE_ID, &ALICE, SENTINEL_AMOUNT));
+        let balance_free_before_alice = Balances::free_balance(&ALICE);
+        assert_ok!(PredictionMarkets::admin_destroy_market(Origin::signed(SUDO), 0));
+        assert_eq!(Balances::reserved_balance_named(&RESERVE_ID, &ALICE), SENTINEL_AMOUNT);
+        let balance_free_after_alice = Balances::free_balance(&ALICE);
+        assert_eq!(balance_free_before_alice, balance_free_after_alice);
+    });
+}
```
sea212, line 122: There is a lot of duplicate code, also in some of the tests below. I think we can move a lot of this code in a separate helper function.

sea212 (changes requested):

zrml/swaps/src/lib.rs
maltekliemann: My plan was to exclude Rikiddo from all this stuff for now (since it's disabled anyways). Is that OK? I could add a TODO marker that mentions that these functions are only to be used with `CPMM` markets.

zrml/prediction-markets/src/lib.rs
maltekliemann: (Same as below, I'd really like to skip Rikiddo for now)

zrml/swaps/src/lib.rs
sea212: This is fine in my opinion. But in case we go that path, please:
1. Enforce in `admin_destroy_market` that the `ScoringRule` is `CPMM`. The function should ideally just throw an error that indicates that the operation is not supported for `RikiddoSigmoidFeeMarketEma` markets.
2. Create a new issue to implement the support and reference both my review comments in it (because they state what changes are necessary).

In regards to this review, besides the Rikiddo compatibility the only change that I would insist on is [changing the zero weight to something significant](https://github.com/zeitgeistpm/zeitgeist/pull/586#discussion_r871332474) like double-digit weight per block.

zrml/swaps/src/lib.rs
maltekliemann: Okay, I went with 10% of the block. But it doesn't really matter, actually, since we don't use the result of `destroy_pool` in `admin_destroy_market`. I would have to correct the benchmarks for that.

But I don't think it matters, because `admin_destroy_market` currently uses 56% of available block time (worst-case, which assumes that there are 900 accounts that have to be slashed).

zrml/prediction-markets/src/lib.rs
```diff
@@ -36,19 +39,26 @@
...
+//!   to resolved. Can only be called by `ResolveOrigin`.
 //!
-//! #### `ApprovalOrigin`, `CloseOrigin`, `DestroyOrigin` and `ResolveOrigin` Dispatches
+//! #### `ApprovalOrigin` Dispatches
 //!
-//! Those origins are mainly minimum vote proportions from the advisory committee, the on-chain governing body of Zeitgeist
-//! that is responsible for maintaining a list of high quality markets and slash low quality markets.
+//! Those origins are mainly minimum vote proportions from the advisory committee, the on-chain
+//! governing body of Zeitgeist that is responsible for maintaining a list of high quality markets
+//! and slash low quality markets.
```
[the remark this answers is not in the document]
maltekliemann: Yeah, sounds good. Did some rewriting of the docs.

zrml/swaps/src/lib.rs
maltekliemann: Ok, done!

sea212 (approved):
Thanks!
The `admin_destroy_market` function must be `transactional` since it can fail after state changes were applied, but afaik this is already handled by another PR.

maltekliemann:
Yeah, that #576 takes care of that.

zrml/swaps/src/tests.rs
maltekliemann: I'm not sure what's going on, but I get 

```
note: an implementation of `PartialEq` might be missing for `pallet::Error<mock::Runtime>`
```

I'm surprised this is an issue. I could add `#[derive(PartialEq)]` to the `Error` enum, but I have a feeling that something else is off.

maltekliemann: Oh, ok, `DispatchError` implements `PartialEq` by comparing module and error index; but `pool_by_id` returns `Error<T>`.

sea212 (approved):

PR #590, opened 2022-04-25
Make all in/out limits optional

### LINKED ISSUE #589: [swaps] Make all limits optional
By _limits_ I mean things like `max_price`, `min_asset_amount`, etc. These should probably be made optional for the following reason:

- We want to specify slippage limits in the frontend in terms of the total price (i.e. slippage of 5% means that the total price slips only by 5%). It seems rather difficult to calculate how to calculate a reasonable limit for the per-unit price.
- During testing, we very often don't want to bother with entering these values, especially when using polkadot-js.

maltekliemann:
Fixes #589.


maltekliemann:
To explain this a bit more: Many `max_price` and `max_asset_in` (etc.) specifications are just sufficiently large number because we don't care if we get rekt by slippage. I think making these values `Option` correctly expresses that sometimes, people just want to buy at market price.

maltekliemann:
Remaining tasks:

- [ ] Update CHANGELOG

zrml/swaps/src/tests.rs
```diff
@@ -1365,6 +1381,39 @@ fn swap_exact_amount_in_exchanges_correct_values_with_cpmm() {
...
+}
+
+#[test]
+fn swap_exact_amount_in_fails_if_max_price_is_not_satisfied_with_cpmm() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        // We're swapping 1:1, but due to slippage the price will exceed _1, so this should raise an
+        // error:
+        assert_noop!(
+            Swaps::swap_exact_amount_in(alice_signed(), 0, ASSET_A, _1, ASSET_B, None, Some(_1)),
```
Chralt98, line 1411: I don't really get this. Could you please explain this simply like I am a five year old?

zrml/swaps/src/tests.rs
```diff
@@ -1434,6 +1507,39 @@ fn swap_exact_amount_out_exchanges_correct_values_with_cpmm() {
...
+                ASSET_A,
+                Some(expected_amount - 100),
+                ASSET_B,
+                _1,
+                None,
+            ),
+            crate::Error::<Runtime>::LimitIn,
+        );
+    });
+}
```
Chralt98, line 1528: Same as mentioned above.

zrml/swaps/src/tests.rs
```diff
@@ -1655,6 +1759,107 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
...
+fn pool_join_with_exact_pool_amount_fails_if_max_asset_amount_is_violated() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        assert_noop!(
+            Swaps::pool_join_with_exact_pool_amount(
+                alice_signed(),
+                0,
+                ASSET_A,
+                _1,
+                40604010000 - 100,
```
Chralt98, line 1800: (above)

zrml/swaps/src/tests.rs
```diff
@@ -1655,6 +1759,107 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
...
+    });
+}
+
+#[test]
+fn pool_exit_fails_if_min_assets_out_is_violated() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        assert_ok!(Swaps::pool_join(alice_signed(), 0, _1, vec!(_1, _1, _1, _1)));
+        assert_noop!(
+            Swaps::pool_exit(alice_signed(), 0, _1, vec!(_1, _1, _1 + 1, _1)),
```
Chralt98, line 1813: mentioned above (could be any asset bound)

zrml/swaps/src/tests.rs
```diff
@@ -1541,24 +1663,6 @@ fn create_pool_fails_if_base_asset_is_not_in_asset_vector() {
...
-                    *right_val - *left_val
-                };
-                if diff > $precision {
-                    panic!("{} is not {}-close to {}", *left_val, *precision_val, *right_val);
-                }
-            }
-        }
-    };
-}
-
```
Chralt98, line 1561: Because of this reminder, which says, that it's not used?

zrml/swaps/src/tests.rs
```diff
@@ -1726,6 +1931,25 @@ fn subsidize_and_start_rikiddo_pool(
...
+                    *right_val - *left_val
+                };
+                if diff > $precision {
+                    panic!("{} is not {}-close to {}", *left_val, *precision_val, *right_val);
+                }
+            }
+        }
+    };
+}
+
```
Chralt98, line 1952: Oh, nevermind of this comment above.

Chralt98 (commented):
Looks fine and saves time for the polkadot-js users.

zrml/swaps/src/benchmarks.rs
```diff
@@ -365,21 +365,21 @@ benchmarks! {
...
             Some(BASE.saturated_into()), ScoringRule::RikiddoSigmoidFeeMarketEma, true);
-        let max_amount_asset_in: BalanceOf<T> = (BASE * 1024).saturated_into();
+        let max_asset_amount_in: Option<BalanceOf<T>> = Some((BASE * 1024).saturated_into());
         let asset_amount_out: BalanceOf<T> = BASE.saturated_into();
-        let max_price = (BASE * 1024).saturated_into();
-    }: swap_exact_amount_out(RawOrigin::Signed(caller), pool_id, *assets.last().unwrap(), max_amount_asset_in,
+        let max_price = Some((BASE * 1024).saturated_into());
+    }: swap_exact_amount_out(RawOrigin::Signed(caller), pool_id, *assets.last().unwrap(), max_asset_amount_in,
             assets[0], asset_amount_out, max_price)
 }
```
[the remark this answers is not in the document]
maltekliemann: We could run benchmarks on a per-PR basis. In this case, the benchmarks shouldn't change significantly because we haven't really changed anything. There also seems to be a problem with running benchmarks from a fork, so I think we can skip them for now (and I'm going to stop using a fork for making PRs).

zrml/swaps/src/lib.rs
```diff
@@ -1869,7 +1869,9 @@ mod pallet {
                             T::RikiddoSigmoidFeeMarketEma::cost(pool_id, &outstanding_after)?;
                         cost_before.checked_sub(&cost_after).ok_or(ArithmeticError::Overflow)?
                     };
-                    ensure!(asset_amount_out >= min_asset_amount_out, Error::<T>::LimitOut);
+                    if let Some(maao) = min_asset_amount_out {
```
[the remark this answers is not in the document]
maltekliemann: :wink: Usually, I'd just replace `min_asset_amount_out` with `maybe_min_asset_amount_out` and then do `if let Some(min_asset_amount_out) = maybe_min_asset_amount_out`, but since these parameter names show up in polkadot-js, that doesn't sound like a good idea.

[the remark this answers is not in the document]
Chralt98: I understand. It wasn't really meant like a change proposal anyways 😄.

zrml/swaps/src/tests.rs
```diff
@@ -1365,6 +1381,39 @@ fn swap_exact_amount_in_exchanges_correct_values_with_cpmm() {
...
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        let expected_amount = 9900990100;
+        assert_noop!(
+            Swaps::swap_exact_amount_in(
+                alice_signed(),
+                0,
+                ASSET_A,
+                _1,
+                ASSET_B,
+                Some(expected_amount + 100),
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in 94ed576d8fb6ecf38bd50fd48d206b211bb39be9

zrml/swaps/src/tests.rs
maltekliemann: Mitigated in 94ed576d8fb6ecf38bd50fd48d206b211bb39be9

zrml/swaps/src/tests.rs
```diff
@@ -1655,6 +1759,107 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
     });
 }
 
+#[test]
+fn pool_join_fails_if_max_assets_in_is_violated() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        assert_noop!(
+            Swaps::pool_join(alice_signed(), 0, _1, vec!(_1, _1, _1 - 1, _1)),
```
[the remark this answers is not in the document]
maltekliemann: Yeah, just trying to be "random". There's always the option of going full hog and testing every index. Seems to be a situation of diminishing returns to me.

zrml/swaps/src/tests.rs
```diff
@@ -1655,6 +1759,107 @@ fn create_pool_fails_on_total_weight_above_maximum_total_weight() {
...
+fn pool_join_with_exact_asset_amount_fails_if_min_pool_tokens_is_violated() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        assert_noop!(
+            Swaps::pool_join_with_exact_asset_amount(
+                alice_signed(),
+                0,
+                ASSET_A,
+                _1,
+                2490679300 + 100,
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in 86292ef00b241adc623ea0e991592aba25909c64

zrml/swaps/src/tests.rs
maltekliemann: Mitigated in 86292ef00b241adc623ea0e991592aba25909c64

zrml/swaps/src/tests.rs
maltekliemann: I think that adding `1` is a good indicator that we're just going "a bit too far". Think of it this way: We just joined the pool 1:1 and now to to get a little more out than we put in. This should fail.

maltekliemann: In fact, `1` is not an arbitrary number in this case.

zrml/swaps/src/tests.rs
```diff
@@ -1365,6 +1381,39 @@ fn swap_exact_amount_in_exchanges_correct_values_with_cpmm() {
     });
 }
 
+#[test]
+fn swap_exact_amount_in_fails_if_min_asset_amount_out_is_not_satisfied_with_cpmm() {
+    ExtBuilder::default().build().execute_with(|| {
+        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, true);
+        let expected_amount = 9900990100;
```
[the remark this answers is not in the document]
maltekliemann: Oh, regarding you suggestion of using the math functions: Yeah, that's an option too. Generally, I try to avoid this. There's a subtle difference between the two tests: In one case, you're testing that the function works with a particular, separately verified number. When using your suggestion, you're testing that the function works with whatever `calc_out_given_in` produces. The latter carries the risk that when we break `calc_out_given_in`, the test might not raise an error.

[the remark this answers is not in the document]
maltekliemann: "My" version, on the other hand, is super annoying, because we manually have to replace the concrete number when `calc_out_given_in` changes. :wink: However, I don't think we will ever change the price functions.

zrml/swaps/src/tests.rs
```diff
@@ -146,7 +146,15 @@ fn allows_the_full_user_lifecycle() {
...
-        assert_ok!(Swaps::swap_exact_amount_out(alice_signed(), 0, ASSET_A, _2, ASSET_B, _1, _3,));
+        assert_ok!(Swaps::swap_exact_amount_out(
+            alice_signed(),
+            0,
+            ASSET_A,
+            Some(_2),
+            ASSET_B,
+            _1,
+            None
+        ));
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in a99afbdc82fb607e835d32964c0ae0b4afdb36b3

You raise a really good point here. The changes seem strange. I've restored most of the old tests to their original form (just with all values wrapped in `Some`). The tests that verify correctness of values now pass twice, once with limits, once without.

Chralt98 (approved):
Looks good to me. Apart from change log

PR #592, opened 2022-04-29
Fix documentation of `MarketDestroyed` event

maltekliemann:


zrml/prediction-markets/src/lib.rs
```diff
@@ -1267,7 +1267,7 @@ mod pallet {
         MarketApproved(MarketIdOf<T>, MarketStatus),
         /// A market has been created \[market_id, creator\]
         MarketCreated(MarketIdOf<T>, Market<T::AccountId, T::BlockNumber, MomentOf<T>>),
-        /// A market has been created \[market_id, creator\]
+        /// A market has been destroyed \[market_id, creator\]
```
Chralt98, line 1270:
```suggestion
        /// A market has been destroyed. \[market_id\]
```

Chralt98 (commented):
You had one job. :-D just kidding. Maybe we should write for every event a period after each sentence consequently. That's okay to me then.

zrml/prediction-markets/src/lib.rs
maltekliemann: :facepalm:

Chralt98 (approved):
Looks good to me 👍

PR #593, opened 2022-05-03
Change clippy usage and make code cleaner

maltekliemann:
Fixes zeitgeistpm/runtime-audit-1#6.

clippy exceptions are now moved into the files in which they are required. Many of the warning that we previously filtered are now errors. I've fixed wherever they occur.

I've kept the shell script for now, because I don't want to break up our GitHub Actions workflows.


sea212 (approved):
Much better now :+1:

vivekvpandya (approved):

PR #594, opened 2022-05-04
Make `create_*_market` emit the `market_account`, `create_pool` emit the `pool_account`

### LINKED ISSUE #588: [Squid] Market/Pool account address as event param
Subsquid looks at `system.NewAccount` event to tag account as market or pool. This provides perspective on the balances of such accounts. In cases where more than one `NewAccountEvent` is present inside a block, processor will require more info while mapping them.

Therefore, it would be great to have account address as event param in `predictionMarkets.MarketCreated` and `swaps.PoolCreate`.

Not sure when the market account is really created, but looking at the events `system.NewAccount` always comes along with the event `predictionMarkets.BoughtCompleteSet`.

maltekliemann:
Fixes #588.

### DIFF
--- a/docs/changelog_for_devs.md
+++ b/docs/changelog_for_devs.md
@@ -1,3 +1,8 @@
+# v0.3.2
+
+- Added a field for `market_account` to `MarketCreated` event and `pool_account`
+  field to `PoolCreate` event.
+
 # v0.3.1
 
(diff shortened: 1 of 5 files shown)


Chralt98 (approved):
This adds the AccountId to the creation events. Looks fine to me.

Chralt98 (approved):

PR #600, opened 2022-05-06
Fix tokenomics, set initial liquidity mining to zero

maltekliemann:
Fixes zeitgeistpm/runtime-audit-1#29. Changes made according to internal communication. The inital funds for liquidity mining have been set to zero.

### DIFF
--- a/primitives/src/constants/ztg.rs
+++ b/primitives/src/constants/ztg.rs
@@ -2,35 +2,32 @@ use sp_runtime::Perbill;
 
 // Distribution
 
-/// Total ZTG amount for airdrops
-pub const AIRDROPS: u128 = 3_000_000;
-
-/// Total ZTG amount for collators
-pub const COLLATORS: u128 = 10_000_000;
-
-/// Total ZTG amount for community incentives
+/// Total ZTG amount for community incentives.
 pub const COMMUNITY_INCENTIVES: u128 = 2_000_000;
 
-/// Total ZTG amount for liquidity mining
-pub const LIQUIDITY_MINING: u128 = 10_000_000;
+/// Total ZTG amount for collators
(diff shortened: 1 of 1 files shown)


primitives/src/constants/ztg.rs
```diff
@@ -2,35 +2,32 @@ use sp_runtime::Perbill;
...
 /// Total ZTG amount for collators
-pub const COLLATORS: u128 = 10_000_000;
-
-/// Total ZTG amount for community incentives
-pub const COMMUNITY_INCENTIVES: u128 = 2_000_000;
+pub const COLLATORS: u128 = 0;
 
 /// Total ZTG amount for liquidity mining
-pub const LIQUIDITY_MINING: u128 = 10_000_000;
+pub const LIQUIDITY_MINING: u128 = 0;
```
sea212, line 12: I am not sure if it should be zero. Yes, during genesis on mainnet they were set to zero (this is the case for everything), nevertheless afaik we have reserved a certain amount of the complete initial ZTG pool for liquidity mining which should be reflected here.
Please verify @lsaether

sea212 (commented):
Thanks for updating those. I'll approve once the LM question has been resolved.

primitives/src/constants/ztg.rs
sea212: I guess we can stick with that, since that is the current state I see on the chain.

lsaether: Currently we've reserved the remaining token supply (approx 50M ZTG) in the Zeitgeist Foundation wallets, which will employ these tokens in a variety of uses including liquidity mining. However, we need to still determine the exact allocations of this pot of funds.

lsaether (approved):

PR #604, opened 2022-05-10
OracleBond unreserve in reject_market

Chralt98:
I am thinking about if we should either slash or unreserve the OracleBond when a market gets rejected (`reject_market` function). At the moment I decided for an unreserve.

### DIFF
--- a/zrml/prediction-markets/src/tests.rs
+++ b/zrml/prediction-markets/src/tests.rs
@@ -20,9 +20,6 @@ use zeitgeist_primitives::{
     },
 };
 use zrml_market_commons::MarketCommonsPalletApi;
-use zrml_prediction_markets::RESERVE_ID;
-
-const SENTINEL_AMOUNT: u128 = BASE;
 
 const SENTINEL_AMOUNT: u128 = BASE;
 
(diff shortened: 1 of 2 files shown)


zrml/prediction-markets/src/tests.rs
```diff
@@ -202,6 +205,72 @@ fn it_allows_the_advisory_origin_to_reject_markets() {
...
+        assert_ok!(PredictionMarkets::reject_market(Origin::signed(SUDO), 0));
+
+        assert_noop!(
+            MarketCommons::market(&0),
+            zrml_market_commons::Error::<Runtime>::MarketDoesNotExist
+        );
+
+        // AdvisoryBond gets slashed after reject_market
+        // OracleBond gets unreserved after reject_market
+        // need to non-equal to see the difference between AdvisoryBond and OracleBond
```
maltekliemann, line 258: I'd remove this line.

```suggestion
```

maltekliemann (changes requested):
Looks good altogether, just a bunch of small corrections. :thumbsup:

The PR adds functionality to `reject_market` that unreserves the oracle bond und slashes the advisory bond, and it adds a test that verifies that this behavior is achieved.

maltekliemann (approved):
Looks perfectly fine! 👍

PR #614, opened 2022-05-11
Set maximum for the market period

### LINKED ISSUE #480: [primitives] `MarketPeriod` musn't end before it starts
Currently, it's possible to create a market with `period.start > period.end`, like this, for example:

```javascript
let tx = api.tx.predictionMarkets.createCpmmMarketAndDeployAssets(
  ALICE,
  api.createType("MarketPeriod", { Block: [100000, 1000] }),
  api.createType("MultiHash", { Sha3_384: METADATA }),
  api.createType("MarketType", { Categorical: 2 }),
  api.createType("MarketDisputeMechanism", { Authorized: ALICE }),
  100 * BASE,
  [100 * BASE, 100 * BASE],
  [5 * MIN_WEIGHT, 5 * MIN_WEIGHT, 10 * MIN_WEIGHT],
  [0, 0]
);
```

This should probably not happen, although the only negative consequence seems to be that the market is unusable.

maltekliemann:
This fixes zeitgeistpm/runtime-audit-2#14 and fixes #480.

The idea is as follows: We implement a `MaxMarketPeriod` constant, which `period.start` and `period.end` may not exceed. And while we're at it, we're also requiring that `period.end > period.start`.

The `MaxMarketPeriod` constant is set to `u64::max() / 2` (recall that `BlockNumber` and `Moment` are both `u64`). Just to give you an idea of the scope: When using `BlockNumber`, this gives us `3509654504130` years, and when using  `Moment` we get `292471208` years (divide by `12000ms`).

It takes a little bit of explaining why this fixes zeitgeistpm/runtime-audit-2#14. The problem is that `report` adds the report period to the market end time, potentially causing an overflow. This is, of course, still possible now, but only if you configure the report period to be larger than `u64::max() / 2 - 1`. Long-term, we may want to add integration tests that do a sanity check like "`MaxMarketPeriod + ReportPeriod + MaxDisputes * DisputePeriod` doesn't overflow".

Anyways, I don't see how I could write a test that checks that the behavior exposed in the audit is no longer possible, because my solution is to prevent the chain from ever reaching that invalid state.

By the way, the type of `MaxMarketPeriod` is `Moment`. This isn't exactly ideal, because `MaxMarketPeriod` is also applied to the `Block` variant of `MarketPeriod`, which has type `BlockNumber`. I've opened an issue a while ago suggesting that we should probably get rid of one of the `MarketPeriod` variants: #616

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10766,6 +10766,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "test-case"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6344589a99d3971d6fa4e8314dbcbeca2df6273a6b642e46906971779159af1f"
+dependencies = [
+ "cfg-if 1.0.0",
+ "proc-macro-error",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "test-runtime-constants"
 version = "0.9.17"
 source = "git+https://github.com/purestake/polkadot?branch=moonbeam-polkadot-v0.9.17#66cb46bb3a7bd7295d1eb39a391559bc6efb2a22"
@@ -12396,6 +12409,7 @@ dependencies = [
  "sp-io",
  "sp-runtime",
  "substrate-fixed 0.5.8",
+ "test-case",
  "zeitgeist-primitives",
  "zrml-authorized",
  "zrml-court",
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -74,8 +74,7 @@ parameter_types! {
     pub const MinSubsidyPeriod: Moment = 60_000;
     // 2_678_400_000 = 31 days.
     pub const MaxSubsidyPeriod: Moment = 2_678_400_000;
-    // TODO Exact requirements: MaxPeriod + ReportingPeriod + MaxDisputes * DisputePeriod <
-    // u64::MAX.
+    // Requirements: MaxPeriod + ReportingPeriod + MaxDisputes * DisputePeriod < u64::MAX.
     pub const MaxMarketPeriod: Moment = u64::MAX / 2;
     pub const OracleBond: Balance = 50 * CENT;
     pub const PmPalletId: PalletId = PalletId(*b"zge/pred");
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -74,6 +74,9 @@ parameter_types! {
     pub const MinSubsidyPeriod: Moment = 60_000;
     // 2_678_400_000 = 31 days.
     pub const MaxSubsidyPeriod: Moment = 2_678_400_000;
+    // TODO Exact requirements: MaxPeriod + ReportingPeriod + MaxDisputes * DisputePeriod <
+    // u64::MAX.
+    pub const MaxMarketPeriod: Moment = u64::MAX / 2;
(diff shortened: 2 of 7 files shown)


sea212:
I completely agree with the approach on `period.end > period.start`.
I wonder why we have to configure a `MaxMarketPeriod` to avoid overflows. Is there a reason why just using only saturated math won't do the job?

Nevertheless I think we should think about a `MaxMarketPeriod` anyways, since at some point it should be encouraged to create multiple markets. For example, I think it does not make much sense to create a market about "Will humanity reach the climate goal XY until 2030?" that's open until 2030. In that case, multiple markets that crystallize information of the intermediate steps to that goal would be way more valuable and the right approach from my perspective. Consequently, I tend towards introducing a `MaxMarketPeriod` that also ensures that we'll have practically never ending markets. That's just my opinion though, some additional input would be great.

maltekliemann:
> I wonder why we have to configure a MaxMarketPeriod to avoid overflows. Is there a reason why just using only saturated math won't do the job?

My thinking was that a market with `period.end = u64::max()` would be un-reportable due to the overflow. Given that this could only happen in a couple of million years, this is kind of a technicality. I could just add saturated math, if you like, just to prevent overflow for aesthetic reasons?

Your suggestion would mean that we require a maximum duration for each market, like one year, or something. We could da that, as well.

sea212:
> Your suggestion would mean that we require a maximum duration for each market, like one year, or something. We could da that, as well.

Yes exactly, I was thinking about a duration. I've created an issue https://github.com/zeitgeistpm/zeitgeist/issues/624 that we can discuss some time in the future.

> My thinking was that a market with `period.end = u64::max()` would be un-reportable due to the overflow. Given that this could only happen in a couple of million years, this is kind of a technicality. I could just add saturated math, if you like, just to prevent overflow for aesthetic reasons?

I am not quite sure where my error in thinking is, but with saturated math a market with `period.end = u64::MAX` should be reportable, since
```rust
if current_block <= range.end.saturating_add(T::ReportingPeriod::get().into()) {
    should_check_origin = true;
}
```

In case of `range.end = u64::MAX` would evaluate to:

```rust
if u64::MAX <= u64::MAX {
    should_check_origin = true;
}
```

I am also confused right now that the code contains some parts without proper overflow handly. Our pipeline should fail at the Clippy stage, since it does not allow integer arithmetic: `-Dclippy::integer_arithmetic`
We should always use an adequate function for overflow handling.

maltekliemann:
Yeah, you're right, it's reportable. Anyways, at that point we run out of room to store our timestamp/block number anyways. I guess that's what I was thinking.

I was scratching my head about the integer arithmetic as well, but forgot about that in the heat of battle. I'll replace it with saturating math and open an issue to investigate clippy.

sea212:
Since all operations are saturated now, do we even need `MaxMarketPeriod`? Does it also ensure that the market can run through every state in-time?

maltekliemann:
I'd say that any market which ends on a reasonable time scale can run through every state in-time, even _with_ the overflowing arithmetic. Thinking about this in detail, I don't really understand the issue anymore.

The scenario that `MaxMarketPeriod` prevents is the following: If a market ends on `u64::MAX` or a little earlier, you may still be able to report it, but time will run out (i.e. no more room in the `u64`) before the dispute period ends. But this concerns markets that will end in a couple of million years.

So I would say that `MaxMarketPeriod` wasn't ever needed.

sea212 (approved):

Issue #429, opened 2022-01-13
[Tests] Implement try-runtime


sea212:
Try-Runtime is a feature that can be used to sync the state of a live blockchain and execute tests on that specific state.

https://docs.substrate.io/v3/tools/try-runtime/

PR #615, opened 2022-05-12
Integration of try-runtime

Chralt98:
Related to #429.

I started the [How-To-Guide on the substrate page](https://docs.substrate.io/how-to-guides/v3/tools/try-runtime/), but it's not really easy, because the guide is a little bit imprecise.

[This](https://github.com/paritytech/substrate/pull/10909/files) and [this](https://github.com/paritytech/substrate/blob/master/utils/frame/try-runtime/cli/src/lib.rs) might be helpful too.


Chralt98:
[This](https://github.com/paritytech/substrate/pull/10909/files) and [this](https://github.com/paritytech/substrate/blob/master/utils/frame/try-runtime/cli/src/lib.rs) might be helpful too.

runtime/src/lib.rs
```diff
@@ -1339,6 +1339,25 @@ impl_runtime_apis! {
...
+    impl frame_try_runtime::TryRuntime<Block> for Runtime {
+        fn on_runtime_upgrade() -> (frame_support::weights::Weight, frame_support::weights::Weight) {
+            log::info!("try-runtime::on_runtime_upgrade.");
+            // NOTE: intentional unwrap: we don't want to propagate the error backwards, and want to
+            // have a backtrace here. If any of the pre/post migration checks fail, we shall stop
+            // right here and right now.
+            let weight = Executive::try_runtime_upgrade().map_err(|err|{
+                log::info!("try-runtime::on_runtime_upgrade failed with: {:?}", err);
+                err
+            }).unwrap();
```
maltekliemann, line 1353:
```suggestion
            let weight = Executive::try_runtime_upgrade().unwrap();
```

This one confused me quite a bit. While `Executive::try_runtime_upgrade()` does return a `Result<Weight, 'static &str>`, it _cannot_ return an `Err`. See https://paritytech.github.io/substrate/master/src/frame_executive/lib.rs.html#265-272. If pre/post checks fail, it panics. Otherwise, it returns `Ok(weight)`. So we never get to see the message `"try-runtime::on_runtime_upgrade failed with: {:?}"`.

If you want to keep the message (insurance against future changes of `try_runtime_upgrade`), you can also do this, which should be funcionally equivalent to your suggestion:

```suggestion
            let weight = Executive::try_runtime_upgrade().expect(
                "try-runtime::on_runtime_upgrade failed with"
            );
```

runtime/Cargo.toml
```diff
@@ -271,6 +278,62 @@ std = [
 txfilter = [
     "log"
 ]
+try-runtime = [
+    "frame-executive/try-runtime",
+    "frame-try-runtime",
+    "frame-support/try-runtime",
+
+    # For every pallet in the runtime include try-runtime
```
maltekliemann, line 286: You've added a `try-runtime` feature to all Zeitgeist pallets. I can see that Parity does the same thing for FRAME pallets. Why?

It doesn't seem like any pallet uses the feature. Aren't we basically adding an unused dependency here?

runtime/Cargo.toml
```diff
@@ -271,6 +278,62 @@ std = [
...
+    "frame-system/try-runtime",
+    "pallet-timestamp/try-runtime",
+    "pallet-randomness-collective-flip/try-runtime",
+    "pallet-scheduler/try-runtime",
+    "pallet-preimage/try-runtime",
+
+    # Money runtime pallets
+    "pallet-balances/try-runtime",
+    "pallet-transaction-payment/try-runtime",
+    # "pallet-transaction-payment-rpc-runtime-api/try-runtime",
```
maltekliemann, line 298: [nitpick] I'd remove these commons.

maltekliemann (changes requested):
I tested this on Battery Station. Works like a charm! :thumbsup:

But we should clarify the dependencies before we proceed, I think. I checked with a couple of other projects, and they all equip each pallet with a (seemingly useless) `try-runtime` feature.

Just another quick point (apparently you can't make GitHub suggestions in unedited files): Maybe we can add something like
  ```
  .PHONY: try-runtime-upgrade
  try-runtime-upgrade:
	cargo run --bin zeitgeist --features parachain --features try-runtime try-runtime on-runtime-upgrade live --uri wss://bsr.zeitgeist.pm:443
```

to the Makefile?

File: runtime/src/lib.rs (hunk 1 above)
Comment (maltekliemann) [reply runtime/src/lib.rs:1353]: That's true. My second suggestion was merely meant as a functionally equivalent alternative to what you already have (in case `try_runtime_upgrade` ever actually returns an error).

File: runtime/Cargo.toml (hunk 2 above)
Comment (Chralt98) [reply runtime/Cargo.toml:286]: At least it produces no error, when removing the `try-runtime` of all the Zeitgeist pallets. I don't really know why Parity includes this in every pallet although it's only used in a few ones. If you worry, that it would be unnecessary overhead, I see no reason to not temporarily leave that out. The two reasons to let it stay in each pallet is, that a storage migration / runtime upgrade depends on one storage item of a pallet OR the time we use it, we wouldn't have to think about including it in the runtime and `Cargo.toml` file of the pallet. 

But wouldn't cargo ignore the features anyways, when it's not activated (without `--features try-runtime`)?

[This is the PR](https://github.com/paritytech/substrate/pull/8194) where the pallets got add.

Apart from that: It think there is no dependency (overhead) added, because it comes from `frame-support` which is needed anyways.
```text
try-runtime = [
  "frame-support/try-runtime",
]
```

maltekliemann: Ok, I understand now. Of course we need the `try-runtime` feature if we want to use `try-runtime` in pallet-specific migrations. Ok, leave them in. :thumbsup:

Makefile
```diff
@@ -13,6 +13,10 @@ check:
 check-dummy:
 	BUILD_DUMMY_WASM_BINARY= cargo check
 
+.PHONY: try-runtime-upgrade
+try-runtime-upgrade:
+	cargo run --bin zeitgeist --features parachain --features=try-runtime try-runtime on-runtime-upgrade live --uri wss://bsr.zeitgeist.pm:443
+
```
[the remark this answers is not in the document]
Chralt98: What was the command you executed? Was it: `RUST_BACKTRACE=1 cargo run --bin zeitgeist --features parachain --features=try-runtime try-runtime on-runtime-upgrade live --uri wss://bsr.zeitgeist.pm:443`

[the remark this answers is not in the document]
maltekliemann: Yes

primitives/Cargo.toml
```diff
@@ -27,6 +27,9 @@ std = [
     "sp-core/std",
     "sp-runtime/std",
 ]
+try-runtime = [
+  "frame-support/try-runtime",
+]
```
sea212, line 32: This crate (primitives) is not a pallet and it won't use the functionality that try-runtime provides. Consequently we don't need to compile `frame-support` with the `try-runtime` feature.

runtime/Cargo.toml
```diff
@@ -271,6 +278,55 @@ std = [
...
+    "pallet-identity/try-runtime",
+    "pallet-sudo/try-runtime",
+    "pallet-utility/try-runtime",
+    
+    # ORML runtime pallets
+    "orml-currencies/try-runtime",
+    "orml-tokens/try-runtime",
+
+    # Zeitgeist runtime pallets
+    "zeitgeist-primitives/try-runtime",
```
sea212, line 317:
```suggestion
```

We don't need that feature on primitives (see comment in primitives crate)

node/Cargo.toml
```diff
@@ -36,6 +36,10 @@ sp-timestamp = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com
 sp-transaction-pool = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
 substrate-frame-rpc-system = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
 
+# Try-Runtime
+
+try-runtime-cli = { branch = "moonbeam-polkadot-v0.9.19", optional = true, git = "https://github.com/purestake/substrate"}
```
sea212, line 41: Would be good if we could adhere to the general style in this file (lexicographical ordering in any case)
```suggestion
try-runtime-cli = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate", optional = true}
```

node/Cargo.toml
```diff
@@ -148,6 +152,13 @@ runtime-benchmarks = [
     "polkadot-service?/runtime-benchmarks",
     "zeitgeist-runtime/runtime-benchmarks"
 ]
+cli = [
+    'try-runtime-cli',
+]
```
sea212, line 157: We don't need that feature, it's not used anywhere.
```suggestion
```

sea212 (commented):
Thanks for the integrations of this feature, this will help a lot.
Overall the PR looks fine. Functionality wise it's good as it is.

node/Cargo.toml
Chralt98: So you mean, that nowhere in the code we use `feature = "cli"`, I understand. So later, we can't use the command line interface for the storage migrations? I don't really get what this feature could be needed apart from that then. Why would the tutorial (https://docs.substrate.io/how-to-guides/v3/tools/try-runtime/) include this then?

primitives/Cargo.toml
Chralt98: So I removed `frame_support/try-runtime` from the feature `try-runtime`.

node/Cargo.toml
sea212: The tutorial has taken the code from the [substrate node](https://github.com/paritytech/substrate/blob/86b1176d3ba2bd49b87e99e1b074bc0279849021/bin/node/cli/src/lib.rs), which uses a feature `cli` to build the client (otherwise the code will only offer means to create all necessary structures to build your own client). In our case, the client is build by default, there is no feature to toggle building it. That's why we don't need that feature. Our client will always be built and offer the complete try-runtime functionality in case it is built with the `try-runtime` feature.
You can see that the [substrate node template](https://github.com/paritytech/substrate/blob/86b1176d3ba2bd49b87e99e1b074bc0279849021/bin/node-template/node/Cargo.toml#L68-L73) does it the same way, always build the client.

sea212 (approved):

maltekliemann (approved):

Makefile
```diff
@@ -13,6 +13,10 @@ check:
 check-dummy:
 	BUILD_DUMMY_WASM_BINARY= cargo check
 
+.PHONY: try-runtime-upgrade
```
maltekliemann, line 16:
```suggestion
```

Sorry, one more thing: No longer necessary due to #653.

maltekliemann (changes requested):

maltekliemann (approved):
Thanks! :thumbsup:

PR #627, opened 2022-05-15
Benchmark runtime

sea212:
Benchmark the runtime and update the weights. Also adds `parachain-staking` to the list of pallets to benchmarks again.

PR #628, opened 2022-05-15
Downgrade author-mapping and parachain-staking

sea212:
We planned to use the latest `author-mapping` and `parachain-staking` pallets. Unfortunately the migration of `author-mapping` leads to a corrupted storage. It was identified that the nimbus key is not migrated to the new storage.
Upgrading those pallets is not critical and should be postponed to a later release.

### DIFF
--- a/runtime/src/weights/pallet_author_mapping.rs
+++ b/runtime/src/weights/pallet_author_mapping.rs
@@ -30,39 +30,27 @@ use frame_support::{
 /// Weight functions for pallet_author_mapping (automatically generated)
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> pallet_author_mapping::weights::WeightInfo for WeightInfo<T> {
-    // Storage: AuthorMapping MappingWithDeposit (r:1 w:1)
-    // Storage: System Account (r:1 w:1)
-    fn add_association() -> Weight {
-        (72_930_000 as Weight)
-            .saturating_add(T::DbWeight::get().reads(2 as Weight))
-            .saturating_add(T::DbWeight::get().writes(2 as Weight))
-    }
-    // Storage: AuthorMapping MappingWithDeposit (r:2 w:2)
-    fn update_association() -> Weight {
-        (45_930_000 as Weight)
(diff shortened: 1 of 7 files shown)


runtime/src/weights/pallet_author_mapping.rs
```diff
@@ -32,37 +32,25 @@ pub struct WeightInfo<T>(PhantomData<T>);
...
+	#[rustfmt::skip]
     fn add_association() -> Weight {
-        (72_930_000 as Weight)
-            .saturating_add(T::DbWeight::get().reads(2 as Weight))
-            .saturating_add(T::DbWeight::get().writes(2 as Weight))
-    }
+		(34_696_000 as Weight)
+			.saturating_add(T::DbWeight::get().reads(2 as Weight))
+			.saturating_add(T::DbWeight::get().writes(2 as Weight))
+	}
```
maltekliemann, line 40: [nitpick] A couple of tabs snuck in here. I'm okay with that in auto-generated files, but I thought I'd mention it.

runtime/src/weights/parachain_staking.rs
```diff
@@ -27,310 +27,346 @@ use core::marker::PhantomData;
...
-            .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(y as Weight)))
-            .saturating_add(T::DbWeight::get().writes(3 as Weight))
-            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(y as Weight)))
-    }
-    // Storage: ParachainStaking Round (r:1 w:0)
-    fn base_on_initialize() -> Weight {
-        (8_420_000 as Weight)
-            .saturating_add(T::DbWeight::get().reads(1 as Weight))
-    }
+	// Storage: ParachainStaking DelegatorState (r:1 w:1)
```
maltekliemann, line 30: [nitpick] Tabs in this file, as well.

maltekliemann (approved):
Looks good to me! :thumbsup:

runtime/src/weights/pallet_author_mapping.rs
sea212: I will remove them in the next and last PR before v0.3.2 #627

PR #631, opened 2022-05-17
Replace `event_exists` with `assert_last_event`/`assert_has_event`

### LINKED ISSUE #602: [prediction-markets/swaps] Use `frame_system::{assert_has_event, assert_last_event}` instead of homebrew `event_exists`
See https://docs.substrate.io/rustdocs/latest/frame_system/pallet/struct.Pallet.html#method.assert_has_event

maltekliemann:
Closes #602.

### DIFF
--- a/zrml/prediction-markets/src/tests.rs
+++ b/zrml/prediction-markets/src/tests.rs
@@ -1,7 +1,7 @@
 #![cfg(all(feature = "mock", test))]
 
 use crate::{
-    mock::*, Config, Error, Event, MarketIdsPerDisputeBlock, MarketIdsPerReportBlock, RESERVE_ID,
+    mock::*, Config, Error, MarketIdsPerDisputeBlock, MarketIdsPerReportBlock, RESERVE_ID,
 };
 use core::ops::Range;
 use frame_support::{
@@ -360,7 +360,7 @@ fn create_categorical_market_deposits_the_correct_event() {
         let market_id = 0;
         let market = MarketCommons::market(&market_id).unwrap();
         let market_account = PredictionMarkets::market_account(market_id);
-        assert!(event_exists(Event::MarketCreated(0, market_account, market)));
+        System::assert_last_event(Event::PredictionMarkets(crate::Event::MarketCreated(0, market_account, market)));
     });
 }
 
@@ -376,7 +376,7 @@ fn create_scalar_market_deposits_the_correct_event() {
         let market_id = 0;
         let market = MarketCommons::market(&market_id).unwrap();
         let market_account = PredictionMarkets::market_account(market_id);
-        assert!(event_exists(Event::MarketCreated(0, market_account, market)));
+        System::assert_last_event(Event::PredictionMarkets(crate::Event::MarketCreated(0, market_account, market)));
     });
 }
 
@@ -556,7 +556,7 @@ fn it_allows_to_buy_a_complete_set() {
         let market_account = PredictionMarkets::market_account(0);
         let market_bal = Balances::free_balance(market_account);
         assert_eq!(market_bal, CENT);
-        assert!(event_exists(Event::BoughtCompleteSet(0, CENT, BOB)));
+        System::assert_last_event(Event::PredictionMarkets(crate::Event::BoughtCompleteSet(0, CENT, BOB)));
     });
 }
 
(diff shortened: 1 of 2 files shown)


zrml/prediction-markets/src/tests.rs
```diff
@@ -1,7 +1,7 @@
 #![cfg(all(feature = "mock", test))]
 
 use crate::{
-    mock::*, Config, Error, Event, MarketIdsPerDisputeBlock, MarketIdsPerReportBlock, RESERVE_ID,
+    mock::*, Config, Error, MarketIdsPerDisputeBlock, MarketIdsPerReportBlock, RESERVE_ID,
```
Chralt98, line 4:
```suggestion
    mock::*, Config, Error, Event as PEvent, MarketIdsPerDisputeBlock, MarketIdsPerReportBlock, RESERVE_ID,
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -360,7 +360,11 @@ fn create_categorical_market_deposits_the_correct_event() {
         let market_id = 0;
         let market = MarketCommons::market(&market_id).unwrap();
         let market_account = PredictionMarkets::market_account(market_id);
-        assert!(event_exists(Event::MarketCreated(0, market_account, market)));
+        System::assert_last_event(Event::PredictionMarkets(crate::Event::MarketCreated(
+            0,
+            market_account,
+            market,
+        )));
```
Chralt98, line 367:
```rust
fn check_last(event: PEvent<Runtime>) {
    System::assert_last_event(Event::PredictionMarkets(event));
}
```

```suggestion
       
check_last(PEvent::MarketCreated(
            0,
            market_account,
            market,
        ));
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -1503,37 +1515,30 @@ fn full_scalar_market_lifecycle() {
...
             50 * BASE,
             1250 * CENT, // 12.5
-            CHARLIE
+            CHARLIE,
         )));
 
         assert_ok!(PredictionMarkets::redeem_shares(Origin::signed(EVE), 0));
         let ztg_bal_eve_after = Balances::free_balance(&EVE);
         assert_eq!(ztg_bal_eve_after, 101250 * CENT); // 12.5 (SHORT) + 1000 (balance)
-        assert!(!event_exists(Event::TokensRedeemed(
```
Chralt98, line 1524: Event should not exist here. You wrote this @maltekliemann one month ago. Is there a specific reason. Maybe use `!assert_has_event`?

Chralt98 (changes requested):
The `frame_system` call `assert_last_event` is much more precise than the method before. This is a good improvement to prevent future test bugs.

Chralt98:
I wonder why there are no event checks for the other pallets.

zrml/prediction-markets/src/tests.rs
maltekliemann: Testing for an event _not_ to exist seems like a bad idea in hindsight. Every dispatch should not emit most events.

But... If there was a buildin function that could test for the lack of events, I would have kept the test. But `assert_has_event` doesn't return a bool, it just panicks if the event doesn't exist, so `!assert_has_event` won't compile.

maltekliemann:
> I wonder why there are no event checks for the other pallets.

Most other pallets just don't have any events. The events of `court` will be tested in the new iteration of that pallet.

zrml/prediction-markets/src/tests.rs
maltekliemann: The problem here is that we may want to check for events emitted from other pallets. But I've managed to improve the aesthetics a bit.

Chralt98 (approved):
LGTM

maltekliemann:
> Yeah, I meant the pallets `court, liquidity-mining, orderbook-v1`. You probably mean [this](https://github.com/zeitgeistpm/zeitgeist/pull/642) for the `court` pallet. Apart from that, I think `orderbook-v1` doesn't matter anymore? So `liquidity-mining` is left.

Yeah, `liquidity-mining` very much needs to be tested before we start using it (it's currently disbabled in the runtime; see https://github.com/zeitgeistpm/zeitgeist/blob/ab2a8e629d736f5fdb744583b232300b0918adef/runtime/src/lib.rs#L295-L401).

By new iteration of court I mean these incoming changes: https://github.com/zeitgeistpm/zeitgeist/pull/642

PR #633, opened 2022-05-17
Remove `MarketCommons::report`

maltekliemann:
This PR removes the `report` function from the `MarketCommonsApi`:

Reason for removal:

- It's used exactly twice in the entire code (one of these uses seems redundant)
- It's purpose if strongly related to the `prediction-markets` pallet, _not_ `market-commons`
- It returns the `market-commons`-specific error `NoReport`; but the `prediction-markets` pallet has its own error `MarketIsNotReported` - so now we can also remove the essentially duplicate `NoReport` error


sea212 (approved):
Fine from my side.

PR #637, opened 2022-05-19
Let `admin_move_market_to_closed` error on non-active market

### LINKED ISSUE #612: [prediction-markets] Audit `admin_move_market_to_closed`
It seems like there are a couple of things wrong with `admin_move_market_to_closed`:

- It can be run on any market, not just an active one
- ~~It doesn't seem like it takes Rikiddo markets which are collecting subsidy into account~~

maltekliemann:
Fixes #612 by letting `admin_move_market_to_closed` error if the market is not actually active. Also adds lots of tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12713,6 +12713,7 @@ dependencies = [
  "frame-benchmarking",
  "frame-support",
  "frame-system",
+ "more-asserts",
  "orml-currencies",
  "orml-tokens",
  "orml-traits",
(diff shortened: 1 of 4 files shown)


sea212 (approved):
Makes sense, the PR is perfectly fine imo. I am just wondering how changing the `end` time could interfere with the state transition to `Closed` (in respect to the market state transition solution you're currently working on)

maltekliemann:
> I am just wondering how changing the `end` time could interfere with the state transition to `Closed` (in respect to the market state transition solution you're currently working on)

The other PR completely rewrites this function. There's no more changes to the `period`, then.

PR #642, opened 2022-05-20
Update `court` pallet

maltekliemann:
Some of the changes to the `court` pallet are made in accordance to the recent changes to the specification of the `DisputeApi`:

- Let `on_resolution` return `None` if none of the jurors voted.
- Replace `TreasuryPalletId` with `Slash` trait. The purpose of this is to fix a simple design problem: Before this change, the `court` pallet used the `TreasuryPalletId` constant to calculate the treasury account. However, if the `treasury` pallet changes the way if calculates the treasury account, this breaks the `court` pallet. Instead, we now take a cue from the democracy pallet and replace the `TreasuryPalletId` with `type Slash: OnUnbalance<...>` and move unbalances created by slashing into the treasury using `Slash::on_unbalance`.
- Catch some errors in `vote` that were not considered earlier.
- Improve tests:
  - Replace `unwrap()` with `assert_ok!`.
  - Lower block amounts to shorten test duration.


maltekliemann:
Dropped in favor of a complete re-write of the court system.

PR #649, opened 2022-05-24
Fix and test invalid outcome ranges

maltekliemann:
It's currently possible to use a `range` with `range.end() <= range.start()` in `create_scalar_market`. This doesn't cause any serious errors (SHORT _always_ redeems for 1 ZTG, LONG for 0 ZTG), but obviously we don't want this situation.

We now check for this case in `create_scalar_markets`.


Chralt98 (changes requested):
The effective extrinsic logic looks good. There are some comment improvements.

Chralt98 (approved):
LGTM

PR #664, opened 2022-06-07
Fix ScoringRule checks with match

Chralt98:
Fixes https://github.com/zeitgeistpm/runtime-audit-1/issues/46


maltekliemann (changes requested):
Looks great, just a tiny change required! :thumbsup:

Also, a style question: https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md suggests that...

> - `match` arms may be either blocks or have a trailing , but not both.
> - Blocks should not be used unnecessarily.

I don't know if this is the way to go. How do you feel about this?

Chralt98:
You meant this https://github.com/zeitgeistpm/zeitgeist/pull/664/commits/664b74429fef2992884b70e729999064dc2fb4a9 ? I agree with the style guide of Parity.

zrml/swaps/src/utils.rs
```diff
@@ -157,56 +157,63 @@ where
...
+    match p.pool.scoring_rule {
+        ScoringRule::CPMM => {
+            ensure!(spot_price_after >= spot_price_before, Error::<T>::MathApproximation);
+        }
+        ScoringRule::RikiddoSigmoidFeeMarketEma => {
+            ensure!(
+                spot_price_before.saturating_sub(spot_price_after) < 20u8.into(),
+                Error::<T>::MathApproximation
+            );
+        }
```
maltekliemann, line 194: If we're going with the Parity style guide, these blocks are unnecessary.

zrml/swaps/src/utils.rs
```diff
@@ -157,56 +157,63 @@ where
...
-        T::RikiddoSigmoidFeeMarketEma::update_volume(p.pool_id, volume)?;
+    match p.pool.scoring_rule {
+        ScoringRule::CPMM => {
+            ensure!(
+                spot_price_before
+                    <= bdiv(asset_amount_in.saturated_into(), asset_amount_out.saturated_into())?
+                        .saturated_into(),
+                Error::<T>::MathApproximation
+            );
+        }
```
maltekliemann, line 209: If we're going with the Parity style guide, this block is unnecessary.

zrml/swaps/src/lib.rs
```diff
@@ -1365,75 +1365,85 @@ mod pallet {
...
-                }
+            let (pool_status, total_subsidy, total_weight, weights, pool_amount) =
+                match scoring_rule {
+                    ScoringRule::CPMM => {
+                        ensure!(amount.is_some(), Error::<T>::InvalidAmountArgument);
+                        ensure!(
+                            amount_unwrapped >= T::MinLiquidity::get(),
+                            Error::<T>::InsufficientLiquidity
+                        );
+                        swap_fee.ok_or(Error::<T>::InvalidFeeArgument)?;
```
maltekliemann, line 1376: You could also got with `ensure!(swap_fee.is_some())` here. But this is nitpick territory. ;)

maltekliemann (approved):
According to the Parity style guide, two `match` expressions in `utils.rs` need to be reformatted. You decide if this actually improves the code.

Other than that, yeah, good stuff.

zrml/swaps/src/utils.rs
Chralt98: `cargo fmt` wants to use blocks here somehow. I would probably stick with `cargo fmt` here, because when not using it, the workflow fails.

zrml/swaps/src/lib.rs
Chralt98: No, I think it's a really good suggestion, because `ok_or` unwraps the value, which is not needed here.

zrml/swaps/src/utils.rs
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/664/commits/ef4fe627e82b21b6bca6fa442f64b7db6fbaded7

zrml/swaps/src/lib.rs
```diff
@@ -2011,10 +2033,13 @@ mod pallet {
...
-                Ok(T::WeightInfo::swap_exact_amount_out_cpmm())
-            } else {
-                Ok(T::WeightInfo::swap_exact_amount_out_rikiddo(pool.assets.len().saturated_into()))
+            match pool.scoring_rule {
+                ScoringRule::CPMM => Ok(T::WeightInfo::swap_exact_amount_out_cpmm()),
+                ScoringRule::RikiddoSigmoidFeeMarketEma => {
+                    Ok(T::WeightInfo::swap_exact_amount_out_rikiddo(
+                        pool.assets.len().saturated_into(),
+                    ))
+                }
```
sea212, line 2042: Whenever we leave the (theoretically unnecessary) curly brackets away, `cargo fmt` will automatically insert them when the body of that match arm cannot fit in one line. It seems that this is required by the rust styleguide, as stated by the [cargo fmt configuration overview](https://rust-lang.github.io/rustfmt/?version=v1.5.0&search=#match_arm_blocks). There is another case in *swaps/src/utils.rs*. `Cargo fmt` does allow to configure this, but I think it makes sense to adhere to the rust style guide and from my opinion the curly brackets also make it more readable.

sea212 (approved):
Thanks Chris, now we will never forgot to handle new `ScoringRules` in the existing functions :muscle:

Issue #497, opened 2022-03-21
[swaps] Calling `create_pool` with duplicate assets should raise an error


maltekliemann:
Currently, you can do the following without raising an error.

```rust
Swaps::create_pool(
    BOB,
    vec!(ASSET_A, ASSET_A),
    None,
    0,
    ScoringRule::CPMM,
    Some(0),
    Some(vec!(_2, _2)),
);
```

This shouldn't be a problem, given that this is only an internal function, but it should be fixed regardless.

PR #666, opened 2022-06-08
Pool creation should not allow duplicate assets.

vivekvpandya:
This fixes https://github.com/zeitgeistpm/zeitgeist/issues/497

### DIFF
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -1345,7 +1345,7 @@ mod pallet {
         #[frame_support::transactional]
         fn create_pool(
             who: T::AccountId,
-            mut assets: Vec<Asset<T::MarketId>>,
+            assets: Vec<Asset<T::MarketId>>,
             base_asset: Asset<T::MarketId>,
             market_id: Self::MarketId,
             scoring_rule: ScoringRule,
@@ -1362,12 +1362,11 @@ mod pallet {
             let mut map = BTreeMap::new();
             let mut total_weight = 0;
             let amount_unwrapped = amount.unwrap_or_else(BalanceOf::<T>::zero);
-            // Sort assets for future binary search, for example to check if an asset is included.
-            let sort_assets = assets.as_mut_slice();
-            sort_assets.sort();
-            let has_duplicates = sort_assets
+            let mut sorted_assets = assets.clone();
+            sorted_assets.sort();
+            let has_duplicates = sorted_assets
                 .iter()
-                .zip(sort_assets.iter().skip(1))
+                .zip(sorted_assets.iter().skip(1))
                 .fold(false, |acc, (&x, &y)| acc || x == y);
             ensure!(!has_duplicates, Error::<T>::SomeIdenticalAssets);
 
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -1406,7 +1406,7 @@ mod pallet {
             }
 
             let pool = Pool {
-                assets,
+                assets: sorted_assets,
                 base_asset,
                 market_id,
                 pool_status: if scoring_rule == ScoringRule::CPMM {
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -807,6 +807,8 @@ mod pallet {
         /// Tried to create a pool that has more assets than the upper threshhold specified by
         /// a constant.
         TooManyAssets,
+        /// Tried to create a pool with at least two identical assets.
+        SomeIdenticalAssets,
         /// The pool does not support swapping the assets in question.
         UnsupportedTrade,
         /// The outcome asset specified as the winning asset was not found in the pool.
@@ -1360,6 +1362,14 @@ mod pallet {
             let mut map = BTreeMap::new();
             let mut total_weight = 0;
             let amount_unwrapped = amount.unwrap_or_else(BalanceOf::<T>::zero);
+            // Sort assets for future binary search, for example to check if an asset is included.
+            let sort_assets = assets.as_mut_slice();
+            sort_assets.sort();
+            let has_duplicates = sort_assets
+                .iter()
+                .zip(sort_assets.iter().skip(1))
+                .fold(false, |acc, (&x, &y)| acc || x == y);
+            ensure!(!has_duplicates, Error::<T>::SomeIdenticalAssets);
 
             if scoring_rule == ScoringRule::CPMM {
                 ensure!(amount.is_some(), Error::<T>::InvalidAmountArgument);
@@ -1396,9 +1406,6 @@ mod pallet {
                 let _ = T::RikiddoSigmoidFeeMarketEma::create(next_pool_id, rikiddo_instance)?;
             }
 
-            // Sort assets for future binary search, for example to check if an asset is included.
-            let sort_assets = assets.as_mut_slice();
-            sort_assets.sort();
             let pool = Pool {
                 assets,
                 base_asset,
--- a/zrml/swaps/src/tests.rs
+++ b/zrml/swaps/src/tests.rs
@@ -51,6 +51,28 @@ const _105: u128 = 105 * BASE;
 const _1234: u128 = 1234 * BASE;
 const _10000: u128 = 10000 * BASE;
 
+#[test_case(vec![ASSET_A, ASSET_A]; "short vector")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_A]; "start and end")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_E]; "successive at end")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "start and middle")]
+fn create_pool_fails_with_duplicate_assets(assets: Vec<Asset<<Runtime as Config>::MarketId>>) {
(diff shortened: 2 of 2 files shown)


Chralt98 (changes requested):
I like your solution for finding duplicates, because we need to sort the asset vector anyways. Keep in mind, that this comes with the cost, that the sorting needs to be done earlier as before. But I am okay with that, because it shouldn't be too much work. Please add some more complex test scenarios.

zrml/swaps/src/lib.rs
```diff
@@ -803,6 +803,8 @@ mod pallet {
         /// Tried to create a pool that has more assets than the upper threshhold specified by
         /// a constant.
         TooManyAssets,
+        /// Tried to create a pool that has duplicate assets.
+        DuplicateAssets,
```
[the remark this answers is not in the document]
maltekliemann: I'm more in favor of `DuplicateAssets`. I guess we'll have general discussion about naming errors soon to streamline the naming process.

zrml/swaps/src/lib.rs
```diff
@@ -1353,6 +1355,14 @@ mod pallet {
             let mut map = BTreeMap::new();
             let mut total_weight = 0;
             ensure!(assets.contains(&base_asset), Error::<T>::BaseAssetNotFound);
+            // Sort assets for future binary search, for example to check if an asset is included.
+            let sort_assets = assets.as_mut_slice();
+            sort_assets.sort();
+            let has_dups = sort_assets
+                .iter()
+                .zip(sort_assets.iter().skip(1))
+                .fold(false, |acc, (&x, &y)| acc || x == y);
```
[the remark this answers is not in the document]
maltekliemann: I'd rename `has_dups` into `has_duplicates`. Abbreviations in variable names should generally be avoided.

zrml/swaps/src/tests.rs
```diff
@@ -49,6 +49,24 @@ const _101: u128 = 101 * BASE;
...
+                ASSET_A,
+                0,
+                ScoringRule::CPMM,
+                Some(0),
+                Some(vec!(_2, _2)),
+            ),
+            crate::Error::<Runtime>::DuplicateAssets
+        );
+    });
+}
```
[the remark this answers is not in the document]
maltekliemann: I second this. Using `test_cases`, this should be relatively simple.

maltekliemann (changes requested):
Look fine. :thumbsup:

One more thing: We formulate commit messages in the imperative, so you should use something like "Disallow pool creation with duplicate assets" instead of "Pool creation should not allow duplicate assets."

Chralt98 (changes requested):
Apart from the base asset check I am okay with that. Well done!

zrml/swaps/src/tests.rs
```diff
@@ -51,6 +51,28 @@ const _105: u128 = 105 * BASE;
 const _1234: u128 = 1234 * BASE;
 const _10000: u128 = 10000 * BASE;
 
+#[test_case(vec![ASSET_A, ASSET_A]; "case-1 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_A]; "case-2 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_E]; "case-3 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "case-4 with duplicate assets")]
+fn pool_creation_fails_with_duplicate_assets(assets: Vec<Asset<<Runtime as Config>::MarketId>>) {
```
maltekliemann, line 58: I usually go by this pattern when naming tests: `{function_under_test}_{behavior_during_test}`

```suggestion
fn create_pool_fails_with_duplicate_assets(assets: Vec<Asset<<Runtime as Config>::MarketId>>) {
```

zrml/swaps/src/tests.rs
```diff
@@ -51,6 +51,28 @@ const _105: u128 = 105 * BASE;
 const _1234: u128 = 1234 * BASE;
 const _10000: u128 = 10000 * BASE;
 
+#[test_case(vec![ASSET_A, ASSET_A]; "case-1 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_A]; "case-2 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_E]; "case-3 with duplicate assets")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "case-4 with duplicate assets")]
```
maltekliemann, line 57: Not a big fan of the test names: `"case-{i} with duplicate assets"`. This means that the full test name is `pool_creation_fails_with_duplicate_assets::case_{i}_with_duplicate_assets`. I'd either go with detailed descriptions, or, if you want to go with running numbers, I'd just put `1`, `2`, etc.

```suggestion
#[test_case(vec![ASSET_A, ASSET_A]; "short vector")]
#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_A]; "start and end")]
#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_E]; "successive at end")]
#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "start and middle")]
```

maltekliemann (changes requested):
Okay, just a very small critique, then we're good to go!

Chralt98 (approved):
Looks pretty to me now.

zrml/swaps/src/tests.rs
```diff
@@ -51,6 +51,28 @@ const _105: u128 = 105 * BASE;
 const _1234: u128 = 1234 * BASE;
 const _10000: u128 = 10000 * BASE;
 
+#[test_case(vec![ASSET_A, ASSET_A]; "short vector")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_A]; "start and end")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_D, ASSET_E, ASSET_E]; "successive at end")]
+#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "start end middle")]
```
maltekliemann, line 57:
```suggestion
#[test_case(vec![ASSET_A, ASSET_B, ASSET_C, ASSET_A, ASSET_E, ASSET_D]; "start and middle")]
```

maltekliemann (changes requested):

zrml/swaps/src/lib.rs
```diff
@@ -1360,6 +1362,14 @@ mod pallet {
...
             let mut total_weight = 0;
             let amount_unwrapped = amount.unwrap_or_else(BalanceOf::<T>::zero);
+            // Sort assets for future binary search, for example to check if an asset is included.
+            let sort_assets = assets.as_mut_slice();
+            sort_assets.sort();
+            let has_duplicates = sort_assets
+                .iter()
+                .zip(sort_assets.iter().skip(1))
+                .fold(false, |acc, (&x, &y)| acc || x == y);
+            ensure!(!has_duplicates, Error::<T>::SomeIdenticalAssets);
```
maltekliemann, line 1372: Sorry, just noticed another problem. You need to either clone `assets` or put this code blob _after_ the next `if` expression, because the order of `assets` needs to remain unchanged until the weight map is created in case the pool uses CPMM as scoring rule.

For design reasons, I would make `assets` non-mutable and clone it to prevent these types of errors in the future.

zrml/swaps/src/tests.rs
```diff
@@ -51,6 +51,28 @@ const _105: u128 = 105 * BASE;
...
+                ASSET_A,
+                0,
+                ScoringRule::CPMM,
+                Some(0),
+                Some(<Runtime as crate::Config>::MinLiquidity::get()),
+                Some(vec!(_2, _2)),
+            ),
+            crate::Error::<Runtime>::SomeIdenticalAssets
+        );
+    });
```
maltekliemann, line 73: I discovered another style problem here: Calling `create_pool` with these parameters will result in an error even if we don't error on `SomeIdenticalAssets`. For example, `BOB`, doesn't have enough funds, and the number of weights specified don't always match the number of assets.

```suggestion
    ExtBuilder::default().build().execute_with(|| {
        assets.iter().cloned().for_each(|asset| {
            let _ = Currencies::deposit(asset, &BOB, _10000);
        });
        let asset_count = assets.len();
        assert_noop!(
            Swaps::create_pool(
                BOB,
                assets,
                ASSET_A,
                0,
                ScoringRule::CPMM,
                Some(0),
                Some(<Runtime as crate::Config>::MinLiquidity::get()),
                Some(vec![_2; asset_count]),
            ),
            crate::Error::<Runtime>::SomeIdenticalAssets
        );
    });
```

This change isn't _necessary_, but I feel like it's quite sensible.

maltekliemann (changes requested):

zrml/swaps/src/lib.rs
Chralt98: good catch 👍

maltekliemann (changes requested):

maltekliemann (approved):
Cool, great addition! :thumbsup:

Issue #367, opened 2021-11-03
Disputed market skipped resolution, but has a resolved outcome


lsaether:
Market id - 155

![image](https://user-images.githubusercontent.com/25539605/140071997-28b32764-4963-40ee-8727-b1f72564569d.png)

lsaether:
This may have been caused by a timeout in block production, experienced when we had the first Battery Station network stall. We should do an audit of the entire pallet though.

Chralt98:
When I query the market with the id `155` on Battery Station I get the following result.

```text
{
  creator: dE2ByiLsmPMHKBeAUywwHWw3WSsfHfr73imKA5Fhgz5MYebrr
  creation: Permissionless
  creatorFee: 0
  oracle: dE2ByiLsmPMHKBeAUywwHWw3WSsfHfr73imKA5Fhgz5MYebrr
  metadata: 0x1530b1ee6b8ede326178e4a53ab17ed3d999cc86e5cd2f7aceaf15996cdba7f13174865e121e6d842e433c3de9fff10d125a
  marketType: {
    Categorical: 2
  }
  period: {
    Timestamp: [
      1,635,208,400,363
      1,635,721,148,108
    ]
  }
  scoringRule: CPMM
  status: Resolved
  report: {
    at: 158,870
    by: dE2ByiLsmPMHKBeAUywwHWw3WSsfHfr73imKA5Fhgz5MYebrr
    outcome: {
      Categorical: 0
    }
  }
  resolvedOutcome: {
    Categorical: 0
  }
  mdm: SimpleDisputes
}
```

Why do our outputs differ?

PR #669, opened 2022-06-09
Fix disputed market after resolve bug

Chralt98:
Fixes https://github.com/zeitgeistpm/zeitgeist/issues/367

There was no market state check in the `dispute` extrinsic other than `MarketStatus::Disputed` itself. I included the check of `MarketStatus::Resolved`, but I am not sure, if there is another check to make here.

Maybe the state has to be `MarketStatus::Reported` to call the extrinsic `dispute`. That would make the most sense regarding [this](https://github.com/zeitgeistpm/zeitgeist/blob/main/docs/diagrams/market_state_diagram.png).

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -349,7 +349,10 @@ mod pallet {
             let disputes = Disputes::<T>::get(market_id);
             let curr_block_num = <frame_system::Pallet<T>>::block_number();
             let market = T::MarketCommons::market(&market_id)?;
-            ensure!(market.status != MarketStatus::Resolved, Error::<T>::InvalidMarketStatus);
+            ensure!(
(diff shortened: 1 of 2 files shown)


Chralt98:
I found something even better.

```rust
ensure!(
                matches!(market.status, MarketStatus::Reported | MarketStatus::Disputed),
                Error::<T>::InvalidMarketStatus
);
```

maltekliemann (changes requested):
Great catch! :thumbsup:

It think testing should be more thorough. Regarding the underlying issue: We would have to filter through all events pertaining this market to get a better idea, but I think it's ok to assume that this was the bug that caused the anomaly.

maltekliemann (approved):
:thumbsup:

sea212 (approved):
Well done!

Issue #90, opened 2021-04-22
Automatically determine initial swap pool weights / set initial prices


sea212:
**Issue**
Currently the creator of a swaps pool has to specify the initial weights (see [here](https://github.com/zeitgeistpm/zeitgeist/blob/5bef6d1a24b70d3e8361dc55fc0c3a44a859d520/zrml/swaps/src/lib.rs#L68) and [here](https://github.com/zeitgeistpm/zeitgeist/blob/5bef6d1a24b70d3e8361dc55fc0c3a44a859d520/zrml/prediction-markets/src/lib.rs#L358)), which determine the price ratio between the assets. In addition to the circumstance, that [deploy_swap_pool_for_market](https://github.com/zeitgeistpm/zeitgeist/blob/5bef6d1a24b70d3e8361dc55fc0c3a44a859d520/zrml/prediction-markets/src/lib.rs#L373-L376) does implicitly add the market currency to the assets, but not extend the weights, i.e. expects the user to know that this implicit expansion happens and to manually add the weight, leads to an unintuitive and error-prone situation.

**Potential solution**
TBD. How do we determine the initial weights? Should the user be able to modify them?

maltekliemann:
The use should be able to specify weights. This is important if they believe one outcome to be much more likely or unlikely than another.

My solution would be to add a `base_asset_weight` parameter.

Chralt98:
I like your approach @maltekliemann. It's cleaner to separate the `asset_weights` and the `base_asset_weight`.

Another approach could be to check if the user forgot to specify a weight for the base asset.

```rust
let mut assets = Self::outcome_assets(market_id, &market);
ensure!(
    weights.len() == assets.len().saturating_add(One::one()),
    Error::<T>::WeightsLenMustEqualAssetsLenPlusOneForBaseAsset
);
let base_asset = Asset::Ztg;
assets.push(base_asset);
```

So the last weight would always be for the `base_asset`.

But don't get me wrong, I would go with your solution @maltekliemann, because it's more precise.

maltekliemann:
Hey @Chralt98. I didn't realize this issue still existed. This is basically solved in #672 by allowing the user to set the weights of the outcome tokens (since this allows them to control the initial price of the tokens), but fixing the weight of the base asset to be the sum of the weights of the outcome tokens. Sorry I didn't mark this issue. Completely forgot about it.

[...]

PR #672, opened 2022-06-13
Fix the weight of the base asset to the sum of the weights of the outcome assets

### LINKED ISSUE #667: [swapsl Ensure that the total weight of outcome tokens is equal to the weight of the base asset
This ensures that the sum of all spot prices of outcome assets in a pool are equal to 1 (assuming equal balances)

maltekliemann:
Closes #667 and #90.

The calculation of the base asset is done in `prediction-markets`. The `swaps` pallet does _not_ enforce that the weight of the base asset equals the sum of the weights of the outcome assets. You might think that it should. I disagree. To be honest, the fact that the pools in `swaps` even have a `base_asset` seems more and more like a code smell to me. Isn't `base_asset` purely an aspect of prediction markets? I think `Pool` should be base-asset-agnostic.

The idea of this change is that setting the weight of the base asset to be equal to the total weight of the outcome assets means that when the market is created, the prices of the outcome assets sum to 1 (assuming equal balances). This quickly follows from the formula for the spot price from the [balancer whitepaper](balancer.fi/whitepaper.pdf).


zrml/swaps/src/tests.rs
```diff
@@ -1988,6 +1988,31 @@ fn pool_exit_with_exact_pool_amount_fails_if_max_asset_amount_is_violated() {
...
+            Some(vec!(_1, _2, _3, _4)),
+        ));
+        let pool = Swaps::pool(0).unwrap();
+        let pool_weights = pool.weights.unwrap();
+        assert_eq!(pool_weights[&ASSET_A], _4);
+        assert_eq!(pool_weights[&ASSET_B], _2);
+        assert_eq!(pool_weights[&ASSET_C], _3);
+        assert_eq!(pool_weights[&ASSET_D], _1);
+    });
+}
```
Chralt98, line 2014: This test would have failed if the order changed in the function like with the sort functionality previously. Thanks for this important test.

Chralt98 (commented):
This PR calculates the base asset weight to be the sum of all outcome assets. Therefore it's easy to ensure, that a set of outcomes sum up to one base asset. Great addition!

zrml/prediction-markets/src/lib.rs
```diff
@@ -613,6 +617,8 @@ mod pallet {
             let mut assets = Self::outcome_assets(market_id, &market);
             let base_asset = Asset::Ztg;
             assets.push(base_asset);
+            let base_asset_weight = weights.iter().fold(0, |acc, val| acc.saturating_add(*val));
```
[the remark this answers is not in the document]
maltekliemann: Hmm... that's a very good question. I can't really give a good answer, but I did some research.

There's a trait `Saturating` in substrate, which is implemented for the integer types using a blanket implementation: https://github.com/paritytech/substrate/blob/06ab196c201f93fb4560dc08ff8a94f3070a1dbe/primitives/arithmetic/src/traits.rs#L243-L274

I don't know exactly how type deduction works in Rust, but I'd say `saturating_add` gets `acc` and `val` as parameter. By the definition of `Saturating`, they have the same type. But `val` comes from `weights`, so it's a `u128`. Thus, `acc` is a `u128`.

You can even do `0.saturating_add(1)` or `0.saturating_add(1u32)` and then Rust would assume that both numbers are `i32` and `u32`, resp. (`i32` is the default).

On the other hand, the core implementation of `u128` is done using the [`impl_uint` macro](https://github.com/rust-lang/rust/blob/master/library/core/src/num/int_macros.rs) here: https://github.com/rust-lang/rust/blob/d40f24e956a698e47a209541031c4045acc5a684/library/core/src/num/mod.rs#L857-L866,
and apparently this implementation behaves fundamentally different from the trait based implementation. For example, you can't do `0.saturating_add(1u32)` when using the default implementation. I have no idea why that is.

The error E0689 is defined as "A method was called on an ambiguous numeric type." Just seems strange to me that in an expression `0.fun(1)` for some `trait Fun { fn fun(self, other: Self); }` which is implemented for some integer types, the integers are well-defined `i32`, but somehow `0` and `1` in `0.saturating_add(1)` are ambiguous.

Anyways, if you figure it out, please let me know.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -516,15 +516,16 @@ mod tests {
...
             Origin::signed(ALICE),
             BOB,
             period,
             gen_metadata(0),
-            MarketType::Categorical(5),
+            MarketType::Categorical(category_count),
             MarketDisputeMechanism::Authorized(CHARLIE),
             amount,
-            vec![BASE; 6],
+            vec![BASE; category_count.into()],
```
sea212, line 528: We should not change existing migrations, since when the storage migration is executed, it checks if the storage version equals to a version where those values were still valid. I think you know that already, the PR is just a bit older.
However, since https://github.com/zeitgeistpm/zeitgeist/pull/680 removes all the code, a merge from main will automatically fix this.

sea212 (approved):
LGTM :+1:

zrml/prediction-markets/src/lib.rs
```diff
@@ -551,7 +551,9 @@ mod pallet {
         ///
         /// * `market_id`: The id of the market.
         /// * `amount`: The amount of each token to add to the pool.
-        /// * `weights`: The relative denormalized weight of each asset price.
+        /// * `weights`: The relative denormalized weight of each outcome asset. The sum of the
+        ///     weights must be less of equal to _half_ of the `MaxTotalWeight` constant of the
+        ///     swaps pallet.
```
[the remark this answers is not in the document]
maltekliemann: Yes, exactly. In order to satisfy the requirements of `create_pool`, we need to make sure that the outcome assets only take at most half of the possible weight.

I agree that this isn't ideal. The reasoning here is that `MaxTotalWeight` is a constant of `swaps`. The `prediction-markets` pallet builds on top of that and needs to make additional assumptions.

zrml/prediction-markets/src/migrations.rs
maltekliemann: Fixed in b4d77450266e713a67a89cde0807bfe4fb46ad34

sea212: There are some exceptions though, for example migrations that adjust the storage based on the value of a configuration value. Such migrations are thought to be used anytime a configuration value changes that for example affects the contents of the storage, thus they should be maintained ([see this example](https://github.com/PureStake/moonbeam/pull/1096))

Chralt98 (approved):
👍

PR #673, opened 2022-06-13
Remove ICE workaround & fix clippy errors

### LINKED ISSUE #608: [prediction-markets] Remove workaround for ICE caused by `#[transactional]` use
Applying `#[transactional]` has caused an ICE in #576. The ICE is tracked in https://github.com/rust-lang/rust/issues/96299. We have a workaround, but we should remove it as soon as the issue in the rust is resolved.

maltekliemann:
Closes #608.

After I updated to the newest nightly build, clippy had a couple of complaints, which I fixed as well.

### DIFF
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -185,7 +185,13 @@ fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Public
 /// The extensions for the [`ChainSpec`].
 #[cfg(feature = "parachain")]
 #[derive(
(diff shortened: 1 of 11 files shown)


Chralt98 (commented):
I think `let _ =` was useless, which was detected by clippy. I don't get why `Eq` is required for that `enum`'s and `struct`'s? You didn't add equal checks for that structs in this PR. Because `#[transactional]` needs some sort of equal check?

maltekliemann:
@Chralt98 `Eq` was added to satisfy this lint: https://rust-lang.github.io/rust-clippy/master/#derive_partial_eq_without_eq

Chralt98 (approved):

PR #674, opened 2022-06-13
Force user to use limits when swapping tokens

maltekliemann:
Currently, it's possible to call `swap_exact_amount_in` and `swap_exact_amount_out` without any limits on token amounts of price. Doing so on a public network where frontrunners do mischief might have a devastating effect on the user. This PR makes sure that the user cannot accidentally set both limits to `None`.

Instead, we now raise an error when both limits are `None`. If the user absolutely wants to go full hog and disable all safeties, they can still do so by setting `max_price` to `u128::MAX`.


maltekliemann:
@vivekvpandya Yeah, that's a good question! A little bit of history on this subject: Until recently, both parameter were non-optional. This turned out to be annoying during the development of the frontend (they usually use one of the two limits, so the other one needs to be disabled), so we decided to make both parameters optional, leaving us with the problem that the user might _accidentally_ (this is a very important word in this context) disable both safeties. This is what this PR is supposed to fix. Now the frontend team can specify `None` to disable one of these safeties.

You can _always_ disable both safeties, even if they are not optional: You can just put zero or `u128::MAX`, depending on the context. However, if we _force_ the user to do that, we basically ensure that they didn't disable the safeties by mistake, but by their own volition.

vivekvpandya (approved):
Looks fine to me!

Chralt98 (approved):
Default disabling of both safety features is now restricted. This could have been prevented even without the PR in a good front-end, but there is the possibility of front-runners on the polkadot-js site. So this made the interface more nooby friendly 👍

maltekliemann:
Thanks for the review, guys! :thumbsup:

PR #675, opened 2022-06-14
Bump versions to v0.3.3

sea212:
Update Zeitgeist crate version from *v0.3.2* to *v0.3.3*. Reflect updates in *Cargo.lock*


Chralt98 (approved):
When comparing to [this](https://github.com/zeitgeistpm/zeitgeist/commit/b0d1e04fc427028e604b71e6cad230937ea40217) it looks like the same pattern. So I would agree (with the assumption, that last time it was correct).

PR #678, opened 2022-06-16
Increment RuntimeVersion

sea212:
Increments `spec_version`: Pallets changed in a way that the pallets logic is not idempotent to the current release.
Increments `transaction_version`: Pallet ordering, dispatachables and parameter ordering in dispatchables was changed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12410,7 +12410,7 @@ dependencies = [
 
 [[package]]
 name = "zeitgeist-node"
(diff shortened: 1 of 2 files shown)


sea212 (commented):
Verified in https://github.com/zeitgeistpm/zeitgeist/compare/v0.3.2...bef0b3461dfcce4f81fc361f729eefa340bf19d1

`RuntimeVersion` did not change since v0.3.2

PR #679, opened 2022-06-16
Update weights

sea212:
This PR adds the weights benchmarked on the reference machine based on the latest changes.


runtime/src/weights/frame_system.rs
```diff
@@ -32,7 +32,7 @@ use frame_support::{
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> frame_system::weights::WeightInfo for WeightInfo<T> {
     fn remark(_b: u32) -> Weight {
-        (8_454_000 as Weight)
+        (0 as Weight)
```
sea212, line 35: This is strange, it should not be `0`. I'll revert this to the previous value, since `remark` did not change a bit.
```suggestion
        (8_454_000 as Weight)
```

runtime/src/weights/pallet_scheduler.rs
```diff
@@ -36,9 +36,9 @@ impl<T: frame_system::Config> pallet_scheduler::weights::WeightInfo for WeightIn
     // Storage: Preimage StatusFor (r:1 w:1)
     // Storage: Scheduler Lookup (r:0 w:1)
     fn on_initialize_periodic_named_resolved(s: u32) -> Weight {
-        (36_758_000 as Weight)
-            // Standard Error: 313_000
-            .saturating_add((48_916_000 as Weight).saturating_mul(s as Weight))
+        (0 as Weight)
```
maltekliemann, line 39: Is this as expected? I checked, and this is non-zero in the `weights.rs` of the scheduler pallet 4.0.0.

runtime/src/weights/pallet_scheduler.rs
```diff
@@ -96,29 +96,29 @@ impl<T: frame_system::Config> pallet_scheduler::weights::WeightInfo for WeightIn
...
             .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(s as Weight)))
             .saturating_add(T::DbWeight::get().writes(2 as Weight))
     }
     // Storage: Scheduler Agenda (r:2 w:2)
     // Storage: Scheduler Lookup (r:0 w:1)
     fn on_initialize_periodic_named(s: u32) -> Weight {
-        (45_414_000 as Weight)
-            // Standard Error: 131_000
-            .saturating_add((22_566_000 as Weight).saturating_mul(s as Weight))
+        (0 as Weight)
```
maltekliemann, line 109: Is this as expected? I checked, and this is non-zero in the weights.rs of the scheduler pallet 4.0.0.

runtime/src/weights/pallet_utility.rs
```diff
@@ -32,22 +32,22 @@ use frame_support::{
...
     fn as_derivative() -> Weight {
-        (3_410_000 as Weight)
+        (4_040_000 as Weight)
     }
     // Storage: unknown [0x3a7472616e73616374696f6e5f6c6576656c3a] (r:1 w:1)
     fn batch_all(c: u32) -> Weight {
-        (25_449_000 as Weight)
-            // Standard Error: 63_000
-            .saturating_add((7_641_000 as Weight).saturating_mul(c as Weight))
+        (0 as Weight)
```
maltekliemann, line 44: Is this as expected? I checked, and this is non-zero in the weights.rs of pallet-utility version 4.0.0-dev.

zrml/swaps/src/weights.rs
```diff
@@ -94,11 +94,11 @@ impl<T: frame_system::Config> WeightInfoZeitgeist for WeightInfo<T> {
     // Storage: Tokens Accounts (r:46 w:22)
     // Storage: System Account (r:2 w:1)
     fn distribute_pool_share_rewards(a: u32, b: u32) -> Weight {
-        (70_035_000 as Weight)
-            // Standard Error: 283_000
-            .saturating_add((45_926_000 as Weight).saturating_mul(a as Weight))
-            // Standard Error: 283_000
-            .saturating_add((67_740_000 as Weight).saturating_mul(b as Weight))
+        (0 as Weight)
```
maltekliemann, line 97: This seems weird, this function wasn't modified; `a` is the number of accounts, so we don't have to worry about getting DDoS'ed, but this is curious.

maltekliemann (changes requested):
Looks good, but I'm a little concerned about the difference between the weights in the substrate repository and ours, so I'm putting "Request changes" for now. Surely there must be a good explanation for this?

zrml/swaps/src/weights.rs
Chralt98: Yeah, nothing changed for `distribute_pool_share_rewards` since the last benchmark commit. The comparison to verify is [here](https://github.com/zeitgeistpm/zeitgeist/compare/ce520be36cc06d6687ca2f5c8499d2deb762d32d...e6ca87795f481728bd208487121a32f9ffb6c0d2?diff=split#diff-188fb32939fe9cc9416edd214b3ab316def04705fad5aa33b9130b8d49220244L938-L1061).

Chralt98 (commented):
This PR updates the benchmark weights.
In general, it could be expected, that the weights of our frame pallets and moonbeams doesn't differ so much, but we have our own reference machine, right?
Great catch @maltekliemann for `distribute_pool_share_rewards` in the swaps pallet. I would also expect that nothing changes for the weight here.

zrml/swaps/src/weights.rs
sea212: This is out of my control, sometimes the benchmarks for the same benchmark logic start with a base weight of greater than zero and sometimes it's zero. I think as long as the weight is modified properly by variables this is acceptable - The weight calculated by the variables often overshadows the base weight by magnitudes anyways.
I would be altered though if the base weight is zero and isn't changed by anything else, like with `system.remark` in this PR.

zrml/prediction-markets/src/weights.rs
```diff
@@ -10,7 +10,7 @@
 // pallet
 // --chain=dev
 // --steps=10
-// --repeat=2
+// --repeat=1000
```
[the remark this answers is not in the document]
sea212: [This is not the case](https://github.com/zeitgeistpm/zeitgeist/blob/v0.3.2/zrml/prediction-markets/src/weights.rs). The weights were changed within a PR and `2` was used because the benchmark workflow does not work properly on forks yet.

runtime/src/weights/pallet_utility.rs
sea212: This is out of my control, sometimes the benchmarks for the same benchmark logic start with a base weight of greater than zero and sometimes it's zero. I think as long as the weight is modified properly by variables this is acceptable. If a change in the logic did not happen we can revert the base weight tough, I think it's better to overcharge than to undercharge (undercharging can lead to overweight blocks that can't be produced in time).

runtime/src/weights/pallet_scheduler.rs
sea212: This is out of my control, sometimes the benchmarks for the same benchmark logic start with a base weight of greater than zero and sometimes it's zero. I think as long as the weight is modified properly by variables this is acceptable. If a change in the logic did not happen we can revert the base weight tough, I think it's better to overcharge than to undercharge (undercharging can lead to overweight blocks that can't be produced in time).

runtime/src/weights/pallet_scheduler.rs
sea212: This is out of my control, sometimes the benchmarks for the same benchmark logic start with a base weight of greater than zero and sometimes it's zero. I think as long as the weight is modified properly by variables this is acceptable. If a change in the logic did not happen we can revert the base weight tough, I think it's better to overcharge than to undercharge (undercharging can lead to overweight blocks that can't be produced in time).

sea212 (commented):
Thanks for looking over the changes. The most spotted peculiarity is that some benchmark weights went to `0`. This is out of my control - The same benchmark logic sometime results in a base weight of `0` and sometimes of non `0`. When it changes to `0` usually the impact of the parameters in weight is increased by the benchmark evaluation logic. As long as the major part of the cost is determined by the parameters, this is no major issue to me.

maltekliemann (approved):
Ok, then we can go ahead and publish! :thumbsup:

Chralt98 (approved):
👍

PR #681, opened 2022-06-20
Clear disputes from storage

vivekvpandya:
Fixes https://github.com/zeitgeistpm/runtime-audit-1/issues/45

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1744,6 +1744,8 @@ mod pallet {
                 m.resolved_outcome = Some(resolved_outcome.clone());
                 Ok(())
             })?;
+            //  Market is resolved so remove all disputes.
+            Disputes::<T>::remove(market_id);
             Self::deposit_event(Event::MarketResolved(
                 *market_id,
                 MarketStatus::Resolved,
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1744,7 +1744,7 @@ mod pallet {
                 m.resolved_outcome = Some(resolved_outcome.clone());
                 Ok(())
             })?;
-            //  Market is resolved so remove all disputes.
+            //  Market is resolved so remove any disputes.
             Disputes::<T>::remove(market_id);
             Self::deposit_event(Event::MarketResolved(
                 *market_id,
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -112,7 +112,7 @@ mod pallet {
     pub const RESERVE_ID: [u8; 8] = PmPalletId::get().0;
 
     /// The current storage version.
-    const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+    const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
 
     pub(crate) type TimeFrame = u64;
     pub(crate) type BalanceOf<T> =
@@ -196,6 +196,7 @@ mod pallet {
             Self::clear_auto_close(&market_id)?;
             Self::clear_auto_resolve(&market_id)?;
             T::MarketCommons::remove_market(&market_id)?;
+            Disputes::<T>::remove(market_id);
 
             Self::deposit_event(Event::MarketDestroyed(market_id));
 
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1744,7 +1744,6 @@ mod pallet {
                 m.resolved_outcome = Some(resolved_outcome.clone());
                 Ok(())
             })?;
-            //  Market is resolved so remove any disputes.
             Disputes::<T>::remove(market_id);
             Self::deposit_event(Event::MarketResolved(
                 *market_id,
--- a/zrml/prediction-markets/src/migrations.rs
+++ b/zrml/prediction-markets/src/migrations.rs
@@ -111,7 +111,6 @@ mod tests {
             setup_chain();
             let _ = Currency::deposit(Asset::Ztg, &ALICE, 1_000 * BASE);
 
-            // Markets which end here will have to be closed on migration:
             let short_time: MomentOf<Runtime> = (5 * MILLISECS_PER_BLOCK).into();
 
             create_test_market(
@@ -126,7 +125,7 @@ mod tests {
                 MarketStatus::Resolved,
             );
 
-            // add one simple dispute for alreay resolved market to simulate a case
+            // Add one simple dispute for alreay resolved market to simulate a case
             // where there is pendig dispute(s) for resolver market and that should
             // be cleaned in storage migration.
             let market_dispute =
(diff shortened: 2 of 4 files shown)


maltekliemann (changes requested):
Good stuff! Thanks for extending the tests, as well. :thumbsup:

Strictly speaking, this change does require a storage migration (`market_id` must be removed from `Disputes` for all resolved markets). I know this isn't absolutely necessary, but I think it would be very prudent to do so. Could you add the storage migration?

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,216 @@
...
+        }
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+        log::info!("Starting storage migration of RemoveDisputesOfResolvedMarkets");
+
+        for (market_id, market) in T::MarketCommons::market_iter() {
+            total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+
+            match market.status {
+                MarketStatus::Resolved => {
+                    Disputes::<T>::remove(market_id);
```
Chralt98:
```suggestion
                    Disputes::<T>::remove(market_id);
                    total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
```

Isn't this a storage write?

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,182 @@
...
+            let disputes = crate::Disputes::<Runtime>::get(&0);
+            assert_eq!(disputes.len(), 1);
+
+            RemoveDisputesOfResolvedMarkets::<Runtime>::on_runtime_upgrade();
+
+            assert_eq!(MarketCommons::market(&0).unwrap().status, MarketStatus::Resolved);
+            let disputes = crate::Disputes::<Runtime>::get(&0);
+            assert_eq!(disputes.len(), 0);
+        });
+    }
```
Chralt98, line 142: Maybe create at least two old markets with at least two disputes for each one of them.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,182 @@
...
+                MarketStatus::Resolved,
+            );
+
+            run_to_block(55);
+
+            // add one simple dispute for alreay resolved market to simulate a case
+            // where there is pendig dispute(s) for resolver market and that should
+            // be cleaned in storage migration.
+            let market_dispute =
+                MarketDispute { at: 55, by: CHARLIE, outcome: OutcomeReport::Categorical(0) };
```
Chralt98, line 128: I think the time is not so important anymore, as long as we see that there are `resolved` markets with multiple disputes.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,176 @@
...
+            return total_weight;
+        }
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+        log::info!("Starting storage migration of RemoveDisputesOfResolvedMarkets");
+
+        for (market_id, market) in T::MarketCommons::market_iter() {
+            total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+
+            if market.status == MarketStatus::Resolved {
+                Disputes::<T>::remove(market_id);
```
Chralt98, line 37:
```suggestion
                Disputes::<T>::remove(market_id);
                total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
```

Isn't this a storage write?

zrml/prediction-markets/src/lib.rs
```diff
@@ -1736,6 +1736,8 @@ mod pallet {
                 m.resolved_outcome = Some(resolved_outcome.clone());
                 Ok(())
             })?;
+            //  Market is resolved so remove any disputes.
```
Chralt98, line 1739:
```suggestion
```

Chralt98 (commented):
This PR removes all `Disputes` from `resolved` markets in the past and will in the future! That looks pretty good.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,216 @@
...
+        let mut total_weight = T::DbWeight::get().reads(1);
+
+        if StorageVersion::get::<Pallet<T>>() != PREDICTION_MARKETS_REQUIRED_STORAGE_VERSION {
+            log::info!(
+                "Skipping storage removal of disputes of resolved markets; prediction-markets \
+                 already up to date"
+            );
+            return total_weight;
+        }
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
```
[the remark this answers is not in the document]
vivekvpandya: yes.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -1 +1,190 @@
...
+        });
+    }
+
+    #[test]
+    fn test_on_runtime_upgrade_with_sample_markets() {
+        ExtBuilder::default().build().execute_with(|| {
+            setup_chain();
+            let _ = Currency::deposit(Asset::Ztg, &ALICE, 1_000 * BASE);
+
+            // Markets which end here will have to be closed on migration:
```
maltekliemann, line 114: [nitpick] You don't really depend on how long these markets run since you set them to `Resolved` anyways.

```suggestion
```

maltekliemann (changes requested):
So far, so good! :thumbsup: I ran the migration using a local copy of Battery Station (you can do this by doing `make try-runtime-upgrade`). Passed. ✅ 

There are two things you still need to change, though:

- The `StorageVersion` or the prediction markets pallet must be set to `2`: https://github.com/zeitgeistpm/zeitgeist/blob/1990143de2b6caba1e85ed8898e5f5525b40510c/zrml/prediction-markets/src/lib.rs#L115 This is so that new copies of the chain already have the correct storage version.
- There are two thing that can happen to a disputed market: It can be resolved (the common case which you covered), or it can be deleted by `admin_destroy_market`. It would be very unusual for this to happen, but it is possible. So you need to modify `admin_destroy_market` to also clear the `Disputes` vector. Make sure to clear the disputes only after `clear_auto_resolve` have been called, because that function still uses the disputes.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -119,25 +120,38 @@ mod tests {
...
 
+            let disputes = crate::Disputes::<Runtime>::get(&1);
+            assert_eq!(disputes.len(), 2);
             RemoveDisputesOfResolvedMarkets::<Runtime>::on_runtime_upgrade();
 
             assert_eq!(MarketCommons::market(&0).unwrap().status, MarketStatus::Resolved);
             let disputes = crate::Disputes::<Runtime>::get(&0);
             assert_eq!(disputes.len(), 0);
+            let disputes = crate::Disputes::<Runtime>::get(&1);
+            assert_eq!(disputes.len(), 0);
```
maltekliemann, line 153: I don't understand the purpose of this change. What's the difference between the two markets? You just seems basically run the test twice.

If anything, it would be interesting to have two markets which resolve at different times and then see that between the resolution of the first and second market only one of them has their disputes deleted.

maltekliemann (approved):
Looks good to me. :thumbsup:

I'm just a bit puzzled by the last addition to the test. Not sure if that's what you actually wanted to do...?

zrml/prediction-markets/src/migrations.rs
Chralt98: The initial thought of [this](https://github.com/zeitgeistpm/zeitgeist/pull/681#discussion_r903560760) was, that the market iterator works properly. But yeah, we remove the migrations file anyways. I am okay with your suggestion.

Chralt98 (approved):
LGTM

PR #684, opened 2022-06-21
Fix block timeout when starting a new chain from genesis

### LINKED ISSUE #683: Since v0.3.3 the standalone client halts after 1 block
The standalone client for the `v0.3.3` release just stops producing blocks after 1 block and increasingly consumes memory until the system runs out of memory. The bug was introduced somewhere between `v0.3.2` and `v0.3.3`

sea212:
closes #683 , the strategy is described in that issue and as comments in the code.

I decided to execute the timestamp check after the block-end-time based markets were processed, although logically the check would make more sense before those markets have been evaluated, since when the timestamp is still `0` we are still in the genesis block and no markets can exist. However, moving this check before the block-end-time based markets were processed will result in a massive amount of tests failing and in that the tester is always responsible to set the timestamp although it is irrelevant in the respective tests.

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1943,8 +1943,10 @@ mod pallet {
             // If we are at genesis the timestamp is 0. No market can exist, we skip the evaluation.
             // Without this check, new chains starting from genesis will hang up, since the loop
             // below will run over an interval of 0 to the current time frame.
+            // We check the block number and the timestamp, since technically the timestamp is
+            // undefined at genesis.
             let current_time_frame = Self::calculate_time_frame_of_moment(T::MarketCommons::now());
-            if current_time_frame == 0u64 {
+            if current_time_frame == 0 || now == T::BlockNumber::zero() {
                 return Ok(());
             }
 
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1930,17 +1930,23 @@ mod pallet {
                 Market<T::AccountId, T::BlockNumber, MomentOf<T>>,
             ) -> DispatchResult,
         {
+            // If we are at genesis the timestamp is 0. No market can exist, we skip the evaluation.
+            // Without this check, new chains starting from genesis will hang up, since the loop
+            // below will run over an interval of 0 to the current time frame.
+            let current_time_frame = Self::calculate_time_frame_of_moment(T::MarketCommons::now());
+            if current_time_frame == 0u64 { return Ok(()) }
+
             for market_id in MarketIdsPerCloseBlock::<T>::get(&now).iter() {
                 let market = T::MarketCommons::market(market_id)?;
                 mutation(market_id, market)?;
             }
-            MarketIdsPerCloseBlock::<T>::remove(&now);
 
-            let current_time_frame = Self::calculate_time_frame_of_moment(T::MarketCommons::now());
+            MarketIdsPerCloseBlock::<T>::remove(&now);
             // On first pass, we use current_time - 1 to ensure that the chain doesn't try to check
             // all time frames since epoch.
             let last_time_frame =
                 LastTimeFrame::<T>::get().unwrap_or_else(|| current_time_frame.saturating_sub(1));
+
             for time_frame in last_time_frame.saturating_add(1)..=current_time_frame {
                 for market_id in MarketIdsPerCloseTimeFrame::<T>::get(&time_frame).iter() {
                     let market = T::MarketCommons::market(market_id)?;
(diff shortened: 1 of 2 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -1934,20 +1937,30 @@ mod pallet {
...
+
             MarketIdsPerCloseBlock::<T>::remove(&now);
 
+            // If we are at genesis the timestamp is 0. No market can exist, we skip the evaluation.
+            // Without this check, new chains starting from genesis will hang up, since the loop
+            // below will run over an interval of 0 to the current time frame.
             let current_time_frame = Self::calculate_time_frame_of_moment(T::MarketCommons::now());
+            if current_time_frame == 0u64 {
+                return Ok(());
+            }
```
maltekliemann, line 1949: The API of `pallet-timestamp` states (https://github.com/paritytech/substrate/blob/aacc3e99c0750c3e6c851a8270dc2c4eebca3290/frame/timestamp/src/lib.rs#L297-L299):

> On genesis the time returned is not valid. 

So strictly speaking, this check relies on undefined behavior. Unless there's an argument against this, I think you should just check if the current block number is zero instead.

maltekliemann (changes requested):
Thanks for fixing my mistake! 😬

I'm putting _Request changes_. May seem like a nitpick, but I think we should at least discuss my suggestion.

Other than that, I wonder if there's a way to directly test that this bug won't occur again. For example: Set timestamp to zero, run to block 1, then set timestamp to 1,655,837,424,017 (that's right now) and then see if it returns immediately. If it takes too long, there's something broken. Sounds like a questionable test, but I thought I'd bring it up.

zrml/prediction-markets/src/lib.rs
sea212: That's a great suggestion!

sea212:
Had to commit one more fix for the test [d1503b1](https://github.com/zeitgeistpm/zeitgeist/pull/684/commits/d1503b154b660444ae298089b0efc580a344e929). Previously the storage was modified at `blocknumber = 0`, because `run_to_block(1)` executed `on_initialize` which in turn eventually jumped in `market_close_manager` with `blocknumber = 1`.

zrml/prediction-markets/src/tests.rs
```diff
@@ -916,19 +916,24 @@ fn market_close_manager_skips_the_genesis_block_with_timestamp_zero() {
...
         // Blocknumer != 0 and timestamp != 0
-        run_to_block(1);
-        assert_ok!(PredictionMarkets::market_close_manager(0, noop_mutation));
+        assert_ok!(PredictionMarkets::market_close_manager(1, noop_mutation));
         assert_eq!(
             LastTimeFrame::<Runtime>::get(),
             Some(Timestamp::now() / crate::TimeFrame::from(MILLISECS_PER_BLOCK))
         );
         assert!(LastTimeFrame::<Runtime>::get() != Some(0));
+        println!("{:?}", LastTimeFrame::<Runtime>::get())
```
maltekliemann, line 936:
```suggestion
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -887,6 +887,55 @@ fn on_market_close_successfully_auto_closes_multiple_markets_after_stall() {
...
+        assert_ok!(PredictionMarkets::market_close_manager(1, noop_mutation));
+        assert_eq!(LastTimeFrame::<Runtime>::get(), None);
+
+        // Blocknumer = 0 and timestamp != 0
+        Timestamp::set_timestamp(end);
+        assert_ok!(PredictionMarkets::market_close_manager(0, noop_mutation));
+        assert_eq!(LastTimeFrame::<Runtime>::get(), None);
+
+        // Blocknumer != 0 and timestamp != 0
+        assert_ok!(PredictionMarkets::market_close_manager(1, noop_mutation));
```
vivekvpandya, line 930: How is this check relevant? IIUC this returns Ok(()) from noop_mutation and all other cases above are returned from newly added condition check in market_close_manager right ?

sea212: The first three checks ensure that no timestamp is stored at genesis or when it's zero. The blocknumber is also checked because the timestamp technically is undefined at genesis. The last check verifies that using a correct block number and timestamp results in storing the proper timestamp. The idea behind this is that a zero or undefined timestamp will never be stored within the storage `LastTimeFrame`, effectively avoiding block production stalling ranges from a zero/undefined timestamp to the next current valid timestamp within the range in: https://github.com/zeitgeistpm/zeitgeist/blob/6c0b74dc7078b51fd1426023a5174fb8058ee132/zrml/prediction-markets/src/lib.rs#L1946-L1955

maltekliemann (approved):
:thumbsup:

vivekvpandya (approved):

PR #687, opened 2022-06-27
Fix weights to fix CI on `main`

maltekliemann:
c0f268c failed due to an outdated CI run of #672, which led us to overlook that merging #672 would cause CI to fail on `main` due to a test introduced in #684.

This PR fixed that test.


Chralt98 (approved):
LGTM, I checked it on my local workstation.

PR #693, opened 2022-06-29
Move unreserve `ValidityBond` in `on_resolution` after market state check

Chralt98:
I found it strange, that the `MarketStatus` was checked for reports, but not for the disputes [here](https://github.com/zeitgeistpm/zeitgeist/blob/b70a5c4517dea683ee1018aebcf73fc796937870/zrml/prediction-markets/src/lib.rs#L1997-L2008). I had the fear, that multiple `ValidityBonds` of Alice could have been unreserved [here](https://github.com/zeitgeistpm/zeitgeist/blob/b70a5c4517dea683ee1018aebcf73fc796937870/zrml/prediction-markets/src/lib.rs#L1621-L1627). I made the test and realised, that the `MarketIdsPerDisputeBlock` removed all old disputes for one market id [here](https://github.com/zeitgeistpm/zeitgeist/blob/b70a5c4517dea683ee1018aebcf73fc796937870/zrml/prediction-markets/src/lib.rs#L369) (good job @maltekliemann). That's why the test I made didn't fail. But just in case for future modifications, I like the test and a additional market state check before calling `on_resolution`.


Chralt98:
Shouldn't [this](https://github.com/zeitgeistpm/zeitgeist/blob/b70a5c4517dea683ee1018aebcf73fc796937870/zrml/prediction-markets/src/lib.rs#L1621-L1627) get unreserved after the market status check?

maltekliemann (changes requested):
The code checks the market state when resolving a market when it hits its `MarketIdsPerReportBlock` entry because despite being possibly disputed now, the market's id is still contained in that map. This type of check is not required when resolving a market when it hits its `MarketIdsPerDisputeBlock`.

I don't exactly mind the check that you've added, but it would be much more resonable to have the `dispute` dispatch just delete the entry in `MarketIdsPerReportBlock`.

Other than that, I feel like the test lacks focus. You create three markets, then only interact with the first one. What's the role of the other two markets? And why are you disputing the market multiple times? And what's the purpose of re-testing how the bonds are unreserved/slashed?

Chralt98:
So all in all, I don't mind if we integrate this test or not, because it doesn't fail, but could potentially detect a future bug.

maltekliemann:
> Could the ValidityBond of the market creator get unreserved multiple times (as not intended)?

This would happen if `on_resolution` was called multiple times. I wasn't around when `on_resolution` was implemented, but I'd say that it's part of the API that the caller needs to make sure that `on_resolution` used responsibly, so to speak. In other words, it **should** only be called once, and it is not the job of `on_resolution` to check for misuse.

> That's why I created three markets.

You want to test that when a disputed market resolves it doesn't unreserve more than it should, right? In that case, I would just use a sentinel amount.

Regarding 6ce2e60ad85a3a1bb83e7ea6c0c439b9a872f9bd, sure, we can move the unreserving. But if `on_resolution` assumes from the start that the market is ready to be resolved, there's no problem anyways. Also note that `on_resolution` actually does error when the market status is incorrect, and in both applications (inside `admin_market_to_resolved` and `on_initialize`) a rollback is performed, so the unreserving of fees actually never happens.

Chralt98:
The last point makes so much sense. Thank you for your deep knowledge share.

PR #694, opened 2022-06-30
Rename Market struct's mdm to dispute_mechanism.

vivekvpandya:
This fixes https://github.com/zeitgeistpm/runtime-audit-1/issues/30

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -355,7 +355,7 @@ mod pallet {
                 &who,
                 default_dispute_bond::<T>(disputes.len()),
             )?;
-            match market.mdm {
+            match market.dispute_mechanism {
                 MarketDisputeMechanism::Authorized(_) => {
                     T::Authorized::on_dispute(&disputes, &market_id, &market)?
                 }
@@ -396,7 +396,7 @@ mod pallet {
         /// * `period`: The active period of the market.
         /// * `metadata`: A hash pointer to the metadata of the market.
         /// * `market_type`: The type of the market.
-        /// * `mdm`: The market dispute mechanism.
+        /// * `dispute_mechanism`: The market dispute mechanism.
         /// * `amount`: The amount of each token to add to the pool.
         /// * `weights`: The relative denormalized weight of each asset price.
         #[pallet::weight(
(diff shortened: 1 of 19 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -418,7 +418,7 @@ mod pallet {
             period: MarketPeriod<T::BlockNumber, MomentOf<T>>,
             metadata: MultiHash,
             market_type: MarketType,
-            mdm: MarketDisputeMechanism<T::AccountId>,
+            dispute_mechanism: MarketDisputeMechanism<T::AccountId>,
```
Chralt98, line 421: Because we changed the name of the extrinsic parameter here, we should probably tell the front-end about that change in the [changelog for devs](https://github.com/zeitgeistpm/zeitgeist/blob/main/docs/changelog_for_devs.md). https://github.com/zeitgeistpm/ui/search?q=mdm.

zrml/prediction-markets/src/lib.rs
```diff
@@ -469,7 +469,7 @@ mod pallet {
             metadata: MultiHash,
             creation: MarketCreation,
             market_type: MarketType,
-            mdm: MarketDisputeMechanism<T::AccountId>,
+            dispute_mechanism: MarketDisputeMechanism<T::AccountId>,
```
Chralt98, line 472: Because we changed the name of the extrinsic parameter here, we should probably tell the front-end about that change in the [changelog for devs](https://github.com/zeitgeistpm/zeitgeist/blob/main/docs/changelog_for_devs.md). https://github.com/zeitgeistpm/ui/search?q=mdm.

Chralt98 (commented):
LGTM and thank you for improving the code readability. I think with the change there is a small notice for the front end team missing, because you changed the name of an extrinsic parameter. Other then [here](https://github.com/zeitgeistpm/zeitgeist/blob/b70a5c4517dea683ee1018aebcf73fc796937870/zrml/prediction-markets/src/lib.rs#L1657-L1658) I have nothing found which could be renamed. So good job!

maltekliemann (approved):
Looks good to me.

Also note that a migration is not required when renaming a field of a struct (https://docs.substrate.io/reference/scale-codec/):

> For structures, the values are named, but that is irrelevant for the encoding (names are ignored - only order matters).

Chralt98 (approved):
Pretty nice.

PR #701, opened 2022-07-07
Fix zero division error in `on_resolution`

maltekliemann:
When a dispute is started, but the dispute mechanisms `on_resolution` method returns `None` when it is called at market resolution, then the market defaults to the original report of the oracle:

https://github.com/zeitgeistpm/zeitgeist/blob/815d92af2c8d542147cb8bbe58433e820b2be85d/zrml/prediction-markets/src/lib.rs#L1657-L1673

For example, this is the case if `dispute_mechanism` is `Authorized` and the authorized account doens't set an outcome.

The problem is that this results in the number of correct reporters being zero (recall that the oracle doesn't count as correct reported), resulting in a zero division when the rewards are distributed:

https://github.com/zeitgeistpm/zeitgeist/blob/815d92af2c8d542147cb8bbe58433e820b2be85d/zrml/prediction-markets/src/lib.rs#L1712-L1721

This PR fixes that and adds a test which verifies correct behavior.


zrml/prediction-markets/src/tests.rs
```diff
@@ -2222,6 +2222,60 @@ fn authorized_correctly_resolves_disputed_market() {
...
+        ));
+        let market = MarketCommons::market(&0).unwrap();
+        assert_eq!(market.status, MarketStatus::Disputed);
+
+        let charlie_reserved = Balances::reserved_balance(&CHARLIE);
+        assert_eq!(charlie_reserved, DisputeBond::get());
+
+        run_to_block(<frame_system::Pallet<Runtime>>::block_number() + DisputePeriod::get());
+        let market_after = MarketCommons::market(&0).unwrap();
+        assert_eq!(market_after.status, MarketStatus::Resolved);
```
Chralt98, line 2261: The important check here. Without your change the market status wouldn't be resolved, because the `on_resolution` function returned an ArithmeticError in this case (correct reporters vector is equal to an empty vector, because there was no dispute for the correct outcome). Then brilliant 👍 .

Chralt98 (commented):
Great catch @maltekliemann. I lean towards just removing the possibility of the error overall.

Chralt98 (approved):
Such an important catch!

Chralt98 (approved):
Merge LGTM.

PR #702, opened 2022-07-07
Add feature to remove sudo

sea212:
This PR adds the `without-sudo` feature to remove the sudo pallet from runtime and any initialization from genesis.
I would like to keep sudo as a default feature, because that way anybody can build a node from source without any arguments and have sudo in the standalone chain available.
I decided to use a `without-sudo` feature instead of `with-sudo`, because by default sudo should be enabled. One approach would be to add `with-sudo` to the `default` feature, but in that case disabling it would require disabling all other default features as well afaik `--no-default-features`, which seemed not ideal to me.

### DIFF
--- a/node/src/chain_spec/mod.rs
+++ b/node/src/chain_spec/mod.rs
@@ -303,8 +303,8 @@ fn endowed_accounts_staging_mainnet() -> Vec<EndowedAccountWithBalance> {
             hex!["b449a256f73e59602eb742071a07e4d94aaae91e6872f28e161f34982a0bfc0d"].into(),
             DEFAULT_COLLATOR_BALANCE_MAINNET.unwrap(),
         ),
-        #[cfg(not(feature = "without-sudo"))]
         // dE2nxuZc5e7xBbU1cGikmtVGws9niNPUayigoDdyqB7hzHQ6X
+        #[cfg(not(feature = "without-sudo"))]
         EndowedAccountWithBalance(
             hex!["203ef582312dae988433920791ce584daeca819a76d000175dc6d7d1a0fb1413"].into(),
             DEFAULT_SUDO_BALANCE_MAINNET,


vivekvpandya (approved):
Looks good to me! Is there any relevant bug to understand the context?

runtime/Cargo.toml
```diff
@@ -20,7 +20,7 @@ pallet-preimage = { branch = "moonbeam-polkadot-v0.9.19", default-features = fal
 pallet-proxy = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/substrate" }
 pallet-randomness-collective-flip = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/substrate" }
 pallet-scheduler = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/substrate" }
-pallet-sudo = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/substrate" }
+pallet-sudo = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, optional = true, git = "https://github.com/purestake/substrate" }
```
Chralt98, line 23: Because of this, shall we add a `?` optional [here](https://github.com/zeitgeistpm/zeitgeist/blob/f3db56ff8945ffef19e696bc8d6c026b753c2a39/runtime/Cargo.toml#L306)? According to https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

runtime/src/lib.rs
```diff
@@ -221,35 +220,54 @@ macro_rules! create_zeitgeist_runtime {
...
+}
+
+#[cfg(not(feature = "without-sudo"))]
+create_zeitgeist_runtime_with_additional_pallets!(
+    // Others
+    Sudo: pallet_sudo::{Call, Config<T>, Event<T>, Pallet, Storage} = 150,
 );
 
+#[cfg(feature = "without-sudo")]
+create_zeitgeist_runtime_with_additional_pallets!();
```
Chralt98, line 269: Why is this required here?

runtime/src/lib.rs
```diff
@@ -364,12 +383,15 @@ cfg_if::cfg_if! {
                     | Call::Utility(_)
                     | Call::Vesting(_) => true,
 
+                    #[cfg(not(feature = "without-sudo"))]
+                    Call::Sudo(_) => true,
```
Chralt98, line 387: Maybe depend on the compiler flag check feature [here](https://github.com/zeitgeistpm/zeitgeist/blob/f3db56ff8945ffef19e696bc8d6c026b753c2a39/runtime/src/lib.rs#L419) too.

Chralt98 (changes requested):
Ok, checked it. There are some possible flaws. But all in all good job to go in the direction of full decentralisation.

node/src/chain_spec/dev.rs
```diff
@@ -54,6 +54,7 @@ pub fn dev_config(
                 .into_iter()
                 .map(|acc| EndowedAccountWithBalance(acc, INITIAL_BALANCE))
                 .collect(),
+                #[cfg(not(feature = "without-sudo"))]
```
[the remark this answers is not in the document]
Chralt98: Oh my fault, because of this
Bildschirmfoto 2022-07-08 um 09 18 26

node/src/chain_spec/mod.rs
```diff
@@ -300,6 +303,7 @@ fn endowed_accounts_staging_mainnet() -> Vec<EndowedAccountWithBalance> {
             hex!["b449a256f73e59602eb742071a07e4d94aaae91e6872f28e161f34982a0bfc0d"].into(),
             DEFAULT_COLLATOR_BALANCE_MAINNET.unwrap(),
         ),
+        #[cfg(not(feature = "without-sudo"))]
         // dE2nxuZc5e7xBbU1cGikmtVGws9niNPUayigoDdyqB7hzHQ6X
```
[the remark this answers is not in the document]
sea212: What is the difference? Why prefer one over the other?

node/src/chain_spec/dev.rs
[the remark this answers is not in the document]
sea212: This is not the case, the next line after the feature is `get_account_id_from_seed::<sr25519::Public>("Alice"),`, which defines `Alice` as the superuser account.

runtime/Cargo.toml
sea212: Good point, this was the initial idea why I made it optional. In the first iteration I used `pallet-sudo` and then `with-sudo` features. Turns out one cannot use the inherent `pallet-sudo` feature with the conditional inclusion `?`, and when using `with-sudo` one can use the `?` operator, but since `with-sudo` would be the default option, disabling it would require the rather unintuitive and too powerful flag `--no-default-features`. When using `without-sudo`, the `?` can't be used because `pallet-sudo` is used by default and only disabled with the feature.
So the way to go is to remove the optional parameter.

```suggestion
pallet-sudo = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/substrate" }
```

runtime/src/lib.rs
sea212: I found no way to include that feature requirement inside a macro, so I had to create a new wrapper macro and apply the feature requirement on the whole wrapper macro. Let me know if you know a better approach to this, I already wasted like 20-30 minutes figuring out a nicer approach to that.

runtime/Cargo.toml
Chralt98: Thank you!

runtime/src/lib.rs
sea212: We don't need it in the last case, because when sudo is included, it falls into the `_ => ` match arm and calls won't be filtered. When it's not included it is not regarded at all.

node/src/chain_spec/battery_station.rs
```diff
@@ -20,6 +22,7 @@ pub fn battery_station_staging_config(
                     parachain_id,
                 ),
                 endowed_accounts_staging_testnet(),
+                #[cfg(not(feature = "without-sudo"))]
                 root_key_staging_testnet(),
                 zeitgeist_wasm,
             )
```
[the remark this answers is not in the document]
sea212: Yes, this is correct.

node/src/chain_spec/mod.rs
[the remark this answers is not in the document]
sea212: Technically it does not make a difference (yes, comment are ignored), but the consistency argument is solid :+1:

runtime/src/lib.rs
Chralt98: I understand, you mean because of not including the whole `Sudo` pallet in the `construct_runtime`. That makes sense.

Chralt98 (approved):
LGTM. Let us go in the direction of full decentralisation together!

runtime/src/lib.rs
Chralt98: Sorry, I am out here. Maybe you @maltekliemann or @vivekvpandya ?

vivekvpandya: I also don't have anything better than what @sea212  has already done https://stackoverflow.com/questions/63008781/check-a-feature-flag-in-an-item-defined-in-a-macro

PR #704, opened 2022-07-08
Bump versions to v0.3.4

sea212:


maltekliemann (changes requested):
I believe we also need to increment `spec_version` and `transaction_version`. Note sure about the last one, since all we did was change the _name_ of a parameter.

sea212:
> I believe we also need to increment `spec_version` and `transaction_version`. Note sure about the last one, since all we did was change the _name_ of a parameter.

You are right. It was not very transparent, but this change already happened with the following commit: https://github.com/zeitgeistpm/zeitgeist/commit/723171517d5c7ea94b9bb22ea9289c2dd8d7fc5f

maltekliemann:
> Hmm guess it makes sense to do this with the version bump in the future for more transparency.

Agreed. I guess we could use the changelog for devs to check for changes to extrinsics?

maltekliemann (approved):
Looks good to me! :thumbsup:

sea212:
> > Hmm guess it makes sense to do this with the version bump in the future for more transparency.
> 
> Agreed. I guess we could use the changelog for devs to check for changes to extrinsics?

I think this does only cover some cases. For example, assigning different module indices to the pallets results in a bump in transaction version as well. When a new parameter, dispatchable, pallet, etc. is mentioned in the changelog we can use that as an indicator to bump the version, but no information of that kind in the changelog mustn't be interpreted as "no bump required".

PR #706, opened 2022-07-08
Expose AuthorFilter Call enum

### LINKED ISSUE #697: [Network] Number of forks and reorgs is too high
It seems there is a problem during the author selection, since most of the time most of the collators try to produce a block simultaneously, which leads to an unacceptable high number of forks and reorgs.

sea212:
Closes #697 

We have to set the `eligibleCount` of the `author_slot_filter` pallet to `1` to give every collator even chances of collation and to significantly reduce the number of forks and reorgs. To do so, root has to call a dispatchable of the `author_slot_filter` pallet. Unfortunately the `Call` enum of the `author_slot_filter` pallet is not exposed in the runtime, rendering setting this value through governance impossible.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -232,7 +232,7 @@ macro_rules! create_zeitgeist_runtime_with_additional_pallets {
             // Consensus
             ParachainStaking: parachain_staking::{Call, Config<T>, Event<T>, Pallet, Storage} = 110,
             AuthorInherent: pallet_author_inherent::{Call, Inherent, Pallet, Storage} = 111,
-            AuthorFilter: pallet_author_slot_filter::{Config, Event, Pallet, Storage} = 112,
+            AuthorFilter: pallet_author_slot_filter::{Config, Call, Event, Pallet, Storage} = 112,
             AuthorMapping: pallet_author_mapping::{Call, Config<T>, Event<T>, Pallet, Storage} = 113,
 
             // XCM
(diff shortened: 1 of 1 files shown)


Chralt98 (approved):
You found a temporary fix of disabling the cache, preventing the storage root mismatch because of the known bug. So my first question was, if this PR is still required then. I could imagine, that we would like to use the extrinsic of the `pallet_author_slot_filter` in the future. So I would lean towards approving.

It might be, that I don't get the full context now, but wouldn't setting the `eligibleCount` to `1` to only allow one collator to produce blocks (per active slot)? I could imagine, that this would lead to allowing one malicious lucky collator to rule the network with invalid blocks.

sea212:
> You found a temporary fix of disabling the cache, preventing the storage root mismatch because of the known bug. So my first question was, if this PR is still required then. I could imagine, that we would like to use the extrinsic of the `pallet_author_slot_filter` in the future. So I would lean towards approving.
> 
> It might be, that I don't get the full context now, but wouldn't setting the `eligibleCount` to `1` to only allow [one collator to produce blocks (per active slot)](https://purestake.github.io/nimbus/src/pallet_author_slot_filter/lib.rs.html#93-116)? I could imagine, that this would lead to allowing one malicious lucky collator to rule the network with manipulated blocks.

Just realized I referenced the wrong issue in this PR. Does https://github.com/zeitgeistpm/zeitgeist/issues/697#issuecomment-1179732036 answer your question?

sea212:
In addition to that, collators cannot craft blocks as they like, they have to adhere to the consensus rules, other nodes will reject invalid state transitions that don't adhere to the consensus rules. A collator can control which transactions are executed though. If the complete active collator set is occupied by a malicious entity, it can censor transactions (just never include them) or even halt the network.

Chralt98:
Thanks for your explanation. But why does the `pallet_author_slot_filter` set the default value of `eligibleCount` to `50`? I mean, the number of forks and reorgs in other parachains should be also very high then. As long as the other collators validate the eligible collator which produces a new block, setting `eligibleCount` to `1` sounds reasonable.

maltekliemann (commented):
:thumbsup:

PR #707, opened 2022-07-12
Use CountedStorageMap instead of StorageMap for Jurors in court pallet

vivekvpandya:
Fixes https://github.com/zeitgeistpm/runtime-audit-1/issues/3


Chralt98 (approved):
According to [this](https://github.com/paritytech/substrate/issues/8605) the `CountedStorageMap` allows to `count()` without going throughout the full iterator. This is a performance optimisation. There speaks nothing against this change to me. Cool @vivekvpandya !

vivekvpandya:
> According to [this](https://github.com/paritytech/substrate/issues/8605) the `CountedStorageMap` allows to `count()` without going throughout the full iterator. This is a performance optimisation. There speaks nothing against this change to me. Cool @vivekvpandya !

Actually it must be using a small variable (storage) to track the count. But I think it should be fine given runtime benefit. 
Also this is suggestion from audit people, so thanks to them!

zrml/court/src/migrations.rs
```diff
@@ -0,0 +1,103 @@
...
+        let mut total_weight = T::DbWeight::get().reads(1);
+
+        if StorageVersion::get::<Pallet<T>>() != COURT_REQUIRED_STORAGE_VERSION {
+            log::info!("Skipping storage migration of jurors of court already up to date");
+            return total_weight;
+        }
+        log::info!("Starting storage migration of jurors of court");
+        Jurors::<T>::initialize_counter();
+        total_weight =
+            total_weight.saturating_add(T::DbWeight::get().writes(Jurors::<T>::count() as u64));
```
maltekliemann, line 28: Plus one for writing the count.

```suggestion
        total_weight = total_weight.saturating_add(                            
            T::DbWeight::get().writes(Jurors::<T>::count().saturating_add(1) as u64),                                                                         
        );
```

zrml/court/src/migrations.rs
```diff
@@ -0,0 +1,103 @@
...
+    dispatch::Weight,
+    log,
+    pallet_prelude::PhantomData,
+    traits::{Get, OnRuntimeUpgrade, StorageVersion},
+};
+
+const COURT_REQUIRED_STORAGE_VERSION: u16 = 0;
+const COURT_NEXT_STORAGE_VERSION: u16 = 1;
+
+pub struct JurorsConuntedStorageMapMigration<T>(PhantomData<T>);
```
maltekliemann, line 12: Just a typo

```suggestion
pub struct JurorsCountedStorageMapMigration<T>(PhantomData<T>);
```

zrml/court/src/migrations.rs
```diff
@@ -0,0 +1,103 @@
...
+            frame_support::migration::put_storage_value(
+                b"Court",
+                b"Jurors",
+                &alice_hash,
+                alice_juror,
+            );
+            let juror: Option<Juror> =
+                frame_support::migration::get_storage_value(b"Court", b"Jurors", &alice_hash);
+            assert!(juror.is_some());
+            assert!(Jurors::<Runtime>::get(&ALICE).is_some());
```
maltekliemann, line 93: You should probably put this _after_ running the runtime upgrade. Otherwise, you're testing that `put_storage_value` and `get_storage_value` interact correctly with one another. But what you want to check is that after the runtime upgrade, the map entry is still correct, right?

maltekliemann: Also you should check that the result of `get` is actually equal to `alice_juror`.

maltekliemann (changes requested):

zrml/court/src/migrations.rs
maltekliemann: Okay, but you don't need to test that the `migrations` module works. We need to operate on the assumption that it does.

Also, I guess then we need a test that the item is still in storage after the migration.

maltekliemann (approved):
Thanks for the changes! Looks good to me. :thumbsup:

PR #711, opened 2022-07-16
Let markets open their pools when they start

### LINKED ISSUE #691: Markets open even if `period.start` lies in the future
Trading on a market whose `period.start` lies in the future is possible as soon as the market is created.

The solution for this is to implement something similar to `market_close_manager`, using time frames for handling markets with `Timestamp` based period.

maltekliemann:
This is a pull request very similar to #651, but we're dealing with opening markets instead of closing them. It is based on #695, which implements opening of pools. It fixes #691.

A couple of quick points:

- When a market reaches it's starting point (either block or time frame) _and_ it has a CPMM pool, then that pool is opened. Opening a market currently has no other effect. In particular, I've not introduced a `MarketOpen` event, etc.
- It's debatable whether or not markets should also get an `Initialized` state. I don't think so, as it is currently no different from `Active`.
- I took another look into the chain-stalling bug. It seems that the timestamp is set sometime during block 1 using an inherent extrinsic, so during `on_initialize` of block 1, the timestamp is still not set. So I'm now checking for for block number less or equal to `<=1`. This led to a lot of reorganizing in the tests.
- This PR also introduces a compensation for `now()` lagging one block behind during `on_initialize`. This led to more reorganizing in the tests.


zrml/prediction-markets/src/lib.rs
```diff
@@ -1191,19 +1220,57 @@ mod pallet {
...
-                    Ok(())
-                });
+                // If we are at genesis or the first block the timestamp is be undefined. No
+                // market needs to be opened or closed on blocks #0 or #1, so we skip the
+                // evaluation. Without this check, new chains starting from genesis will hang up,
+                // since the loops in `market_open_manager` and `market_close_manager` below will
+                // run over an interval of 0 to the current time frame.
+                if now <= 1u32.into() {
+                    return TransactionOutcome::Commit(Ok(()));
+                }
```
sea212, line 1230: Not really a noteworthy difference but we can move this out of the transactional body.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1232,6 +1299,24 @@ mod pallet {
         ValueQuery,
     >;
 
+    #[pallet::storage]
+    pub type MarketIdsPerOpenBlock<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        T::BlockNumber,
+        BoundedVec<MarketIdOf<T>, ConstU32<1024>>,
```
sea212, line 1307: I think we can lower boundary here a bit because:
1. 1024 markets that close within a 6 second period is too high in my opinion
2. To reduce the effort required to state transit all the markets

If there should be the case that the boundary is surpassed, the user just defines the market to close 6 seconds earlier or later, effectively handling load balancing for the chain.

```suggestion
        BoundedVec<MarketIdOf<T>, ConstU32<32>>,
```

sea212: If accepted, we should also change it for the closing caches.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1232,6 +1299,24 @@ mod pallet {
...
+        ValueQuery,
+    >;
+
+    #[pallet::storage]
+    pub type MarketIdsPerOpenTimeFrame<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        TimeFrame,
+        BoundedVec<MarketIdOf<T>, ConstU32<1024>>,
```
sea212, line 1316:
```suggestion
        BoundedVec<MarketIdOf<T>, ConstU32<64>>,
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1644,6 +1766,16 @@ mod pallet {
...
+        fn on_market_open(
+            market_id: &MarketIdOf<T>,
+            market: Market<T::AccountId, T::BlockNumber, MomentOf<T>>,
+        ) -> Result<Weight, DispatchError> {
+            match market.status {
+                MarketStatus::Active => Self::open_market(market_id),
+                _ => Err(Error::<T>::InvalidMarketStatus.into()), // Should never occur!
+            }
+        }
+
```
sea212, line 1778: Why are `on_market_open` and `open_market` two separate functions? All that `on_market_open` does is check if it is a valid state transition and call `open_market`, but `open_market` should actually also adhere to the state transition rules, shouldn't it?
Is the idea to adhere to the current conventions?

zrml/prediction-markets/src/lib.rs
```diff
@@ -2016,8 +2148,10 @@ mod pallet {
             Ok(())
         }
 
-        pub(crate) fn market_close_manager<F>(
-            now: T::BlockNumber,
+        pub(crate) fn market_open_manager<F>(
+            block_number: T::BlockNumber,
+            last_time_frame: TimeFrame,
+            current_time_frame: TimeFrame,
```
sea212, line 2154: The `market_close_manager` and `market_open_manager` only differ in the storage they use. I propose to create a common functions that takes the storage as an additional parameter and then use that instead, effectively deduplicating the code.

zrml/prediction-markets/src/tests.rs
```diff
@@ -236,18 +267,19 @@ fn admin_destroy_market_correctly_slashes_permissionless_market_disputed() {
...
         );
-        run_to_block(2);
+        run_to_block(end);
         assert_ok!(PredictionMarkets::report(
             Origin::signed(BOB),
             0,
             OutcomeReport::Categorical(1)
         ));
-        run_to_block(9000); // Wait until market resolves
+        run_blocks(<Runtime as Config>::DisputePeriod::get());
```
sea212, line 282: Great, I like the unambiguousness!

sea212:
Didn't see the state change of this PR, was already deep in the review :face_exhaling:

maltekliemann:
Yeah, sorry about that. Forgot the storage migration. :(

zrml/prediction-markets/src/lib.rs
maltekliemann: Yeah, great suggestion. I'll get on that :thumbsup:

zrml/prediction-markets/src/lib.rs
maltekliemann: Yes, other than conformity, I have no good reason for keeping it this way. I have zero conviction about keeping this in, so if you think it's better to remove this layer of abstraction, I can go ahead and get rid of it.

zrml/prediction-markets/src/lib.rs
maltekliemann: Yeah, I agree.

zrml/prediction-markets/src/lib.rs
maltekliemann: :thumbsup:

maltekliemann: Mitigated in e1168926476897ba86cc50dadd536a538b10d3d9

zrml/prediction-markets/src/lib.rs
sea212: I would merge them, because using `open_market` on a market in an invalid state should not be possible. I also wonder why `open_market` does not actually open the market, i.e. make it active (in case of a permissionless market). Where is the state changed to `MarketStatus::Active`?

maltekliemann: Good question. There is no new `MarketStatus`. Even when the market is not yet open, it is active. The only thing that changes when the market opens is that status of the pool.

If you like, I can add an additional variant to `MarketStatus`. The reason that I chose _not_ to do this was that this would be purely informational. There is no technical need for this status in `prediction-markets`.

zrml/prediction-markets/src/lib.rs
maltekliemann: (Mitigated in fc0555ed1d6d1c0f8c0672a47e8a96372a7db64e)

zrml/prediction-markets/src/lib.rs
maltekliemann: Mitigated in fc0555ed1d6d1c0f8c0672a47e8a96372a7db64e

zrml/prediction-markets/src/tests.rs
maltekliemann: :thumbsup:

zrml/prediction-markets/src/lib.rs
sea212: Okay, let's not introduce the state.

zrml/prediction-markets/src/lib.rs
sea212: You are right, there is not a single case on Battery Station where this occurs, so we can leave out the storage migration (since nobody else uses this pallet).

maltekliemann: Caches for closing fixed in 9644a79948f4152a9544b27497a0103c483fe032

zrml/prediction-markets/src/lib.rs
maltekliemann: Mitigated in b38035681aa03a3fc9533fb1438a450739879744

zrml/prediction-markets/src/lib.rs
maltekliemann: `on_market_open` removed in ebbe90685bb241c8d7a1e7d882f4552ae5766d59

zrml/prediction-markets/src/lib.rs
maltekliemann: Also reduced the size of dispute/report caches in 079720bdafca679b84d898e796a234bb96186242

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,331 @@
...
+                        }
+                        MarketPeriod::Timestamp(ref range) => {
+                            let open_time_frame =
+                                Pallet::<T>::calculate_time_frame_of_moment(range.start);
+                            let _ = MarketIdsPerOpenTimeFrame::<T>::try_mutate(
+                                &open_time_frame,
+                                |ids| ids.try_push(market_id),
+                            );
+                        }
+                    }
```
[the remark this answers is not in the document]
Chralt98: Maybe a read and write here?

Chralt98 (commented):
This PR allows to activate markets automatically in the future. Impactful 👍 !

zrml/prediction-markets/src/lib.rs
```diff
@@ -203,6 +204,7 @@ mod pallet {
                 T::MarketCommons::remove_market_pool(&market_id)?;
             }
 
+            Self::clear_auto_open(&market_id)?;
             Self::clear_auto_close(&market_id)?;
```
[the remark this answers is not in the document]
maltekliemann: Yeah, great suggestion.

I tried this, and unfortunately, I ran into some trouble. You'll probably want to keep `clear_auto_close` and `clear_auto_open`, but let them call a function `clear_auto_cache` which is structured similarly to the `market_status_manager`. Otherwise, you'd have to keep writing `clear_auto_cache::<MarketIdsPerOpenBlock, MarketIdsPerOpenTimestamp>` where `clear_auto_close` used to be.

But then you must also specify whether or not to access `range.start` or `range.end` in `clear_auto_cache`. I don't think this has a good solution. You can either use a macro, or some kind of a convoluted closure construction. Either way, you end up with 30 lines of complex code to prevent 30 lines of code duplications.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1190,20 +1220,66 @@ mod pallet {
...
+            // We add one to the count, because `pallet-timestamp` sets the timestamp _after_
+            // `on_initialize` is called, so calling `now()` during `on_initialize` gives us
+            // the timestamp of the previous block.
+            let current_time_frame =
+                Self::calculate_time_frame_of_moment(T::MarketCommons::now()).saturating_add(1);
+
+            // On first pass, we use current_time - 1 to ensure that the chain doesn't try to
+            // check all time frames since epoch.
+            let last_time_frame =
+                LastTimeFrame::<T>::get().unwrap_or_else(|| current_time_frame.saturating_sub(1));
```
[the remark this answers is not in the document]
maltekliemann: `current_time_frame` will never be zero because we can assume that `pallet-timestamp` works correctly on block 2 or later. But if it were zero, then `last_time_frame` would be zero as well, and we would not handle any markets.

The linked line of code looks correct. Are you worried about something specific?

zrml/prediction-markets/src/lib.rs
```diff
@@ -1190,20 +1220,66 @@ mod pallet {
...
+            // of 0 to the current time frame.
+            if now <= 1u32.into() {
+                return total_weight;
+            }
+
+            // We add one to the count, because `pallet-timestamp` sets the timestamp _after_
+            // `on_initialize` is called, so calling `now()` during `on_initialize` gives us
+            // the timestamp of the previous block.
+            let current_time_frame =
+                Self::calculate_time_frame_of_moment(T::MarketCommons::now()).saturating_add(1);
```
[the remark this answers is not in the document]
maltekliemann: Yeah, it's not exactly part of the API when `pallet-timestamp` actually start working properly. I guess I'll open an issue on `substrate` about this.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1190,20 +1220,66 @@ mod pallet {
...
+                let open = Self::market_status_manager::<
+                    _,
+                    MarketIdsPerOpenBlock<T>,
+                    MarketIdsPerOpenTimeFrame<T>,
+                >(
+                    now,
+                    last_time_frame,
+                    current_time_frame,
+                    |market_id, _| {
+                        let weight = Self::open_market(market_id)?;
```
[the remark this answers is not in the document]
maltekliemann: Then again `on_market_close` has a completely different purpose than `close_market`, and `open_market` should follow that pattern. Plus, we might need `open_market` at another point in the code in the near future.

zrml/prediction-markets/src/migrations.rs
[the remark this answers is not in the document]
maltekliemann: Yeah, pretty sure `try_push` requires a read and a write.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+        let current_time_frame =
+            Pallet::<T>::calculate_time_frame_of_moment(T::MarketCommons::now());
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(2));
+
+        for (market_id, market) in T::MarketCommons::market_iter() {
+            total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+
+            if market.status != MarketStatus::Active {
+                continue;
+            }
```
sea212, line 47: You could inject this constraint into the iteration using the functional programming style function [`filter`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter).

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+                Err(_) => continue,
+            };
+
+            let not_yet_open = match market.period {
+                MarketPeriod::Block(ref range) => current_block < range.start,
+                MarketPeriod::Timestamp(ref range) => {
+                    current_time_frame < Pallet::<T>::calculate_time_frame_of_moment(range.start)
+                }
+            };
+            if not_yet_open {
```
sea212, line 62: This if expressions covers the rest of the code for the pool. Can't we just remove it and instead continue the loop in the match expression before?

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+                        }
+                    }
+                    total_weight =
+                        total_weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+                } else {
+                    log::warn!(
+                        "found pool with unexpected status. market_id: {:?}. pool_id: {:?}",
+                        market_id,
+                        pool_id,
+                    );
```
sea212, line 103: Good job on the extensive logging in this migration :+1:

sea212 (approved):
Looks good :+1:

zrml/prediction-markets/src/migrations.rs
maltekliemann: Mitigated in ff241d88d9e2f51d76b2a9aa8cd8368671245e23

zrml/prediction-markets/src/migrations.rs
maltekliemann: Mitigated in 5daa4aca395497c7a91441a5bbb741a6ac98dee1

sea212 (approved):

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
Chralt98: OK for the first. About the linked code: Nevermind (I just wanted to make sure, that you look at it again).

Chralt98 (approved):
.

PR #713, opened 2022-07-18
Minor refactoring in process_subsidy_collecting_markets()

vivekvpandya:
This fixes https://github.com/zeitgeistpm/runtime-audit-1/issues/12

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -1845,29 +1845,22 @@ mod pallet {
             let one_read = T::DbWeight::get().reads(1);
             let one_write = T::DbWeight::get().writes(1);
 
-            let retain_closure =
-                |subsidy_info: &SubsidyUntil<T::BlockNumber, MomentOf<T>, MarketIdOf<T>>| {
-                    let market_ready = match &subsidy_info.period {
-                        MarketPeriod::Block(period) => period.start <= current_block,
-                        MarketPeriod::Timestamp(period) => period.start <= current_time,
-                    };
+            let retain_closure = |subsidy_info: &SubsidyUntil<
+                T::BlockNumber,
+                MomentOf<T>,
+                MarketIdOf<T>,
+            >| {
+                let market_ready = match &subsidy_info.period {
+                    MarketPeriod::Block(period) => period.start <= current_block,
+                    MarketPeriod::Timestamp(period) => period.start <= current_time,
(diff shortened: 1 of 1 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -1845,29 +1845,22 @@ mod pallet {
...
-                            market_id: {:?}, error: {:?}",
-                                pool_id,
-                                err
-                            );
-                            return true;
-                        }
-
-                        let end_subsidy_result =
-                            T::Swaps::end_subsidy_phase(pool_id.unwrap_or(u128::MAX));
+                    if let Ok(pool_id) = pool_id {
```
Chralt98, line 1862: Feels better than the mitigation suggestion, because if someone would have changed this in the future, we can not be sure, that the `expect` panic would have been caught (although, the tests would fail). It's more code capsulation here now. I am thinking about reducing the code capsulation with functions, but maybe this refactoring is for another PR. 

> What you could do is shadow the variable right after the if-condition a la let pool_id = pool_id.expect("early return happens above in case of Err");. Then you can remove the two .unwrap_or(u128::MAX) in the function.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1981,10 +1973,19 @@ mod pallet {
                                 err
                             );
                         }
+                    } else if let Err(err) = pool_id {
+                        log::error!(
+                            "[PredictionMarkets] Cannot find pool associated to market.
+                            market_id: {:?}, error: {:?}",
+                            subsidy_info.market_id,
```
Chralt98, line 1980: Good that you use the market_id now instead of the [pool result here](https://github.com/zeitgeistpm/zeitgeist/blob/9b1e7091bf429d87a590a74de6f6941d5c50e70f/zrml/prediction-markets/src/lib.rs#L1813).

zrml/prediction-markets/src/lib.rs
Chralt98: I probably like the `match` here more (because it's even more clear, that there are no more cases than `Ok()` or `Err()`), but this would lead to even more code capsulation, so you can leave it as it is.

Chralt98 (approved):
Looks like a nice code improvement.

sea212 (approved):
Nice restructure, more transparency in logs, future me thanks you :+1:

PR #719, opened 2022-07-20
Remove unchecked integer arithmetic

### LINKED ISSUE #625: Investigate why clippy doesn't report unchecked integer arithmetic
For example, we have the following code snippet: https://github.com/zeitgeistpm/zeitgeist/blob/693c5c0039948787322b3f35e83dc8a298bb482b/zrml/prediction-markets/src/lib.rs#L975-L988

Why does clippy not warn about this?

maltekliemann:
Fixes #625

### DIFF
--- a/zrml/court/src/lib.rs
+++ b/zrml/court/src/lib.rs
@@ -287,8 +287,10 @@ mod pallet {
 
             let slash_and_remove_juror = |ai: &T::AccountId| {
                 let all_reserved = CurrencyOf::<T>::reserved_balance_named(&RESERVE_ID, ai);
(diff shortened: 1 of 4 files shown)


maltekliemann:
Thanks for the review!

I `grep`'ed for the following regular expressions: `+`, `-[^>]` (in hindsight, ` + ` might have been better ` - `), `[^\/]\/[^\/]` (in order to avoid doc strings), `\*` (` * ` might have been better) and `[+-*/]=`.

lsaether (approved):
Changes look good. I did not check for any unsigned arithmetic that was missed though.

PR #724, opened 2022-07-28
Add license headers

### LINKED ISSUE #477: Add copyright/license notice to all source files
According to https://www.gnu.org/licenses/gpl-faq.en.html#WhyMustIInclude:

> You should put a notice at the start of each source file, stating what license it carries, in order to avoid risk of the code's getting disconnected from its license.

This seems pretty old-school, but it would make attribution for some code from other repositories easier. If you don't like boilerplate license text at the start of each source file, the [REUSE-3.0 spec](https://reuse.software/spec/) is kind of a solution.

maltekliemann:
This PR adds GPL-3.0-or-later license headers to all `*.rs` files. I've added a citation when I was certain that the file contained code from other projects. For example: https://github.com/maltekliemann/zeitgeist/blob/licenses/runtime/src/benchmarking/currencies.rs. Please check if I've missed something. :pray:

- My method of citation is borrowed from [softwarefreedom.org](https://softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html) and is a common method of citation in the open-source space.
- It is sometimes suggested that _every_ file in the project needs to be equipped with a license header. I've followed other projects of the ecosystem in _not_ doing this (aside from the fact that, for example, `*.json` won't allow this).
- The Balancer code does not contain a copyright notice. I've consulted with the good people of [opensource.stackexchange.com](opensource.stackexchange.com) and decided to cite without a copyright notice and instead add some contact information.
- Unlicensed code (from the substrate node template, for example) is licensed under GPL-3.0-or-later in our repository. I'm following the example of other projects from the Polkadot ecosystem regarding unlicensed code.

This fixes #477 and zeitgeistpm/runtime-audit-1#15. It also marks the last pull request I'm doing from my private fork. :tada:


sea212 (approved):
Thanks, LGTM.

Checked that all source files contain the license header using the following shell script:
```sh
LICENSE_STR="Copyright 2021-2022 Zeitgeist PM LLC."
PROJECT_PATH="$HOME/Downloads/zeitgeist-licenses"
RED='\033[1;31m'
NC='\033[0m'

for filename in $(find $PROJECT_PATH -type f -name "*.rs"); do
    if ! $(grep -q "$LICENSE_STR" $filename); then
        echo -e "WARNING: ${RED} $filename ${NC} does not have a license header!"
    fi
done
```

sea212 (changes requested):
We should also add a `LICENSE` file in the root of the repository that contains the GPLv3 license, [llike in this example](https://github.com/PureStake/moonbeam/blob/e24763f705590c359372d9c478ba6a561fc294b4/LICENSE)

maltekliemann:
> We should also add a `LICENSE` file in the root of the repository that contains the GPLv3 license, [llike in this example](https://github.com/PureStake/moonbeam/blob/e24763f705590c359372d9c478ba6a561fc294b4/LICENSE)

We already got one: https://github.com/zeitgeistpm/zeitgeist/blob/main/LICENSE (to be fair, I could have pointed that out earlier)

sea212:
Oh I should've realized that :facepalm:

sea212 (approved):

sea212 (approved):
LGTM, script confirms that every file has the license header.

lsaether (approved):

PR #725, opened 2022-07-29
pallet/zrml-styx - burn ztg for the ability to claim avatar

yornaath:
**Not decided if should be implemented yet.** _Edit: Seems like its a go_ ✅ 

After a talk with @lsaether we came up with the idea to have the users burn some ZTG for the ability to claim their avatar.

> I like burning because it creates some deflationary pressure of the token to offset inflation while we currently don't have much tx volume. And instead of us being paid to subsidize minting, burning would benefit all ZTG holders equally - @lsaether 

**Work done:**
This pr introduces the pallet `styx` with the core call `cross()`. This `balance.slashes()` 100 ZTG from the calling account letting the account cross and incarnate as an avatar. An account is only able to cross once, storing crossings pr `AccountId` in a `StorageMap`.

**The amount burned can be modified by root or 1/2 council vote `styx.setBurnAmount(amount: u128 (BalanceOf))`.**

The minting service will be in charge of querying the chain storage when the user issues(claims) their avatar to see if they have crossed. This on chain extrinsic for crossing will be submitted through the frontend UI and then the UI will call the minting service api to issue the mint. 

_Loosely coupled event listening to automate the process when a `styx.AccountCrossed` occurs, automatically issuing the mint could also be implemented in the minting service._


yornaath:
Burn amount value is now stored in a `StorageValue` and can be modified by extrinsic submission with 50% Council Vote.

maltekliemann:
You can most likely replace `Currencies`/`AssetManager` and `Tokens` with just `Balances`. This would somewhat simplify `mock.rs` and reduce the number of dependencies.

yornaath:
Ty for the guidance! @maltekliemann 👏 🍻

yornaath:
> You can most likely replace `Currencies`/`AssetManager` and `Tokens` with just `Balances`. This would somewhat simplify `mock.rs` and reduce the number of dependencies.

Yeah I was thinking the same thing 👍  Will get on it tomorrow!

sea212:
Hey Jorn, cool to see you getting started with Substrate.

Just a quick notice: We use only exactly one of the labels for the `s:` type. They indicate the current state of the pull request. For example, the pull request mustn't be `s:in-progress` and `s:review-needed`, because a review should only happen once the work is done / is not in progress anymore.

We also need some [benchmarking logic](https://docs.substrate.io/main-docs/test/benchmark/) to determine the correct [weights](https://docs.substrate.io/main-docs/build/tx-weights-fees/).  *A not desirable alternative would be to use huge weights in the weight annotation of the dispatchable calls, they should definitely be greater than the real value to avoid DDoS.*
They idea behind weights is that they represent an upper boundary for the computation time that this function needs (in respect to some optional parameters). Writing benchmarks is not that difficult once done:  You only have to initialize the state and then execute the dispatchable. If it contains loops, you'll also have to add additional parameters that represent the iteration count of the loop.
I'd be happy to see you implement the benchmarking logic, however I'll also be open to do it if you won't.

yornaath:
> Hey Jorn, cool to see you getting started with Substrate.
> 
> Just a quick notice: We use only exactly one of the labels for the `s:` type. They indicate the current state of the pull request. For example, the pull request mustn't be `s:in-progress` and `s:review-needed`, because a review should only happen once the work is done / is not in progress anymore.
> 
> We also need some [benchmarking logic](https://docs.substrate.io/main-docs/test/benchmark/) to determine the correct [weights](https://docs.substrate.io/main-docs/build/tx-weights-fees/). _A not desirable alternative would be to use huge weights in the weight annotation of the dispatchable calls, they should definitely be greater than the real value to avoid DDoS._ They idea behind weights is that they represent an upper boundary for the computation time that this function needs (in respect to some optional parameters). Writing benchmarks is not that difficult once done: You only have to initialize the state and then execute the dispatchable. If it contains loops, you'll also have to add additional parameters that represent the iteration count of the loop. I'd be happy to see you implement the benchmarking logic, however I'll also be open to do it if you won't.

Ty man!

Ah yeah got it! I put it to in progress and added the review needed since I need some continual guidance haha.
And yeah I would definitely love to tackle implementing benchmarking and proper weighting :D Having a lot of fun with this!

Cargo.toml
```diff
@@ -1,27 +1,28 @@
...
+  "zrml/swaps/runtime-api",
+  "zrml/styx",
 ]
 resolver = "2"
 
 [patch."https://github.com/purestake/moonbeam"]
-pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
-parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
+pallet-author-mapping = {git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping"}
+parachain-staking = {git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping"}
```
maltekliemann, line 28: It's standard style to have whitespace between curly braces and their content.

```suggestion
pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
```

runtime/src/lib.rs
```diff
@@ -47,6 +47,7 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 use substrate_fixed::{types::extra::U33, FixedI128, FixedU128};
 use zeitgeist_primitives::{constants::*, types::*};
+
```
maltekliemann, line 50:
```suggestion
```

zrml/styx/README.md
```diff
@@ -0,0 +1 @@
+License: Unlicense
```
maltekliemann, line 1: This file should be deleted. We will license this code unter GPL-3.0-or-later.

zrml/styx/src/benchmarks.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::Pallet as Styx;
+use frame_benchmarking::{benchmarks, impl_benchmark_test_suite, whitelisted_caller};
+use frame_support::{
+    dispatch::UnfilteredDispatchable,
+    traits::{Currency, EnsureOrigin},
+};
+use frame_system::RawOrigin;
+use sp_runtime::SaturatedConversion;
+
+benchmarks! {
```
maltekliemann, line 20: You should use four spaces for indentation. In case you're wondering: RustFmt cannot reformat code written inside a macro for correctness reasons. :wink:

zrml/styx/src/benchmarks.rs
```diff
@@ -0,0 +1,37 @@
...
+  cross {
+    let caller: T::AccountId = whitelisted_caller();
+    let balance = 900_000_000_000_000_000u128.saturated_into();
```
maltekliemann, line 23:
```suggestion
        let balance = (90_000_000 * BASE).saturated_into();
```

zrml/styx/src/benchmarks.rs
```diff
@@ -0,0 +1,37 @@
...
+    T::Currency::deposit_creating(&caller, balance);
+  }: _(RawOrigin::Signed(caller))
+
+  set_burn_amount {
+    let origin = T::SetBurnAmountOrigin::successful_origin();
+    let caller: T::AccountId = whitelisted_caller();
+    let balance = 100_000_000_000_000u128.saturated_into();
+    let amount = 200_000_000_000u128.saturated_into();
```
maltekliemann, line 31:
```suggestion
        let balance = (10_000 * BASE).saturated_into();
        let amount = (20_000 * BASE).saturated_into();
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+extern crate alloc;
+
+mod benchmarks;
+mod mock;
+mod tests;
+pub mod weights;
+/// Edit this file to define custom logic or remove it if it is not needed.
+/// Learn more about FRAME and the core library of Substrate FRAME pallets:
+/// <https://docs.substrate.io/v3/runtime/frame>
```
maltekliemann, line 12:
```suggestion
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::too_many_arguments)]
```
maltekliemann, line 2: You don't need this. :wink:

```suggestion
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    use zeitgeist_primitives::types::Balance;
+
+    use crate::weights::WeightInfoZeitgeist;
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// The origin that is allowed to destroy markets.
+        type SetBurnAmountOrigin: EnsureOrigin<Self::Origin>;
+
+        /// Pallet event type
```
maltekliemann, line 32: [nitpick]

```suggestion
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    };
+    use frame_system::pallet_prelude::*;
+    use sp_runtime::SaturatedConversion;
+    use zeitgeist_primitives::types::Balance;
+
+    use crate::weights::WeightInfoZeitgeist;
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// The origin that is allowed to destroy markets.
```
maltekliemann, line 29:
```suggestion
        /// The origin that is allowed to set the amount burned when crossing Styx.
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        type SetBurnAmountOrigin: EnsureOrigin<Self::Origin>;
+
+        /// Pallet event type
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+        /// Balances pallet
```
maltekliemann, line 35: [nitpick]

```suggestion
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        type Currency: NamedReservableCurrency<Self::AccountId, ReserveIdentifier = [u8; 8]>;
+
+        /// Weights generated by benchmarks
```
maltekliemann, line 38: [nitpick]

```suggestion
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// The origin that is allowed to destroy markets.
+        type SetBurnAmountOrigin: EnsureOrigin<Self::Origin>;
+
+        /// Pallet event type
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+        /// Balances pallet
+        type Currency: NamedReservableCurrency<Self::AccountId, ReserveIdentifier = [u8; 8]>;
```
maltekliemann, line 36: [nitpick] This is fine, but you're currently not making any use of named reserves, so you could go with a more lightweight trait here, for example the `Currency` trait that's already imported.

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+            if Crossings::<T>::contains_key(&who) {
+                Err(Error::<T>::HasAlreadyCrossed)?;
+            }
+
+            let amount = BurnAmount::<T>::get().saturated_into();
+            let free = T::Currency::free_balance(&who);
+
+            if free < amount {
+                Err(Error::<T>::FundDoesNotHaveEnoughFreeBalance)?;
+            }
```
maltekliemann, line 92: I'd replace this with an application of `can_slash`.

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        type WeightInfo: WeightInfoZeitgeist;
+    }
+
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    pub struct Pallet<T>(_);
+
+    /// Keep track of crossings. Accounts are only able to cross once.
+    #[pallet::storage]
+    pub type Crossings<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, bool>;
```
maltekliemann, line 48: Since you're never using the `bool`, just replace it with the null type `()`:

```rust
pub type Crossings<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, ()>;
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+            Ok(())
+        }
+
+        /// Set the burn amount. Needs 50% council vote.
+        #[pallet::weight(T::WeightInfo::set_burn_amount())]
+        pub fn set_burn_amount(
+            origin: OriginFor<T>,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResult {
+            T::SetBurnAmountOrigin::ensure_origin(origin.clone())?;
```
maltekliemann: [nitpick] No need to `clone` here

```suggestion
            T::SetBurnAmountOrigin::ensure_origin(origin)?;
```

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+            T::Currency::slash(&who, amount);
+            Crossings::<T>::insert(&who, true);
+
+            Self::deposit_event(Event::AccountCrossed(who, amount.saturated_into()));
+
+            Ok(())
+        }
+
+        /// Set the burn amount. Needs 50% council vote.
```
maltekliemann, line 102: "Needs 50% council vote" is not correct. That's how we implement it in the runtime. But here, the calling origin must be `SetBurnAmountOrigin`, that's it.

zrml/styx/src/mock.rs
```diff
@@ -0,0 +1,113 @@
...
+    pub enum Runtime
+    where
+        Block = BlockTest<Runtime>,
+        NodeBlock = BlockTest<Runtime>,
+        UncheckedExtrinsic = UncheckedExtrinsicTest<Runtime>,
+    {
+        Balances: pallet_balances::{Call, Config<T>, Event<T>, Pallet, Storage},
+        Styx: zrml_styx::{Event<T>, Pallet, Storage},
+        System: frame_system::{Call, Config, Event<T>, Pallet, Storage},
+        Timestamp: pallet_timestamp::{Pallet}
```
maltekliemann, line 36: [nitpick] I think we can get rid of `Timestamp`, as well.

maltekliemann (changes requested):
First of all, great job! :thumbsup: Glad you're enjoying substrate. Most proposed changes are quite specific, some of them outright nitpicky.

One very important thing, however: We need maximum test coverage. In case of this pallet, I'm missing the following tests:

- `cross` fails if the user has insufficient funds
- `cross` fails if the user has already crossed
- `cross` succeeds in the sense that the user's id is now in storage and the funds have actually been slashed from the user's wallet
- `set_burn_amount` fails if the origin is not `SetBurnAmountOrigin`
- `set_burn_amount` succeeds in the sense that storage is altered
- (maybe not super important, but I would add this, as well) `set_burn_amount` works in the sense that `cross` now uses the new value
- You're already checking for correct emission of events, so that's good :thumbsup:

Use `asset_noop!` to test for failure. This macro makes sure that chain storage was not altered. So you get the guarantee that you don't accidentally slash someone before you fail the function because they've already crossed Styx.

And one other thing: What's the purpose of `zeitgeist_spec.json`? Looks like an accidental `git add`.

zrml/styx/src/lib.rs
maltekliemann: You can use the `ensure!` macro here, by the way.

zrml/styx/src/lib.rs
yornaath: I wrote that because the message shows in the polkadotjs ui to the user giving some hint that its not callable by anon. Should it read differently you think?

zrml/styx/src/lib.rs
yornaath: Ah didnt see this one. Will fix tmrw 👍

zrml/styx/src/lib.rs
maltekliemann: Yes, the pallet documentation should reflect what's actually happening in the pallet, not what's happening in the runtime that uses the pallet.

Cargo.toml
maltekliemann, line 28:
```suggestion
pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
```

node/Cargo.toml
```diff
@@ -3,158 +3,149 @@ name = "zeitgeist"
 path = "./src/main.rs"
```
maltekliemann, line 3: You seem to have autoformatted this file into oblivion. These changes are not consistent with the standard `.toml` style.

runtime/Cargo.toml
```diff
@@ -1,331 +1,288 @@
 [build-dependencies]
-substrate-wasm-builder = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
+substrate-wasm-builder = {branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate"}
```
maltekliemann, line 3: You seem to have autoformatted this file into oblivion. These changes are not consistent with the standard `.toml` style.

zrml/styx/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[dependencies]
```
maltekliemann, line 1: This file should be formatted like the other `.toml` files.

zrml/styx/src/benchmarks.rs
maltekliemann: You should also indent everything inside `benchmarks!` with four instead of two spaces.

zrml/styx/src/benchmarks.rs
```diff
@@ -0,0 +1,38 @@
...
+      let caller: T::AccountId = whitelisted_caller();
+      let balance = (90_000_000 * BASE).saturated_into();
+      T::Currency::deposit_creating(&caller, balance);
+  }: _(RawOrigin::Signed(caller))
+
+  set_burn_amount {
+      let origin = T::SetBurnAmountOrigin::successful_origin();
+      let caller: T::AccountId = whitelisted_caller();
+      let balance = 100_000_000_000_000u128.saturated_into();
+      let amount = 200_000_000_000u128.saturated_into();
```
maltekliemann, line 32: You should use `BASE` here instead of writing out the numbers.

maltekliemann (changes requested):

Cargo.toml
yornaath: Why doesnt `cargo fmt` respect this?

yornaath: Oh actually I might have done a format on this from my editor so nm

zrml/styx/src/tests.rs
```diff
@@ -0,0 +1,69 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        frame_system::Pallet::<Runtime>::set_block_number(1);
+        let burn_amount = crate::BurnAmount::<Runtime>::get();
+        let original_balance = Balances::free_balance(&ALICE);
+        assert_ok!(Styx::cross(Origin::signed(ALICE)));
+        let balance_after_crossing = Balances::free_balance(&ALICE);
+        let diff = original_balance - balance_after_crossing;
+        let tx_fee_margin = 5_000_000_000u128;
+        assert_eq!(Crossings::<Runtime>::contains_key(&ALICE), true);
+        assert!(diff >= burn_amount && diff <= burn_amount + tx_fee_margin);
```
yornaath, line 17: @maltekliemann Do you have any suggestion for a more accurate way of taking fee into consideration when testing that the amount actually has been slashed by the cross call?

maltekliemann: Sorry, I forgot to comment on this. There are no fees in the mock. So no need to worry about it. You can just check

```rust
assert_ok!(diff, <Runtime as Config>::BurnValue::get())
```

yornaath:
It seems like we are deciding on implementing this when ready. Can you confirm @lsaether?

node/Cargo.toml
```diff
@@ -100,61 +100,52 @@ log = { optional = true, version = "0.4.14" }
...
-    "zeitgeist-runtime/runtime-benchmarks"
+  "frame-benchmarking/runtime-benchmarks",
+  "polkadot-service?/runtime-benchmarks",
+  "zeitgeist-runtime/runtime-benchmarks",
 ]
 try-runtime = [
-    "zeitgeist-runtime/try-runtime",
-    "try-runtime-cli",
+  "zeitgeist-runtime/try-runtime",
+  "try-runtime-cli",
```
maltekliemann, line 148: I think this reformat is a bit too aggressive. I admit that we're somewhat inconsistent with `.toml` formats, so I don't mind changing the indent (as you did in other files), but in this case, the comments get all screwed up.

maltekliemann (approved):
:thumbsup:

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+extern crate alloc;
+
```
vivekvpandya, line 4: Suggestion: Nice to have a README or some DOC here in lib file like other pallets have.

vivekvpandya (approved):
:+1:

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+            Ok(())
+        }
+
+        /// Set the burn amount. Ensures SetBurnAmountOrigin.
+        #[pallet::weight(T::WeightInfo::set_burn_amount())]
+        pub fn set_burn_amount(
+            origin: OriginFor<T>,
+            #[pallet::compact] amount: Balance,
+        ) -> DispatchResult {
+            T::SetBurnAmountOrigin::ensure_origin(origin)?;
```
vivekvpandya, line 97: should we add check for amount != zero? as if it is zero then essentially there is no burning.

zrml/styx/src/lib.rs
maltekliemann: Great point. @yornaath Can you add docs that describe the pallet's use?

zrml/styx/src/lib.rs
maltekliemann: Yes and no. I don't think it's critical here because the `SetBurnAmountOrigin` is assumed to be someone who knows what they're doing.

Also, this seems to be a difficult question of where to draw the line: If you don't allow `0`, then what about `1`? That's also essentially no burning. What about `10`? Etc. etc.

zrml/styx/src/lib.rs
yornaath: > Suggestion: Nice to have a README or some DOC here in lib file like other pallets have.

Is the standard to have the docs as a top level comment in the lib.rs file with markdown syntax?

vivekvpandya:
Rust docs style comments , no need for markup.

zrml/styx/src/lib.rs
maltekliemann: That's not uncommon, but the prevailing style in our repository is to use doc strings.

yornaath:
I will get on this today 👏

yornaath:
@vivekvpandya Docs included in lib.rs, can you check if they conform to our standards?

zrml/styx/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+//! # Styx
+//!
+//! A module for burning som native chain tokens to be able to claim their avatar.
```
maltekliemann, line 3:
```suggestion
//! A module for burning native chain tokens in order to gain entry into a registry for off-chain use.
```

maltekliemann (changes requested):
I've made a small suggestion for how to approach the documentation. I like the naming of the pallet, but the documentation should do without the mythological background and describe exactly what the pallet does.

In particular, there shouldn't really be any mention of the NFT because this pallet doesn't issue the NFT, nor does it give any guarantee that an NFT will be issued by some other system.

yornaath:
I believe issues should be adequately fixed. Let me know.

maltekliemann (approved):
Looks good to me! :thumbsup:

Issue #646, opened 2022-05-23
`MarketIdsPerReportBlock` and `MarketIdsPerDisputeBlock` are never cleaned up


maltekliemann:
`MarketIdsPerReportBlock` and `MarketIdsPerDisputeBlock` are used as helper hash maps to quickly determine where market state needs to be changed. Once the market is resolved, they are no longer needed, but they are never cleaned up. Doing so should save some space in the long run.

Chralt98:
Isn't [this](https://github.com/zeitgeistpm/zeitgeist/blob/main/zrml/prediction-markets/src/lib.rs#L1316-L1364) fixing the issue?

Chralt98:
I realised, that only the admin extrinsics and the `reject_market` function calls [this](https://github.com/zeitgeistpm/zeitgeist/blob/main/zrml/prediction-markets/src/lib.rs#L1316-L1364). So, could be the solution to call these functions in `on_initialize` after the resolution succeeded?

PR #729, opened 2022-08-01
Cleanup MarketIdsPerReportBlock and MarketIdsPerDisputeBlock in resolution_manager()

vivekvpandya:
Fixes https://github.com/zeitgeistpm/zeitgeist/issues/646
Implement a migration to clean up old values from storage.

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -110,7 +110,7 @@ mod pallet {
     pub const RESERVE_ID: [u8; 8] = PmPalletId::get().0;
 
     /// The current storage version.
-    const STORAGE_VERSION: StorageVersion = StorageVersion::new(3);
+    const STORAGE_VERSION: StorageVersion = StorageVersion::new(4);
 
     pub(crate) type BalanceOf<T> = <<T as Config>::AssetManager as MultiCurrency<
         <T as frame_system::Config>::AccountId,
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -119,7 +119,7 @@ mod pallet {
     pub(crate) type MarketIdOf<T> =
         <<T as Config>::MarketCommons as MarketCommonsPalletApi>::MarketId;
     pub(crate) type MomentOf<T> = <<T as Config>::MarketCommons as MarketCommonsPalletApi>::Moment;
-    type CacheSize = ConstU32<64>;
+    pub type CacheSize = ConstU32<64>;
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
@@ -2210,12 +2210,14 @@ mod pallet {
                     cb(id, &market)?;
                 }
             }
+            MarketIdsPerReportBlock::<T>::remove(&block);
 
             // Resolve any disputed markets.
             for id in MarketIdsPerDisputeBlock::<T>::get(&block).iter() {
                 let market = T::MarketCommons::market(id)?;
                 cb(id, &market)?;
             }
+            MarketIdsPerDisputeBlock::<T>::remove(&block);
 
             Ok(())
         }
@@ -2318,7 +2320,7 @@ mod pallet {
         )
     }
 
-    fn remove_item<I: cmp::PartialEq, G>(items: &mut BoundedVec<I, G>, item: &I) {
+    pub fn remove_item<I: cmp::PartialEq, G>(items: &mut BoundedVec<I, G>, item: &I) {
         if let Some(pos) = items.iter().position(|i| i == item) {
             items.swap_remove(pos);
         }
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -2320,7 +2320,7 @@ mod pallet {
         )
     }
 
-    pub fn remove_item<I: cmp::PartialEq, G>(items: &mut BoundedVec<I, G>, item: &I) {
+    fn remove_item<I: cmp::PartialEq, G>(items: &mut BoundedVec<I, G>, item: &I) {
         if let Some(pos) = items.iter().position(|i| i == item) {
             items.swap_remove(pos);
         }
--- a/zrml/prediction-markets/src/migrations.rs
+++ b/zrml/prediction-markets/src/migrations.rs
@@ -1,10 +1,18 @@
-use crate::{Config, MarketIdsPerOpenBlock, MarketIdsPerOpenTimeFrame, Pallet};
+use crate::{
+    Config, MarketIdOf, MarketIdsPerDisputeBlock, MarketIdsPerOpenBlock, MarketIdsPerOpenTimeFrame,
+    MarketIdsPerReportBlock, Pallet,
+};
 use frame_support::{
     dispatch::Weight,
     log,
     pallet_prelude::PhantomData,
-    traits::{Get, OnRuntimeUpgrade, StorageVersion},
+    storage::PrefixIterator,
+    traits::{ConstU32, Get, OnRuntimeUpgrade, StorageVersion},
+    BoundedVec, Twox64Concat,
 };
+use sp_runtime::traits::Saturating;
+extern crate alloc;
+use alloc::vec::Vec;
 use zeitgeist_primitives::types::{MarketPeriod, MarketStatus, PoolStatus};
 use zrml_market_commons::MarketCommonsPalletApi;
 
@@ -14,6 +22,8 @@ const PREDICTION_MARKETS_REQUIRED_STORAGE_VERSION: u16 = 2;
 const PREDICTION_MARKETS_NEXT_STORAGE_VERSION: u16 = 3;
 
 pub struct MigrateMarketPoolsBeforeOpen<T>(PhantomData<T>);
+
+pub struct CleanUpStorageForResolvedOrClosedMarkets<T>(PhantomData<T>);
 
 impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
     fn on_runtime_upgrade() -> Weight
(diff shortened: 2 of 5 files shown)


vivekvpandya:
This should fix https://github.com/zeitgeistpm/zeitgeist/issues/646

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+            log::info!("Skipping storage cleanup; prediction-markets already up to date");
+            return total_weight;
+        }
+
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(1));
+        log::info!("Starting storage cleanup of CleanUpStorageForResolvedOrClosedMarkets");
+
+        let dispute_period = T::DisputePeriod::get();
+        let current_block: T::BlockNumber = <frame_system::Pallet<T>>::block_number();
+        let block = current_block.saturating_sub(dispute_period);
```
Chralt98, line 36:
```suggestion
        let last_dp_end_block = current_block.saturating_sub(dispute_period);
```

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+        let market_ids_tobe_removed_per_dispute: Vec<_> =
+            market_ids_per_dispute_iterator.filter(|v| v.0 <= block).collect();
+        for (k, _v) in market_ids_tobe_removed_per_dispute {
+            total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
+            MarketIdsPerDisputeBlock::<T>::remove(k);
+        }
+
+        let market_ids_per_report_iterator: IterType<T> =
+            frame_support::migration::storage_key_iter::<_, _, Twox64Concat>(
+                b"PredictionMarkets",
```
Chralt98, line 57: Please look above.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+                b"MarketIdsPerReportBlock",
+            );
+
+        let market_ids_tobe_removed_per_report: Vec<_> =
+            market_ids_per_report_iterator.filter(|v| v.0 <= block).collect();
```
Chralt98, line 62: Please look above.

Chralt98 (commented):
Ok, this is a storage migration to delete storage items from resolved markets. This PR also deletes expired storage items after the `on_resolution` calls automatically.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+        let current_block: T::BlockNumber = <frame_system::Pallet<T>>::block_number();
+        let block = current_block.saturating_sub(dispute_period);
+        type IterType<T> = PrefixIterator<(
+            <T as frame_system::Config>::BlockNumber,
+            BoundedVec<MarketIdOf<T>, ConstU32<1024>>,
+        )>;
+
+        let market_ids_per_dispute_iterator: IterType<T> =
+            frame_support::migration::storage_key_iter::<_, _, Twox64Concat>(
+                b"PredictionMarkets",
```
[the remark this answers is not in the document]
vivekvpandya: I had asked this to @maltekliemann in similar PR (for court) earlier and in test I found that pallet id does not work. what seems to work is the name of module as it is added in Runtime.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+    dispatch::Weight,
+    log,
+    pallet_prelude::PhantomData,
+    storage::PrefixIterator,
+    traits::{ConstU32, Get, OnRuntimeUpgrade, StorageVersion},
+    BoundedVec, Twox64Concat,
+};
+
+use sp_runtime::traits::Saturating;
+use sp_std::vec::Vec;
```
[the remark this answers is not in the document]
vivekvpandya: shouldn't we use sp_std when its possible? I remember we discussed this on chat I guess.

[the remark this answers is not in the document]
Chralt98: I also remember the chat about that in the global disputes PR https://github.com/zeitgeistpm/zeitgeist/pull/682#discussion_r923105565.

zrml/prediction-markets/src/migrations.rs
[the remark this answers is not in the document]
Chralt98: Ok, so here https://github.com/zeitgeistpm/zeitgeist/blob/d9d180353dfa80e8e9e75afc7ac89ebc8c11d1bb/runtime/src/lib.rs#L217

zrml/prediction-markets/src/migrations.rs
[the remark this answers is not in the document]
vivekvpandya: okay I see , decision from that discussion is to use native crate.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2083,12 +2084,14 @@ mod pallet {
...
                 }
             }
+            MarketIdsPerReportBlock::<T>::remove(&block);
 
             // Resolve any disputed markets.
             for id in MarketIdsPerDisputeBlock::<T>::get(&block).iter() {
                 let market = T::MarketCommons::market(id)?;
                 cb(id, &market)?;
             }
+            MarketIdsPerDisputeBlock::<T>::remove(&block);
```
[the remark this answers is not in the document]
maltekliemann: Yeah, good point. Could you please add that to one of the issues regarding the upcoming benchmark fixes. the `*_manager` functions should probably just return their own weight plus the weight of the mutation.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -0,0 +1,187 @@
...
+        )>;
+
+        let market_ids_per_dispute_iterator: IterType<T> =
+            frame_support::migration::storage_key_iter::<_, _, Twox64Concat>(
+                b"PredictionMarkets",
+                b"MarketIdsPerDisputeBlock",
+            );
+
+        let market_ids_tobe_removed_per_dispute: Vec<_> =
+            market_ids_per_dispute_iterator.filter(|v| v.0 <= block).collect();
```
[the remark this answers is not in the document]
vivekvpandya: Updated.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -296,6 +447,52 @@ mod tests {
...
+            assert_eq!(MarketIdsPerDisputeBlock::<Runtime>::get(dispute_block).len(), 2);
+            assert_eq!(MarketIdsPerReportBlock::<Runtime>::get(dispute_block).len(), 2);
+            CleanUpStorageForResolvedOrClosedMarkets::<Runtime>::on_runtime_upgrade();
+            assert_eq!(MarketIdsPerDisputeBlock::<Runtime>::get(dispute_block).len(), 0);
+            assert_eq!(MarketIdsPerReportBlock::<Runtime>::get(dispute_block).len(), 0);
+
+            let dispute_block = System::current_block_number();
+            MarketIdsPerDisputeBlock::<Runtime>::insert(dispute_block, market_ids.clone());
+            MarketIdsPerReportBlock::<Runtime>::insert(dispute_block, market_ids.clone());
+            System::set_block_number(System::current_block_number() + 1);
```
Chralt98, line 488: I would probably go with the following.
```suggestion
            System::set_block_number(System::current_block_number() + <Runtime as crate::Config>::DisputePeriod::get() - 1);
```

zrml/prediction-markets/src/migrations.rs
```diff
@@ -157,7 +167,141 @@ impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
...
+            frame_support::migration::storage_key_iter::<_, _, Twox64Concat>(
+                b"PredictionMarkets",
+                b"MarketIdsPerReportBlock",
+            );
+        market_ids_per_reported_iterator.try_for_each(
+            |(dispute_start_block, market_ids)| -> Result<(), &'static str> {
+                assert!(
+                    dispute_start_block > last_dp_end_block,
+                    "found unexpected storage key in MarketIdsPerReportBlock. \
+                     dispute_start_block: {:?}, last_dp_end_block: {:?}",
```
Chralt98, line 286: Please log the market ids here.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -157,7 +167,141 @@ impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
...
+                b"PredictionMarkets",
+                b"MarketIdsPerDisputeBlock",
+            );
+        market_ids_per_dispute_iterator.try_for_each(
+            |(dispute_start_block, market_ids)| -> Result<(), &'static str> {
+                assert!(
+                    dispute_start_block > last_dp_end_block,
+                    "found unexpected storage key in MarketIdsPerDisputeBlock. \
+                     dispute_start_block: {:?}, last_dp_end_block: {:?}",
+                    dispute_start_block,
```
Chralt98, line 258: Please log market ids here too.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -157,7 +167,141 @@ impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
...
+                    last_dp_end_block
+                );
 
+                market_ids.iter().try_for_each(|market_id| -> Result<(), &'static str> {
+                    let market = T::MarketCommons::market(market_id)
+                        .map_err(|_| "invalid market_id found.")?;
+                    assert!(
+                        market.status == MarketStatus::Reported,
```
Chralt98, line 295: I think this can also be `Disputed`. But I am not sure. @maltekliemann @sea212 ?

Chralt98 (approved):
Ok.

maltekliemann (changes requested):
One of the tests should be updated according to the comments (although it doesn't really matter due to the current state of Battery Station).

maltekliemann (approved):
:thumbsup:

zrml/prediction-markets/src/migrations.rs
```diff
@@ -157,7 +172,148 @@ impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
...
+            log::info!("Skipping storage cleanup; prediction-markets already up to date");
+            return total_weight;
+        }
+
+        total_weight = total_weight.saturating_add(T::DbWeight::get().reads(2));
+        log::info!("Starting storage cleanup of CleanUpStorageForResolvedOrClosedMarkets");
+
+        let dispute_period = T::DisputePeriod::get();
+        let current_block: T::BlockNumber = <frame_system::Pallet<T>>::block_number();
+        let last_dp_end_block = current_block.saturating_sub(dispute_period);
```
maltekliemann, line 198: As far as I can tell, the runtime upgrades trigger _before_ the block's `on_initialize` is called:  https://github.com/paritytech/substrate/blob/8e600ee0515122667cd091d632104708d343ca34/frame/executive/src/lib.rs#L292-L319

This means that in terms of disputing, the chain state lags one block behind during upgrades.

```suggestion
        let last_dp_end_block = current_block.saturating_sub(dispute_period).saturating_sub(1);
```

zrml/prediction-markets/src/migrations.rs
```diff
@@ -157,7 +172,148 @@ impl<T: Config> OnRuntimeUpgrade for MigrateMarketPoolsBeforeOpen<T> {
...
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade() -> Result<(), &'static str> {
+        Ok(())
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn post_upgrade() -> Result<(), &'static str> {
+        let dispute_period = T::DisputePeriod::get();
+        let current_block: T::BlockNumber = <frame_system::Pallet<T>>::block_number();
+        let last_dp_end_block = current_block.saturating_sub(dispute_period);
```
maltekliemann, line 252: As far as I can tell, the runtime upgrades trigger _before_ the block's `on_initialize` is called:  https://github.com/paritytech/substrate/blob/8e600ee0515122667cd091d632104708d343ca34/frame/executive/src/lib.rs#L292-L319

This means that in terms of disputing, the chain state lags one block behind during upgrades.

```suggestion
        let last_dp_end_block = current_block.saturating_sub(dispute_period).saturating_sub(1);
```

maltekliemann (changes requested):

zrml/prediction-markets/src/migrations.rs
vivekvpandya: Then should it be  `let last_dp_end_block = current_block.saturating_sub(dispute_period).saturating_add(1);` ?

maltekliemann: I might have made a mistake. My reasoning is this: If we're at the start of block `x`, then the last resolve we did was on block `x - 1`, so all markets that have been reported/disputed on block `x - 1 - DisputePeriod` have already been resolved. So these can be cleared. But we haven't resolved markets that have been reported/disputed on block `x - DisputePeriod`, because we havn't yet run `on_initialize` on block `x`. So `last_dp_end_block` must be `x - 1`.

Right...?

vivekvpandya: but from resolution_manager() we can see that for block X it will be resolved 
`let block = now.saturating_sub(dispute_period);` and from PR it will remove storage.

maltekliemann: Sorry, I'm not sure what you mean by "[...] and from PR it will remove storage." But just to emphasize my point: If you set `last_dp_end_block` to `x - 1 - DisputePeriod`, then on block `x`, the update will be executed before `on_initialize`, and the migration will clear the cache of all markets reported/disputed at block `x - 1 - DisputePeriod` or before. Suppose we have a market that was disputed on block `x - DisputePeriod`. This market's cache will _not_ be deleted. After the update, `on_initialize` will resolve the market and then remove it's cache.

Contrary, if you let `last_dp_end_block = x - DisputePeriod` and there's a market that was reported/disputed on block `x - DisputePeriod`, it will still be `Reported`/`Disputed` at the start of block `x`, and the market's cache will be deleted by the migration. When `on_initialize` is called for block `x`, it will no longer find this market in the cache and the market will never be resolved.

vivekvpandya: makes sense now!

maltekliemann (approved):
:thumbsup:

PR #733, opened 2022-08-04
Add minimum balance to CPMM logic

maltekliemann:


zrml/swaps/src/lib.rs
```diff
@@ -1181,6 +1190,42 @@ mod pallet {
...
+            amount: BalanceOf<T>,
+        ) -> DispatchResult {
+            let pool = Self::pool_by_id(pool_id)?;
+            // No need to prevent a clean pool from getting drained.
+            if pool.pool_status == PoolStatus::Clean {
+                return Ok(());
+            }
+            let pool_account = Self::pool_account_id(pool_id);
+            let balance = T::AssetManager::free_balance(asset, &pool_account);
+            let max_withdraw = balance.saturating_sub(Self::min_balance().saturated_into());
```
sea212, line 1224: Is this also a protection accounts account reaping on dust collection?
We don't have to take action to protect the pool account from being an object of dust collection, because it is on the `DustRemovalWhitelist` (in case the impl is correct):

https://github.com/zeitgeistpm/zeitgeist/blob/301a2a2c9ad48f150d8d7fb01593ea9b80ca5357/runtime/src/lib.rs#L1428-L1459


If you'd like to have an additional protection against dust collection anyways, I propose we give `min_balance` and addtional parameter of type `Asset` that we use to retrieve the `minimum_balance`.

Besides that I propose to specify a constant as the minimum pool amount value of any asset and the LP shares. It is transparent and not object to sudden changes in the `ExistentialDeposit` configuration.

maltekliemann: Oh, that's cool. I didn't know about this.

`min_balance()` wasn't exactly meant to be a protection against dust collection. In previous versions, I used a constant `MIN_BALANCE` and ensured that all balances are larger than this constant. However, having this constant doesn't really make sense when the existential deposit might be larger.

So, to protect against future changes to the minimum balance, I'll go with your suggestion of parametrizing the asset and pool in `min_balance`.

maltekliemann: Mitigated in a6b9f40.

zrml/swaps/src/lib.rs
```diff
@@ -1181,6 +1190,42 @@ mod pallet {
...
+        // The minimum allowed balance in a liquidity pool.
+        pub(crate) fn min_balance() -> BalanceOf<T> {
+            T::AssetManager::minimum_balance(Asset::Ztg)
+        }
+
+        pub(crate) fn ensure_minimum_liquidity_shares(
+            pool_id: PoolId,
+            amount: BalanceOf<T>,
+        ) -> DispatchResult {
+            let pool = Self::pool_by_id(pool_id)?;
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in fa8f5a8. No need to make the parameter optional.

zrml/swaps/src/lib.rs
```diff
@@ -1181,6 +1190,42 @@ mod pallet {
...
+            ensure!(amount <= max_withdraw, Error::<T>::PoolDrain);
+            Ok(())
+        }
+
+        pub(crate) fn ensure_minimum_balance(
+            pool_id: PoolId,
+            asset: Asset<T::MarketId>,
+            amount: BalanceOf<T>,
+        ) -> DispatchResult {
+            let pool = Self::pool_by_id(pool_id)?;
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in fa8f5a8. No need to make the parameter optional.

sea212 (approved):
Great!

zrml/swaps/src/benchmarks.rs
```diff
@@ -261,16 +260,8 @@ benchmarks! {
...
-            T::MinLiquidity::get()
-                - bmul(
-                    T::ExitFee::get().saturated_into(),
-                    T::MinLiquidity::get().saturated_into()
-                )?
-                .saturated_into();
-            a as usize
-        ];
+        let pool_amount = T::MinLiquidity::get() / 2u32.into();
+        let min_assets_out = vec![0u32.into(); a as usize];
```
vivekvpandya, line 264: Not able to figure out why this change?

vivekvpandya (approved):
Lots of new tests! :)

zrml/swaps/src/benchmarks.rs
maltekliemann: Yeah, good question!

First of all, the `min_assets_out` parameter has no influence on the benchmarks. So, in order to make sure that no `LimitOut` error is thrown, I just set it to zero. This is simply because its might be somewhat annoying to figure out how much exactly you're getting out, and there's no point, because we're not doing real trading here.

Also, I set the amount of liquidity I'm withdrawing to half of what's in the pool. Before this change, the entire pool got drained, which, given the state of the pool, was not allowed.

vivekvpandya: thanks for explanation!

PR #734, opened 2022-08-04
Fix rust toolchain install

maltekliemann:


.github/workflows/rust.yml
```diff
@@ -45,16 +45,19 @@ jobs:
...
-          rustup show
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: rustfmt, clippy
+          target: wasm32-unknown-unknown
+          default: true
+          override: true
```
sea212, line 56: This is definitely a more standard way to install rust during the pipeline. I think we should use this approach in the `tests` job above as well.

sea212 (approved):

PR #744, opened 2022-08-10
Allow market approval by single `AdvisoryCommittee` member

### LINKED ISSUE #741: [PM] A single member should be able to approve markets
**Problem**
Voting to approve markets can clog the queue

**Potential solution**
Make approval something a single member of the `AdvisoryCommittee` can do (thanks for that suggestion @maltekliemann )
An approval should also be for free for the approving account, just as the first vote for motions in the `AdvisoryCommittee` is.

**Description**
After an `advised` market is created, the `AdvisoryCommittee` has to `approve` or `reject` that market.
Currently a motion within the `AdvisoryCommittee` collective is launched to vote on `approval` or `rejection`. A simple majority is required to `approve`. Regarding the sheer number of markets and the fact that motions run for more than a day, this could lead to the situation that most of the markets cannot be approved in time if not enough votes come in. In addition, approving a market does not have any consequences besides the market being accepted to launch, whereas on the other hand rejecting a market leads to the slashing of the market creators bond. Consequently `reject` should remain a decision that at least a simple majority agrees upon, but `approve` can be something that a single member can decide.

sea212:
Closes #741 

Also makes any privileged origin not have to pay fees for their service.


sea212:
In regards to tests: Testing if the origin is the correct one is out of scope of unit tests. This requires functional testing or acceptance testing.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1064,17 +1084,13 @@ mod pallet {
...
-            Self::AccountId,
-            CurrencyId = Asset<MarketIdOf<Self>>,
-            ReserveIdentifier = [u8; 8],
-        >;
-
         /// The module identifier.
         #[pallet::constant]
         type PalletId: Get<PalletId>;
 
+        /// The origin that is allowed to approve / reject pending advised markets.
```
maltekliemann, line 1091: Similar fix needs to be applied to the docs of `ApproveOrigin`.

```suggestion
        /// The origin that is allowed to reject pending advised markets.
```

maltekliemann (changes requested):
You may also want to change the documentation at the top of `lib.rs`. Looks good otherwise, there's only one problem. I tested this on a local dev, and it just doesn't seem to work. I sign these three transactions (the last one with Dave's key, of course):

```js
api.tx.sudo.sudo(api.tx.advisoryCommitteeMembership.addMember(DAVE));
api.tx.predictionMarkets.createMarket(
    EVE,
    api.createType("MarketPeriod", { Block: [0, 10000] }),
    api.createType("MultiHash", { Sha3_384: METADATA }),
    "Advised",
    api.createType("MarketType", { Categorical: 3 }),
    api.createType("MarketDisputeMechanism", { Authorized: EVE }),
    "CPMM"
);
api.tx.predictionMarkets.approveMarket(0);
```

The last transaction returns `BadOrigin`. Any explanation why this happens? I'm assuming I'm just doing something completely wrong here, but I have no idea what.

sea212:
> The last transaction returns BadOrigin. Any explanation why this happens?

Yes, you have to call `predictionMarkets.approve` through the `advisoryCommittee.propose` dispatchable, with `threshold = 1`. If the threshold is `1`, it will essentially act like the caller immediately voted `yes` and execute the proposal.

maltekliemann (approved):
:thumbsup:

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -464,7 +464,7 @@ benchmarks! {
             MarketType::Categorical(T::MaxCategories::get()),
             ScoringRule::CPMM
         )?;
-        let approval_origin = T::ApprovalOrigin::successful_origin();
+        let approval_origin = T::ApproveOrigin::successful_origin();
```
Chralt98, line 467: So this works, because `ApproveOrigin` includes the origins of `RejectOrigin`. Maybe use `RejectOrigin` to avoid the need of this background information.

zrml/prediction-markets/src/lib.rs
```diff
@@ -984,7 +1000,14 @@ mod pallet {
...
         /// The origin that is allowed to approve / reject pending advised markets.
-        type ApprovalOrigin: EnsureOrigin<Self::Origin>;
+        type ApproveOrigin: EnsureOrigin<Self::Origin>;
+
+        /// Shares of outcome assets and native currency
+        type AssetManager: ZeitgeistAssetManager<
+            Self::AccountId,
+            CurrencyId = Asset<MarketIdOf<Self>>,
+            ReserveIdentifier = [u8; 8],
+        >;
```
Chralt98, line 1010: I see, my fault. I need to learn this lexicographical order now :-D.

zrml/prediction-markets/src/lib.rs
```diff
@@ -215,22 +220,29 @@ mod pallet {
...
+                        total_accounts.saturated_into(),
+                        share_accounts.saturated_into(),
+                        outcome_assets_amount.saturated_into(),
+                    )),
+                    Pays::No,
+                )
+                    .into())
             } else {
-                Ok(None.into())
+                Ok((None, Pays::No).into())
```
Chralt98, line 245: Early return errors still lead to fees as I understand it correctly? Because you could have used `Pays::No` in the `[pallet::weight]` annotation as default.

Chralt98 (commented):
The advisory committee can now fire approves with lightning speed 🚀 . If for example a member of the collective approves malicious markets, this member should be kicked out of the advisory committee then.

> RejectOrigin rejects market \n or approval period passes

The `docs/diagrams/svg/market_state_diagram.svg` is not updated according to the `docs/diagrams/src/market_state_diagram.puml`.

zrml/prediction-markets/src/benchmarks.rs
sea212: You're right, we should use the respective origins. Mitigated in [d267583](https://github.com/zeitgeistpm/zeitgeist/pull/744/commits/d26758328e49a4d1ff8616557af79f9affae756b).

zrml/prediction-markets/src/lib.rs
sea212: :fire: Great catch :fire: Mitigated in [da863bd](https://github.com/zeitgeistpm/zeitgeist/pull/744/commits/da863bddcd6f90ac7adfed483e8865d81437f49a)

Chralt98 (approved):
👍

maltekliemann (approved):
:thumbsup:

PR #750, opened 2022-08-11
Optimize default build by excluding fuzz tests

sea212:
Previously the workspace configuration lead to fuzz tests being build in the default case (`cargo build` in root). This also lead to the activation of the `mock` feature, which in turn builds the dev-dependencies of all Zeitgeist pallets.
The fuzz tests have been removed from the default build command, leading to a reduction of crates that have to be compiled from 939 to 911 when using `cargo build`


maltekliemann (approved):
Yeah, that's good stuff. We might also want to disable building the fuzz tests when running `cargo test` in individual pallets.

Chralt98 (approved):
OK, `default_members` is a subset of `members`. [Cargo Workspaces Reference](https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection).

PR #755, opened 2022-08-15
Zrml currencies

vivekvpandya:


zrml/currencies/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    impl<T: Config> ZeitgeistAssetManager<T::AccountId> for Pallet<T> {
+        fn accounts_by_currency_id(
+            currency_id: Self::CurrencyId,
+        ) -> Result<(usize, Vec<T::AccountId>), DispatchError> {
+            let accounts: Vec<T::AccountId> =
+                Accounts::<T>::iter_prefix(currency_id).map(|(k2, _v)| k2).collect();
+            Ok((accounts.len(), accounts))
+        }
```
Chralt98, line 274: **Overall:** I mean: what is the point of using another storage double map, when we iterate through the whole original map [here ](https://github.com/zeitgeistpm/zeitgeist/blob/89e36c13995135edbf6ac6b2cc9e025051c3ce44/primitives/src/traits/zeitgeist_multi_reservable_currency.rs#L53-L63) anyways? Or is the plan to remove the whole iterations of https://github.com/zeitgeistpm/zeitgeist/blob/89e36c13995135edbf6ac6b2cc9e025051c3ce44/primitives/src/traits/zeitgeist_multi_reservable_currency.rs#L45-L64 ? And then we only use `zrml_currencies` in the future.

zrml/prediction-markets/src/mock.rs
```diff
@@ -83,6 +83,7 @@ construct_runtime!(
         Balances: pallet_balances::{Call, Config<T>, Event<T>, Pallet, Storage},
         Court: zrml_court::{Event<T>, Pallet, Storage},
         AssetManager: orml_currencies::{Call, Pallet, Storage},
+        ZrmlCurrencies: zrml_currencies::{Call, Pallet, Storage, Event<T>},
```
Chralt98, line 86:
```suggestion
        ZrmlCurrencies: zrml_currencies::{Pallet, Storage},
```

There are no dispatchables or events in the pallet.

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
```diff
@@ -29,13 +29,13 @@ pub trait ZeitgeistAssetManager<AccountId>: NamedMultiReservableCurrency<Account
     /// If the `currency_id` is the native currency, then return None.
     fn accounts_by_currency_id(
         currency_id: Self::CurrencyId,
-    ) -> Result<(usize, Vec<(AccountId, AccountData<Self::Balance>)>), DispatchError>;
+    ) -> Result<(usize, Vec<AccountId>), DispatchError>;
```
Chralt98, line 32: When you delete this function, it's much easier to see, if you really looked at every code dependency references. (If the overall plan is it to remove this, as I think it is).

zrml/currencies/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+            T::Currencies::transfer(currency_id, from, to, amount)
+        }
+        fn deposit(
+            currency_id: Self::CurrencyId,
+            who: &T::AccountId,
+            amount: Self::Balance,
+        ) -> DispatchResult {
+            Accounts::<T>::insert(currency_id, who, ());
+            T::Currencies::deposit(currency_id, who, amount)
+        }
```
Chralt98, line 242: We don't like to save the information for `ZTG`, the main asset. I think.. ?

Chralt98 (commented):
The main good approach is to limit the number of storage reads and writes for the `ZeitgeistAssetManager`. This can be done with a wrapper as you implemented. We need to keep in mind, that it's extremely heavy to read and write unlimited number of accesses.

At the moment I don't see any other way to delete the unimportant information of resolved outcome asset (we have to clear the `Accounts` storage map anyways).

We have four currency modules then. `pallet_balances` as `Balances`, `orml_currencies` as `AssetManager`, `orml_tokens` as `Tokens`, `zrml_currencies` as `Currencies`. That's really confusing. Maybe we should name `orml_currencies` as `Currencies` and our currency class as `AssetManager`.
Or as you proposed, which I like too: `OrmlCurrencies` and `ZrnlCurrencies`.

zrml/currencies/src/tests.rs
```diff
@@ -0,0 +1,56 @@
...
+use proptest::{collection::hash_set, prelude::*, strategy::ValueTree, test_runner::TestRunner};
+use zeitgeist_primitives::{
+    traits::ZeitgeistAssetManager,
+    types::{AccountIdTest, Asset, CurrencyId, ScalarPosition},
+};
+
+proptest! {
+    #![proptest_config(ProptestConfig::with_cases(20))]
+    #[test]
+    fn validate_storages( _i : u32) {
```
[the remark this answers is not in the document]
vivekvpandya: Yes other wise proptest macro gets confused and can not compile.

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
```diff
@@ -44,26 +44,30 @@ where
...
     fn accounts_by_currency_id(
         currency_id: Self::CurrencyId,
-    ) -> Result<(usize, Vec<(T::AccountId, AccountData<Self::Balance>)>), DispatchError> {
+    ) -> Result<(usize, Vec<T::AccountId>), DispatchError> {
         let mut total = 0;
         #[allow(
             // Iterator will never yield more than `usize::MAX` elements
             clippy::integer_arithmetic
         )]
         let accounts = <Accounts<T>>::iter()
```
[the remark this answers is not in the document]
vivekvpandya: But after this PR it will be used only for testing.

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
```diff
@@ -44,26 +44,30 @@ where
...
     }
 
     fn destroy_all<I>(currency_id: Self::CurrencyId, accounts: I) -> Result<(), DispatchError>
     where
-        I: Iterator<Item = (T::AccountId, AccountData<Self::Balance>)>,
+        I: Iterator<Item = T::AccountId>,
     {
-        for (k0, _) in accounts {
+        for k0 in accounts {
             <Accounts<T>>::remove(k0, currency_id);
```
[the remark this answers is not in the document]
vivekvpandya: Yes this will be used, but I don't see any other option.

zrml/currencies/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    impl<T: Config> ZeitgeistAssetManager<T::AccountId> for Pallet<T> {
+        fn accounts_by_currency_id(
+            currency_id: Self::CurrencyId,
+        ) -> Result<(usize, Vec<T::AccountId>), DispatchError> {
+            let accounts: Vec<T::AccountId> =
+                Accounts::<T>::iter_prefix(currency_id).map(|(k2, _v)| k2).collect();
+            Ok((accounts.len(), accounts))
```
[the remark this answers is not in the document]
vivekvpandya: Agreed.

zrml/currencies/src/lib.rs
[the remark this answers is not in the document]
vivekvpandya: In StorageMap use of Vec is not recommended.

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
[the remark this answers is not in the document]
vivekvpandya: that seems feasible, IIUC you are recommending to have just one method which deletes storage from ORML token while iterating over the storage from wrapper.

zrml/currencies/src/lib.rs
vivekvpandya: yes only remove_all will be used from orml_token, accounts_by.. will be used from zrml_currencies

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
vivekvpandya: I am using this in test.

zrml/currencies/src/lib.rs
vivekvpandya: ZTG will not be handled by fungibles I think, it goes via fungible.

zrml/currencies/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    #[pallet::storage]
+    pub type Accounts<T: Config> = StorageDoubleMap<
+        _,
+        Blake2_128Concat,
+        CurrencyIdOf<T>,
+        Blake2_128Concat,
+        T::AccountId,
+        (),
+        OptionQuery,
+    >;
```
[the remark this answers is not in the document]
vivekvpandya: As per current code, in this double map I can't keep ValueQuery behavior otherwise every query will return () and that is not what I want.

zrml/currencies/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+            let accounts: Vec<T::AccountId> =
+                Accounts::<T>::iter_prefix(currency_id).map(|(k2, _v)| k2).collect();
+            Ok((accounts.len(), accounts))
+        }
+
+        fn destroy_all<I>(currency_id: Self::CurrencyId, accounts: I) -> Result<(), DispatchError>
+        where
+            I: Iterator<Item = T::AccountId>,
+        {
+            Accounts::<T>::remove_prefix(currency_id, None);
```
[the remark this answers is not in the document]
vivekvpandya: :thinking:

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
```diff
@@ -44,26 +44,30 @@ where
...
             // Iterator will never yield more than `usize::MAX` elements
             clippy::integer_arithmetic
         )]
         let accounts = <Accounts<T>>::iter()
-            .filter_map(|(k0, k1, v)| {
-                total += 1;
-                if k1 == currency_id { Some((k0, v)) } else { None }
+            .filter_map(|(k0, k1, _v)| {
+                if k1 == currency_id {
+                    total += 1;
```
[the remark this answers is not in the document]
vivekvpandya: there is a mistake from me. I will revert this change.

zrml/currencies/src/lib.rs
Chralt98: `orml_tokens` does not handle `ZTG`, but `orml_currencies` does. https://github.com/zeitgeistpm/zeitgeist/blob/48ba93bb12cd0520c0ee0c878dc9ca0446b9454d/primitives/src/traits/zeitgeist_multi_reservable_currency.rs#L84-L88

vivekvpandya: I see , we should add similar code in zrml currencies too.

zrml/currencies/src/lib.rs
Chralt98: Avoid using `.collect()`, because this requires an expensive heap allocation. Rather return the Iterator itself.
https://nnethercote.github.io/perf-book/iterators.html

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
Chralt98:
```bash
grep -rI --exclude-dir="./target" "accounts_by_currency_id" *
```

primitives/src/traits/zeitgeist_multi_reservable_currency.rs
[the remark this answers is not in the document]
Chralt98: I realised, that you can throw an iterator in `destroy_all`, so that you can divide the work over multiple blocks if needed. The `TotalIssuance` is removed right at the first call of `destroy_all`. Hopefully no bad consequences with that.

So in swaps and pm-pallet the `accounts_by_currency_id` is called, in which a vector is returned at the moment. and after that the accounts are destroyed by `destroy_all`. But at some point the vector is too large. So my suggestion is, that we should return an iterator for `accounts_by_currency_id` and should delegate the work for the calls [here](https://github.com/zeitgeistpm/zeitgeist/blob/48ba93bb12cd0520c0ee0c878dc9ca0446b9454d/zrml/prediction-markets/src/lib.rs#L1695-L1701) for example to some kind of multiple block deletion. 

https://github.com/zeitgeistpm/zeitgeist/blob/48ba93bb12cd0520c0ee0c878dc9ca0446b9454d/zrml/prediction-markets/src/lib.rs#L1695-L1701

This above is a time bomb.

zrml/currencies/src/lib.rs
vivekvpandya: it seems I should be using drain_prefix() as it provides iterator and removes the key values as well so , I could avoid allocation required while .collect()

zrml/swaps/src/lib.rs
```diff
@@ -1578,8 +1578,9 @@ mod pallet {
             let pool_share_id = Self::pool_shares_id(pool_id);
             let (_, liquidity_providers) =
                 T::AssetManager::accounts_by_currency_id(pool_share_id).unwrap_or((0usize, vec![]));
```
Chralt98, line 1580:
```suggestion
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1758,29 +1750,20 @@ mod pallet {
...
-                            let (total_accounts, accounts) =
-                                T::AssetManager::accounts_by_currency_id(asset)
-                                    .unwrap_or((0usize, vec![]));
-                            total_asset_accounts =
-                                total_asset_accounts.saturating_add(accounts.len());
-
-                            let _ = T::AssetManager::destroy_all(asset, accounts.iter().cloned());
-                            total_accounts
+                            let _ = T::AssetManager::destroy_all(asset);
+                            0
```
Chralt98, line 1754:
```suggestion
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1758,29 +1750,20 @@ mod pallet {
...
                         } else {
                             0
                         }
```
Chralt98, line 1757:
```suggestion
                        }
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1758,29 +1750,20 @@ mod pallet {
...
                     };
 
-                    if let Some(first_asset) = assets_iter.next() {
-                        total_accounts = manage_asset(first_asset, winning_asset_idx);
-                    }
                     for asset in assets_iter {
                         let _ = manage_asset(asset, winning_asset_idx);
```
Chralt98, line 1761:
```suggestion
                        manage_asset(asset, winning_asset_idx);
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -199,14 +197,10 @@ mod pallet {
 
             // Delete market's outcome assets, clear market and delete pool if necessary.
             let mut destroy_asset = |asset: Asset<_>| -> Option<usize> {
```
Chralt98, line 199:
```suggestion
            let mut destroy_asset = |asset: Asset<_>|  {
```

Chralt98 (commented):
Removing one benchmark component properly is a big plus.

zrml/currencies/src/tests.rs
```diff
@@ -0,0 +1,56 @@
...
+        let deposit_accounts = deposit_input.new_tree(&mut runner).unwrap().current();
+        let transfer_accounts = transfer_input.new_tree(&mut runner).unwrap().current();
+        let currency_id : CurrencyId = Asset::ScalarOutcome(0, ScalarPosition::Short);
+        let balance =  100_u32;
+        for (account_a, account_b) in deposit_accounts.iter().zip(transfer_accounts.iter()) {
+            <Currencies as MultiCurrency<AccountIdTest>>::deposit(currency_id, account_a, balance.into()).expect("deposit failed");
+            // transfer half balance to account_b
+            <Currencies as MultiCurrency<AccountIdTest>>::transfer(currency_id, account_a, account_b, (balance/2).into()).expect("deposit failed");
+            let account_from_currencies_zrml = <Currencies as ZeitgeistAssetManager<AccountIdTest>>::accounts_by_currency_id(currency_id).expect("accounts_by_currency_id failed");
+            let account_from_currencies_orml = <OrmlCurrencies as ZeitgeistAssetManager<AccountIdTest>>::accounts_by_currency_id(currency_id).expect("accounts_by_currency_id failed");
```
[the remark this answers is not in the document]
vivekvpandya: zrml_currencies may have account noted in storage which is no longer present in orml storage. of course not as per current code of the test but I can try making such variation.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -371,24 +350,20 @@ benchmarks! {
...
-        // a = total accounts
-        let a in 0..10;
         // b = num. accounts with assets
         let b in 0..10;
         // c = num. asset types
         let c in (T::MinCategories::get().into())..T::MaxCategories::get().into();
 
         let c_u16 = c.saturated_into();
-        let (_, market_id) = setup_resolve_common_categorical_after_dispute::<T>(a, b, c_u16)?;
+        let (_, market_id) = setup_resolve_common_categorical_after_dispute::<T>(b, c_u16)?;
```
Chralt98, line 359: I would expect a `Reported` market here. I would assume, that the market is `Disputed`. Maybe add a benchmark `verify` check at the end of this benchmark to check for the right status.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -371,24 +350,20 @@ benchmarks! {
...
     }: {
         let market = T::MarketCommons::market(&market_id)?;
         let disputes = crate::Disputes::<T>::get(&market_id);
         T::SimpleDisputes::on_resolution(&disputes, &market_id, &market)?
```
Chralt98, line 363: I would expect to call `PredictionMarket::on_resolution` here.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -397,7 +372,7 @@ benchmarks! {
         let d in 0..T::MaxDisputes::get();
 
         let c_u16 = c.saturated_into();
-        let (caller, market_id) = setup_resolve_common_categorical_after_dispute::<T>(a, b, c_u16)?;
+        let (caller, market_id) = setup_resolve_common_categorical_after_dispute::<T>(b, c_u16)?;
```
Chralt98, line 375: I would expect to benchmark `PedictionMarket::on_resolution` here (not `SimpleDisputes::on_resolution`).

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -414,7 +389,7 @@ benchmarks! {
     internal_resolve_scalar_reported {
         let total_accounts = 10u32;
         let asset_accounts = 10u32;
-        let (_, market_id) = setup_resolve_common_scalar_after_dispute::<T>(total_accounts, asset_accounts)?;
+        let (_, market_id) = setup_resolve_common_scalar_after_dispute::<T>(asset_accounts)?;
     }: {
         let market = T::MarketCommons::market(&market_id)?;
         let disputes = crate::Disputes::<T>::get(&market_id);
```
Chralt98, line 395: I would expect a benchmark of `PredictionMarket::on_resolution` instead of `T::SimpleDisputes` here. In addition to that I would expect the status `Reported` (because of the benchmark name `internal_resolve_scalar_reported`), but a dispute already happened with `setup_resolve_common_scalar_after_dispute`

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -426,7 +401,7 @@ benchmarks! {
         let asset_accounts = 10u32;
         let d in 0..T::MaxDisputes::get();
 
-        let (caller, market_id) = setup_resolve_common_scalar_after_dispute::<T>(total_accounts, asset_accounts)?;
+        let (caller, market_id) = setup_resolve_common_scalar_after_dispute::<T>(asset_accounts)?;
```
Chralt98, line 404: Same here: `SimpleDisputes` instead of `PredictionMarket::on_resolution`.

Chralt98 (commented):
Only a comment on the benchmarks of `on_resolution`. At the moment the weight calculation of `on_resolution` is incorrect. I would propose, that you write a benchmark for `PredictionMarkets::on_resolution` in general. Then you could ignore the weights of the sub functions. The problem we are constantly running into is, that whenever there is some addition to `on_resolution`, that is not a function with a weight return,  the weight is most likely not properly added.

zrml/currencies/src/lib.rs
```diff
@@ -273,12 +273,33 @@ pub mod pallet {
...
         {
-            Accounts::<T>::remove_prefix(currency_id, None);
-            T::Currencies::destroy_all(currency_id, accounts)
+            // Accounts::<T>::remove_prefix(currency_id, None);
+            // T::Currencies::destroy_all(currency_id, accounts)
+            let mut accounts = 0_usize;
+            for (account, _) in Accounts::<T>::drain_prefix(currency_id) {
+                accounts += 1;
+                T::Currencies::remove(currency_id, account)?;
+            }
```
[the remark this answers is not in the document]
vivekvpandya: okay and then remaining accounts will be removed by garbage collector proposed by https://github.com/zeitgeistpm/zeitgeist/issues/792 ? if so then we can do this suggested change when we implement GC?

sea212:
We are not using pallet-assets and garbage collection is low priority. In case we need this in the future, we'll reuse code within this.

PR #758, opened 2022-08-16
Remove old migrations

sea212:
Removes old migrations. I left the utilities in the prediction market migration to modify swap pools commented out, seems to me we might need them again in the future.


maltekliemann (approved):
:+1:

PR #759, opened 2022-08-17
Update github workflow to upload test coverage reports

### LINKED ISSUE #262: [Analytics] Create and commit coverage reports within workflows (overarching issue)
Depends on #263 and #264 

**TODO**
- [x] #263 
- [x] #264

### LINKED ISSUE #263: [Analytics] Extend test workflow to generate and commit coverage reports
[The workflow that contains the job to execute the tests](https://github.com/zeitgeistpm/zeitgeist/blob/51c79db9f44a6d8491b2b398b911466fa32e58df/.github/workflows/rust.yml) should be extended to [collect test coverage data](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/instrument-coverage.html#test-coverage) during the *test* job. Another step within the *test* job should [create a coverage report](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/instrument-coverage.html#creating-coverage-reports) for the test coverage data. Another job, that depends on the *test* job, should finally commit the reports into a specific folder within the project.

### LINKED ISSUE #264: [Analytics] Extend fuzz workflow to generate and commit coverage reports
[The workflow that contains the job to execute the fuzz tests](https://github.com/zeitgeistpm/zeitgeist/blob/51c79db9f44a6d8491b2b398b911466fa32e58df/.github/workflows/rust.yml) should be extended to [collect test coverage data](https://rust-fuzz.github.io/book/cargo-fuzz/coverage.html#generate-code-coverage-data) during the *fuzz* job. Another step within the *fuzz* job should [create a coverage report](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/instrument-coverage.html#creating-coverage-reports) for the fuzz test coverage data. Another job, that depends on the *fuzz* job, should finally commit the reports into a specific folder within the project.

Chralt98:
Fixes #262. Fixes #263. Fixes #264.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,11 +33,28 @@ jobs:
           default: true
           override: true
           profile: minimal
-          components: rustfmt, clippy
+          components: rustfmt, clippy, llvm-tools-preview
+
+      - name: Download grcov
+        run: |
+          mkdir -p "${HOME}/.local/bin"
+          curl -sL https://github.com/mozilla/grcov/releases/download/v0.8.11/grcov-x86_64-unknown-linux-gnu.tar.bz2 | tar jxf - -C "${HOME}/.local/bin"
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+      
+      - name: Run xtask coverage
+        uses: actions-rs/cargo@v1
+        with:
+          command: xtask
+          args: coverage
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v3
+        with:
+          files: coverage/*.lcov
 
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@v1
-        
+
       - name: Tests
         run: ./scripts/tests/${{ matrix.scripts }}.sh
 
@@ -56,11 +73,28 @@ jobs:
           default: true
           override: true
           profile: minimal
-          components: rustfmt, clippy
+          components: rustfmt, clippy, llvm-tools-preview
 
       - uses: actions-rs/install@v0.1
         with:
           crate: cargo-fuzz
           use-tool-cache: true
+
+      - name: Download grcov
+        run: |
+          mkdir -p "${HOME}/.local/bin"
+          curl -sL https://github.com/mozilla/grcov/releases/download/v0.8.11/grcov-x86_64-unknown-linux-gnu.tar.bz2 | tar jxf - -C "${HOME}/.local/bin"
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+      
+      - name: Run xtask coverage
+        uses: actions-rs/cargo@v1
+        with:
+          command: xtask
+          args: coverage
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v3
+        with:
+          files: coverage/*.lcov
 
       - run: ./scripts/tests/fuzz.sh


Chralt98:
@sea212 @lsaether Is it okay to use and sign up on https://docs.codecov.com/docs#step-1-sign-up-for-codecov ? I have no admin access to the repository to integrate Codecov.

vivekvpandya (approved):
It looks fine to me. I hope total turn around time is not increased much. Also I am wondering why grcov is used instead of llvm-cov. llvm-cov is available and it makes commands much simple with this cargo wrapper https://github.com/taiki-e/cargo-llvm-cov
however I am not able to find any comparison of grcov vs llvm-cov , only thing is that grcov is written in rust.

maltekliemann:
Just a dumb question: The report says 92.37% coverage. What exactly does that number mean? There seem to be two different workflows running, one for unit tests and one for fuzz tests, so I would have expected at least two numbers... right?

Also, doesn't this number seems surprisingly high? Can we get a complete list of lines of code reached/not reached?

.github/workflows/rust.yml
```diff
@@ -33,14 +33,25 @@ jobs:
...
+          curl -sL https://github.com/mozilla/grcov/releases/download/v0.8.11/grcov-x86_64-unknown-linux-gnu.tar.bz2 | tar jxf - -C "${HOME}/.local/bin"
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
       - name: Tests
         run: ./scripts/tests/${{ matrix.scripts }}.sh
 
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v3
+        with:
+          files: /misc/tests.lcov
```
sea212, line 53: Might want to use */tmp/tests.lcov* instead to avoid potential permission issues.

sea212 (commented):
Does CodeCov still need access to our repository?

Chralt98:
> Does CodeCov still need access to our repository?

That's a good question. I don't really know. I would assume no, because Codecov could comment on this PR, so I think it has been already added by Logan.

.github/workflows/rust.yml
```diff
@@ -33,14 +33,25 @@ jobs:
...
 
       - name: Cache Dependencies
         uses: Swatinem/rust-cache@v1
-        
+
+      - name: Download grcov
+        run: |
+          mkdir -p "${HOME}/.local/bin"
+          curl -sL https://github.com/mozilla/grcov/releases/download/v0.8.11/grcov-x86_64-unknown-linux-gnu.tar.bz2 | tar jxf - -C "${HOME}/.local/bin"
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
```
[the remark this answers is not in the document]
Chralt98: I tried it. When running the `grcov` github action it always returns the following error.

```text
Error: Unable to find any coverage files, was `cargo test` executed correctly?
```

Although I executed `cargo test` as step before the `grcov` github action. I don't really know how to solve this.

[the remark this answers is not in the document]
Chralt98: I couldn't find an example for grcov GH action for source-based coverage.

.github/workflows/rust.yml
Chralt98: addressed in https://github.com/zeitgeistpm/zeitgeist/pull/759/commits/eb06f3518ad20aecf7150f32c76b6bb2b8d17435

scripts/tests/fuzz.sh
```diff
@@ -33,35 +33,38 @@ RIKIDDO_WITH_CALCULATED_FEE_FACT=1750
...
+CARGO_INCREMENTAL=0 RUSTFLAGS='-Cinstrument-coverage' LLVM_PROFILE_FILE='cargo-fuzz-%p-%m.profraw' RUST_BACKTRACE=1 cargo fuzz run --release --fuzz-dir zrml/rikiddo/fuzz rikiddo_with_initial_fee -- -runs=$(($(($RUNS * $RIKIDDO_WITH_INITIAL_FEE_FACT)) / $BASE))
+CARGO_INCREMENTAL=0 RUSTFLAGS='-Cinstrument-coverage' LLVM_PROFILE_FILE='cargo-fuzz-%p-%m.profraw' RUST_BACKTRACE=1 cargo fuzz run --release --fuzz-dir zrml/rikiddo/fuzz rikiddo_with_calculated_fee -- -runs=$(($(($RUNS * $RIKIDDO_WITH_CALCULATED_FEE_FACT)) / $BASE))
 # This actually needs approx. 107 seconds. Need to find a way to optimize fuzzing on-chain
-RUST_BACKTRACE=1 cargo fuzz run --release --fuzz-dir zrml/rikiddo/fuzz rikiddo_pallet -- -runs=$(($(($RUNS * $RIKIDDO_PALLET_FACT)) / $BASE))
\ No newline at end of file
+CARGO_INCREMENTAL=0 RUSTFLAGS='-Cinstrument-coverage' LLVM_PROFILE_FILE='cargo-fuzz-%p-%m.profraw' RUST_BACKTRACE=1 cargo fuzz run --release --fuzz-dir zrml/rikiddo/fuzz rikiddo_pallet -- -runs=$(($(($RUNS * $RIKIDDO_PALLET_FACT)) / $BASE))
+
+mkdir tmp
```
sea212, line 69: Just to be on the safe side, ensure that `mkdir` does not fail when the folder already exists:
```suggestion
mkdir -p tmp
```

aside from that, is there a reason why we don't just use */tmp*? We assume execution on Unix anyways and there the folder is present, ready to help us achieve exactly what we need here :)

scripts/tests/misc.sh
```diff
@@ -23,3 +23,6 @@ do
     echo "TEST $package std,runtime-benchmarks"
   fi
 done
+
+mkdir tmp
```
sea212, line 27: Just to be on the safe side, ensure that `mkdir` does not fail when the folder already exists:
```suggestion
mkdir -p tmp
```

sea212 (approved):

scripts/tests/fuzz.sh
Chralt98: I have this failed workflow here https://github.com/zeitgeistpm/zeitgeist/actions/runs/3037854786/jobs/4890827497 

But I think, it's because we are in the `zeitgeist` folder and not in the Unix root. So you would like to save the lcov file in the Unix tmp folder?

Chralt98: I use the temporary folder of the Github Actions Runner. https://github.com/zeitgeistpm/zeitgeist/pull/759/commits/2db1628ca948c9f087c4a1b0073b3a30f90ea9f7

Chralt98:
Bildschirmfoto 2022-09-13 um 12 18 11

@lsaether @sea212 Codecov is currently requested from me for @zeitgeistpm. Hope you accept the request to allow automatic code coverage reports.

scripts/tests/misc.sh
Chralt98: addressed in https://github.com/zeitgeistpm/zeitgeist/pull/759/commits/c1d8c5bd9eb2ee86413843aa3c55b4e789557bb9

Chralt98:
@mkl 

> The report says 92.37% coverage. What exactly does that number mean?

The problem from the Codecov report is, that there exists no `codecov report` **for the main branch**. That's why we can't see the diff for the message. I also realised, that the Codecov message updates (with github edits). It does not come again as a Github message.

> There seem to be two different workflows running, one for unit tests and one for fuzz tests, so I would have expected at least two numbers... right?

That's right. That is real problem. I need to fix that. You are right, we should see two codecov reports.

> Also, doesn't this number seem surprisingly high? Can we get a complete list of lines of code reached/not reached?

This number seems correct. 

Complete list here https://codecov.io/gh/zeitgeistpm/zeitgeist/tree/c1d8c5bd9eb2ee86413843aa3c55b4e789557bb9

scripts/tests/aux-functions.sh
```diff
@@ -15,5 +15,6 @@ test_package_with_feature() {
     local features=$2
 
     /bin/echo -e "\e[0;33m***** Testing '$package' with features '$features' *****\e[0m\n"
-    cargo test --features $features --manifest-path $package/Cargo.toml --no-default-features --release
```
Chralt98, line 18: We should avoid profile release for tests, so that `debug_assert!` macros are activated and we can see, that something is not as expected.

Chralt98:
I tried https://github.com/taiki-e/cargo-llvm-cov out @vivekvpandya . It had exactly the same result as we got with the current solution. Thanks for your investigation! I think it's fine as it is. With using `cargo-llvm-cov` (because it's a wrapper around cargo test) it's not so clear what actually happens (so the future maintainers would need to look up `cargo-llsm-cov` to understand that it's basically cargo test with code coverage).

Chralt98:
We got a slightly better time for the current fuzz workflow. It's now `56` minutes instead of on the main branch `66` minutes. This is because of the `--release` profile for the swaps fuzz tests. Just in case if somebody asks about if compile time with `dev` (`debug`) beats the speed of `--release`.

sea212 (approved):

vivekvpandya (approved):

PR #765, opened 2022-08-19
Add missing license files

maltekliemann:
The citation of Acala got gobbled up during a merge. Also, I discovered some code in `node/` that came from Polkadot (`ExecuteWithClient`).


sea212 (commented):
I think the attribution is a bit overkill.
As per [a discussion on stackexchange](https://opensource.stackexchange.com/questions/4577/does-gplv3-require-attribution), it seems to be sufficient to replace our copyright statement with:
> Copyright 2017-2020 Parity Technologies (UK) Ltd.
> Modifications Copyright 2021-2022 Zeitgeist PM LLC.

The license is the same, GPL3, and the notice is already included, so no need to include that.
We also have to include a changelog as it seems. I am not sure if this applies if we provide a git project, since it is included inherently, but to be safe we might want to add something like this:
> Changes to the original file: https://github.com/zeitgeistpm/zeitgeist/commits/main/node/src/cli.rs

maltekliemann:
Hmm... While in this particular case this might fit well, I wonder if this is actually a good pattern:

```
Copyright 2017-2020 Parity Technologies (UK) Ltd.
Modifications Copyright 2021-2022 Zeitgeist PM LLC.
```

For example, if I use one function from another project and add it to a 2000 line file (most of which is my work), then it will look like most of the file is no longer ours. I would opt for this, instead (sorted by date):

```
Copyright 2021-2022 Zeitgeist PM LLC.
Copyright 2017-2020 Parity Technologies (UK) Ltd.
```

Regarding the list of changes... This seems to pan into tedious extra work (when moving or renaming the file, for example), so I'll argue against it. ;) On adding the changelog, the GPL-3.0-or-later states in Section 5a (the one cited in the stackexchange post that you linked):

> The work must carry prominent notices stating that you modified it, and giving a relevant date.

IANAL, but doesn't adding our copyright notice satisfy that condition? There's no mention of a list of changes, as far as I can tell.

Furthermore, I'd say that the answer you're quoting errs so far on the side of caution that it actually creates new problems. For example, it states (regarding missing copyright notices, as in the case of the code from Balancer Labs)

> If program(s) is missing a copyright of the original author, I should add a copyright statement. If program(s) is missing a GPL notice, I should add a GPL notice: for this I can use the provided [examples](https://www.gnu.org/licenses/gpl-3.0.html#howto).

This is the exact thing that I was _warned against_  in a more recent post (https://opensource.stackexchange.com/questions/12941/citing-project-with-license-but-without-copyright-notice?noredirect=1#comment33516_12941):

> You probably should not modify or add their copyright notice. If you make a mistake, there could be legal implications.

I'm not viscerally opposed to having this one line in the file, I just wonder if we're really doing ourselves a favor.

sea212:
In regards to the example we discuss about, do you think adding
> Copyright 2021-2022 Zeitgeist PM LLC.
> Copyright 2017-2020 Parity Technologies (UK) Ltd.
is sufficient, i.e. can we remove their license header since it reflects the same license?

sea212:
@maltekliemann Can you finalize this PR?

maltekliemann:
Sorry, I thought I had answered your question already. You correct, we can remove the duplicate of the GPL-3.0-or-later license.

node/src/cli.rs
```diff
@@ -14,6 +14,25 @@
...
+//     the Free Software Foundation, either version 3 of the License, or
+//     (at your option) any later version.
+//
+//     Polkadot is distributed in the hope that it will be useful,
+//     but WITHOUT ANY WARRANTY; without even the implied warranty of
+//     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//     GNU General Public License for more details.
+//
+//     You should have received a copy of the GNU General Public License
+//     along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
```
sea212, line 35: It's the same license as we use, we should just put their Copyright statement "Copyright 2017-2020 Parity Technologies (UK) Ltd." next to ours.

maltekliemann: Sorry, I was 100% I had removed that. Is fixed now.

sea212 (approved):

PR #767, opened 2022-08-23
Toml auto format script and github workflow update.

vivekvpandya:


scripts/tests/format.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
maltekliemann, line 1: I'm wondering if this script is overkill. Here's what I mean:

- When I'm working on a PR, I can just run `cargo fmt` and then `taplo fmt` from the project directory. I don't care what the check says, I'm just happy when cargo and taplo do their job. Maybe I would like `format.sh` to run both of these commands.
- And when running a workflow, I just want to run `cargo fmt --check` and `taplo fmt --check`. Installing the necessary tools is already handled in previous steps. If they return a non-zero code, the workflow will fail (no need for `check_return_code`, I think). I wouldn't put this in a script, but rather directly in the workflow.

So in both cases, it seems to me like this script is not required. Did you have a particular purpose in mind for this?

maltekliemann (commented):
Again, great to see that `.toml` formatting is finally getting automated. :thumbsup:

This looks good, but I'd like to mention two things:

- It seems like most files are getting changed from 4 spaces per indentation to 2 spaces per indentation. If 4 spaces are OK for you, I think it would be better since it means fewer changes to the git history.
- (See comment below) Feels like the `format.sh` script is overkill (or maybe I just don't get it)

scripts/tests/format.sh
vivekvpandya: format script can have other formatting related checks for example, checking for spelling mistakes, formatting .TS files etc.
Also personally I am not very much inclined towards scripts and to me only fuzz.sh seems value able all other scripts are overkill, clippy.sh has a single line :sweat_smile:

maltekliemann: Yeah, the `clippy.sh` is my fault. 😅

I guess what I'm asking is: Why isn't `format.sh` the following right now:

```shell
cargo fmt --check
taplo fmt --check
```

vivekvpandya: I think it has just simple functions which lets specify --check when in-place changes are not expected (like CI) and sames script execution with out --check runs commands such that required changes are made.

vivekvpandya: but I am fine with any way.

maltekliemann: Oh, I see. I think we need check only in CI and in-place (without check) on local copies. So I don't see much use for the separation. But I'll let the second review decide. @sea212 @Chralt98

scripts/tests/format.sh
```diff
@@ -0,0 +1,92 @@
...
+    fi
+
+    check_return_code
+}
+
+# install taplo if it isn't already
+has_taplo=$(whereis taplo)
+if [[ ${has_taplo} = "taplo: " ]]; then
+    cargo install taplo-cli 2>/dev/null
+fi
```
sea212, line 69: We should do the same for cargo fmt, i.e. `whereis cargo fmt` and maybe `cargo install rustfmt`

sea212 (approved):
Nice Vivek, now the style in also congruent in the toml files.
We could also invoke `cargo fmt` and `taplo` in an embedded script in the job, but I personally prefer the solution you provided. It also gives developers easy access to verifying the code style locally in one command, even to such an extend that it ensures all the tools are available.

scripts/tests/format.sh
vivekvpandya: Added in https://github.com/zeitgeistpm/zeitgeist/pull/767/commits/39bb03030fb8d85702a1f4d8733a1c6914ce93f4

PR #771, opened 2022-08-25
Add new runtime

sea212:
Adds a new runtime and integrates it into the client. Also fixes initial `parachainStaking.inflationConfig` at genesis.


sea212:
We do not need a third runtime as the Kusama Zeitgeist network was migrated into the Polkadot network.

Issue #785, opened 2022-09-07
[swaps] Fix `admin_clean_up_pool` for the Rikiddo case


maltekliemann:
Calling `admin_clean_up_pool` calls `Swaps::clean_up_pool`, which calls `distribute_share_rewards` if the market is a categorical Rikiddo market. This function is not properly benchmarked and most likely subject to change when we rework Rikiddo.

Issue #728, opened 2022-08-01
Verify benchmark logic


sea212:
Verify the benchmark logic for every Zeitgeist pallet currently in use:

- [x] authorized
- [x] prediction-markets
- [x] swaps
- [x] styx

In addition to that, resolve [weight related audit issues](https://github.com/zeitgeistpm/runtime-audit-1/issues/49) (depends on https://github.com/zeitgeistpm/zeitgeist/pull/755).

sea212:
@vivekvpandya Will check the `on_resolution` and admin resolution function once https://github.com/zeitgeistpm/zeitgeist/pull/755 is merged.

Chralt98:
**prediction-markets**
- [ ] on_initialize
- [ ] admin_destroy_market
- [ ] admin_move_market_to_closed
- [ ] admin_move_market_to_resolved
- [ ] approve_market
- [ ] buy_complete_set
- [ ] dispute
- [ ] create_cpmm_market_and_deploy_assets
- [ ] create_market
- [ ] deploy_swap_pool_and_additional_liquidity
- [ ] deploy_swap_pool_for_market
- [ ] redeem_shares
- [ ] reject_market
- [ ] report
- [ ] sell_complete_set

**swaps**
- [ ]  admin_clean_up_pool
- [ ] pool_exit
- [ ] pool_exit_subsidy
- [ ] pool_exit_with_exact_asset_amount
- [ ] pool_exit_with_exact_pool_amount
- [ ] pool_join
- [ ] pool_join_subsidy
- [ ] pool_join_with_exact_asset_amount
- [ ] pool_join_with_exact_pool_amount
- [ ] swap_exact_amount_in
- [ ] swap_exact_amount_out

PR #786, opened 2022-09-07
Update `zrml_swaps` benchmarks

maltekliemann:
Partially #728 fix.

- Capped parameter `a` of the `pool_join`/`pool_exit` weight infos to `MaxAssets` for obvious reasons to prevent unnecessary spending of funds in case of excessively long values for `min_assets_out`. For the reviewers: This is not a perfectly defensive choice, since removing (through human error) the `ensure!` that... ensures... that `min_assets_out` has the same length of `pool.assets` and is therefore bounded by `MaxAssets` and/or changing the way we iterate might cause problems in the future. But I think worrying about that would be nonsensical.
- Added a `Weight` section to each documentation, giving a short summary in Landau notation of what the current _expected_ weight is
- Won't fix the benchmarks of `admin_clean_up_pool` for the Rikiddo case, which does not take the distribution of pool share rewards into account. See #785 for details.
- @Chralt98 pointed out that it might be a good idea to take the length of the vectors in the `Pool` struct into account when it comes to serialization. I'm surprised to find that even with short vectors like `assets` and `weights`, this makes a difference. We're currently benchmarking the worst-case scenario. Since the length is bounded by quite a low number, I'd like to keep the benchmarks and not make this a benchmark parameter.
- Added worst-case scenarios for `swap_exact_amount_*` functions. This means maximizing the number of iterations that `bpow_approx` requires (we're ignoring `bpowi`, since that seems benign in terms of computation time). Haven't done that for `pool_*_with_exact_*_amount`, because the worst-case scenarios with many iterations in `bpow_approx` require us to use fewer assets, which seems to reduce the weight of the benchmarked call (probably due to the faster serialization). It's not possible to somehow map this behavior using a benchmark parameter since the number of iterations is not linearly related to any extrinsic parameter.
- Some minor clean up


maltekliemann:
@Chralt98 Hmm, not sure if I understand. What problem would that solve?

Chralt98:
If the case happens, that some contributor replaces things or loops over the vector before the size check, then we would run into a problem.

maltekliemann:
Hmm, okay. I guess we could do that. Two arguments against that:

- To stay consistent, we'd have to implement this defensive coding practice for all of our functions.
- I don't think we can use pallet constants as length for the bounded vector, so we're probably double-booking the maximum length.

Let's discuss this in the next meeting!

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,57 +475,133 @@ benchmarks! {
...
+            ScoringRule::CPMM,
+            false,
+            Some(weights),
+        );
+        let asset_in = assets[0];
+        T::AssetManager::deposit(asset_in, &caller, u64::MAX.saturated_into()).unwrap();
+        let asset_out = assets[asset_count as usize - 1];
+        let min_asset_amount_out: Option<BalanceOf<T>> = Some(0u128.saturated_into());
+        let max_price = Some(u128::MAX.saturated_into());
+    }: swap_exact_amount_in(
```
Chralt98, line 525: That is a little bit misleading, because there exists an extrinsic called `swap_exact_amount_in` and a same called function. Here you are calling the extrinsic, because you use the origin instead of the account id as the first parameter. In the [function `swap_exact_amount_in` you use this benchmark](https://github.com/zeitgeistpm/zeitgeist/blob/bb3e6185bc8d2cf9e849c1336ea7e52e817c71ed/zrml/swaps/src/lib.rs#L2153) result as return value. This might grow into a problem, when we add overhead in the extrinsic, that weight will automatically get added to the sub-function additionally.

Mitigation Approach: Maybe benchmark the sub-function and not the extrinsic. But because there is no overhead at the moment in the extrinsic, this is a nitpick.

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,57 +475,133 @@ benchmarks! {
...
+        RawOrigin::Signed(caller),
```
Chralt98, line 526:
```suggestion
        caller,
```

zrml/swaps/src/lib.rs
```diff
@@ -636,6 +706,11 @@ mod pallet {
         /// * `asset_out`: Asset leaving the pool.
         /// * `asset_amount_out`: Amount that will be transferred from the pool to the provider.
         /// * `max_price`: Market price must be equal or less than the provided value.
+        ///
+        /// # Weight
+        ///
+        /// Complexity: `O(1)` if the scoring rule is CPMM, `O(n)` where `n` is the amount of
+        /// assets if the scoring rule is Rikiddo.
         #[pallet::weight(T::WeightInfo::swap_exact_amount_out_rikiddo(T::MaxAssets::get().into()))]
```
Chralt98, line 714: I would probably use the path of CPMM here and add a TODO for rikiddo benchmark check.

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,57 +475,133 @@ benchmarks! {
...
+            ScoringRule::CPMM,
+            false,
+            Some(weights),
+        );
+        let asset_in = assets[0];
+        T::AssetManager::deposit(asset_in, &caller, u64::MAX.saturated_into()).unwrap();
+        let asset_out = assets[asset_count as usize - 1];
+        let max_asset_amount_in: Option<BalanceOf<T>> = Some(u128::MAX.saturated_into());
+        let max_price = Some(u128::MAX.saturated_into());
+    }: swap_exact_amount_out(
```
Chralt98, line 584: Vice versa: 

That is a little bit misleading, because there exists an extrinsic called `swap_exact_amount_in` and a same called function. Here you are calling the extrinsic, because you use the origin instead of the account id as the first parameter. In the [function `swap_exact_amount_in` you use this benchmark](https://github.com/zeitgeistpm/zeitgeist/blob/bb3e6185bc8d2cf9e849c1336ea7e52e817c71ed/zrml/swaps/src/lib.rs#L2153) result as return value. This might grow into a problem, when we add overhead in the extrinsic, that weight will automatically get added to the sub-function additionally.

Mitigation Approach: Maybe benchmark the sub-function and not the extrinsic. But because there is no overhead at the moment in the extrinsic, this is a nitpick.

zrml/swaps/src/weights.rs
```diff
@@ -71,7 +72,16 @@ impl<T: frame_system::Config> WeightInfoZeitgeist for WeightInfo<T> {
     // Storage: MarketCommons Markets (r:1 w:0)
     // Storage: MarketCommons MarketPool (r:1 w:0)
     // Storage: Swaps Pools (r:1 w:1)
-    fn admin_clean_up_pool() -> Weight {
+    fn admin_clean_up_pool_cpmm_categorical(_: u32) -> Weight {
+        (87_620_000 as Weight)
+            .saturating_add(T::DbWeight::get().reads(4 as Weight))
+            .saturating_add(T::DbWeight::get().writes(2 as Weight))
+    }
```
Chralt98, line 79: I wonder why the weight doesn't depend on the component for the number of assets at the moment. 🤨 It should.

Chralt98 (approved):
I verified all extrinsics for swaps again. You often put in `MaxAssets` to charge the worst case (I can live with that). The weights should be safe (it's better to charge more than too less).

I especially like your time complexity comments.

zrml/swaps/src/weights.rs
maltekliemann: Oh, I never ran the benchmarks. I just took the one that was already there and added an used parameter to keep the PR clean.

zrml/swaps/src/lib.rs
```diff
@@ -599,6 +664,11 @@ mod pallet {
         /// * `asset_out`: Asset leaving the pool.
         /// * `min_asset_amount_out`: Minimum asset amount that can leave the pool.
         /// * `max_price`: Market price must be equal or less than the provided value.
+        ///
+        /// # Weight
+        ///
+        /// Complexity: `O(1)` if the scoring rule is CPMM, `O(n)` where `n` is the amount of
+        /// assets if the scoring rule is Rikiddo.
         #[pallet::weight(T::WeightInfo::swap_exact_amount_in_rikiddo(T::MaxAssets::get().into()))]
```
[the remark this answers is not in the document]
maltekliemann: Fixed in 2b420a37a292f1eae806cb9d51cd6eb0f25364e2

zrml/swaps/src/lib.rs
maltekliemann: Fixed in 2b420a37a292f1eae806cb9d51cd6eb0f25364e2

zrml/swaps/src/weights.rs
maltekliemann: When I run the benchmark locally, I get

```
========
Storage: unknown [0x3a7472616e73616374696f6e5f6c6576656c3a] (r:1 w:1)
Storage: MarketCommons Markets (r:1 w:0)
Storage: MarketCommons MarketPool (r:1 w:0)
Storage: Swaps Pools (r:1 w:1)

Median Slopes Analysis
========
-- Extrinsic Time --

Model:
Time ~=    57.62
    + a    0.875
              µs

Reads = 4 + (0 * a)
Writes = 2 + (0 * a)
```

Chralt98: Ok, then it looks like expected 👍

Chralt98 (approved):

Chralt98:
I wanted to comment again about [`bpow_approx`](https://github.com/zeitgeistpm/zeitgeist/blob/294c66c4d070171fbe4389fbb63b1608e7df3ba8/zrml/swaps/src/fixed.rs#L113-L164). Wasn't an action required there? Did you benchmark the worst case separately? What was the weight for the worst case. I just ask, because I am interested in the comparison with the other weights.

zrml/swaps/src/benchmarks.rs
```diff
@@ -163,12 +172,72 @@ benchmarks! {
                 status: MarketStatus::Active,
             }
         )?;
-        let _ = T::MarketCommons::insert_market_pool(0u32.saturated_into(), 0u128);
-        let _ = bench_create_pool::<T>(caller, Some(T::MaxAssets::get().into()), None, ScoringRule::CPMM, false);
-        let _ = Pallet::<T>::mutate_pool(0, |pool| {
-            pool.pool_status = PoolStatus::Closed; Ok(())
+        let market_id = T::MarketCommons::latest_market_id()?;
```
sea212, line 175: The `market_id` is returned by `push_market` above:
```rust
    /// Pushes a new market into the storage, returning its related auto-incremented ID.
    fn push_market(
        market: Market<Self::AccountId, Self::BlockNumber, Self::Moment>,
    ) -> Result<Self::MarketId, DispatchError>;
```

We should use the market id retrieved by that instead of doing an additional storage read. With the current default benchmark configuration each benchmark is executed up to 10,000 times, so a storage operation can become quite heavy in total.

zrml/swaps/src/benchmarks.rs
```diff
@@ -123,7 +128,7 @@ fn bench_create_pool<T: Config>(
         scoring_rule,
         if scoring_rule == ScoringRule::CPMM { Some(Zero::zero()) } else { None },
         if scoring_rule == ScoringRule::CPMM { Some(T::MinLiquidity::get()) } else { None },
-        if scoring_rule == ScoringRule::CPMM { Some(weights) } else { None },
+        if scoring_rule == ScoringRule::CPMM { some_weights } else { None },
     )
     .unwrap();
     let pool_id = <NextPoolId<T>>::get() - 1;
```
sea212, line 134: `create_pool(...)` above returns the pool id, we should use that instead (to avoid storage operation).

zrml/swaps/src/benchmarks.rs
```diff
@@ -163,12 +172,72 @@ benchmarks! {
...
+            Ok(())
+        });
+
+        let call = Call::<T>::admin_clean_up_pool {
+            market_id,
+            outcome_report: OutcomeReport::Scalar(33),
+        };
+    }: {
+        call.dispatch_bypass_filter(RawOrigin::Root.into())?;
+    }
```
sea212, line 240:
```suggestion
    }: admin_clean_up_pool(RawOrigin::Root, market_id, OutcomeReport::Scalar(33))
```

zrml/swaps/src/benchmarks.rs
```diff
@@ -163,12 +172,72 @@ benchmarks! {
...
         });
-    }: _(RawOrigin::Root, 0u32.into(), OutcomeReport::Categorical(0))
+
+        let call = Call::<T>::admin_clean_up_pool {
+            market_id,
+            outcome_report: OutcomeReport::Categorical(0),
+        };
+    }: {
+        call.dispatch_bypass_filter(RawOrigin::Root.into())?;
+    }
```
sea212, line 197:
```suggestion
}: admin_clean_up_pool(RawOrigin::Root, market_id, OutcomeReport::Categorical(0))
```

zrml/swaps/src/benchmarks.rs
```diff
@@ -163,12 +172,72 @@ benchmarks! {
...
+                metadata: vec![0; 50],
+                oracle: caller.clone(),
+                period: MarketPeriod::Block(0u32.into()..1u32.into()),
+                report: None,
+                resolved_outcome: None,
+                scoring_rule: ScoringRule::CPMM,
+                status: MarketStatus::Active,
+            }
+        )?;
+        let market_id = T::MarketCommons::latest_market_id()?;
```
sea212, line 217:
```suggestion
        let market_id = T::MarketCommons::push_market(
            Market {
                creation: MarketCreation::Permissionless,
                creator_fee: 0,
                creator: caller.clone(),
                market_type: MarketType::Scalar(0..=99),
                dispute_mechanism: MarketDisputeMechanism::Authorized(caller.clone()),
                metadata: vec![0; 50],
                oracle: caller.clone(),
                period: MarketPeriod::Block(0u32.into()..1u32.into()),
                report: None,
                resolved_outcome: None,
                scoring_rule: ScoringRule::CPMM,
                status: MarketStatus::Active,
            }
        )?;
```

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,57 +475,133 @@ benchmarks! {
...
     }
 
     swap_exact_amount_in_cpmm {
-        let a = T::MaxAssets::get();
+        // We're trying to get as many iterations in `bpow_approx` as possible. Experiments have
+        // shown that y = 3/4, weight_ratio=1/2 (almost) maximizes the number of iterations for
+        // calculating y^r within the set of values allowed in `swap_exact_amount_in` (see
+        // `math::calc_out_given_in`). To get these values, we use the following parameters:
+        // amount_in = 1/3 * balance_in, weight_in = 1, weight_out = 2.
+        let asset_count = T::MaxAssets::get();
```
sea212, line 499: Using this value probably overcharges the users a bit in most cases, because it assumes during pool encoding that we'll have the maximum amount of `assets` and `weights`. I think though, that the value we overcharge in this case is so low, that we can neglect it here and save the extra time required for the benchmark to determine it.

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,57 +475,133 @@ benchmarks! {
...
         )?;
     }
 
     swap_exact_amount_in_cpmm {
-        let a = T::MaxAssets::get();
+        // We're trying to get as many iterations in `bpow_approx` as possible. Experiments have
+        // shown that y = 3/4, weight_ratio=1/2 (almost) maximizes the number of iterations for
+        // calculating y^r within the set of values allowed in `swap_exact_amount_in` (see
+        // `math::calc_out_given_in`). To get these values, we use the following parameters:
+        // amount_in = 1/3 * balance_in, weight_in = 1, weight_out = 2.
```
sea212, line 498: Nice :+1:

sea212 (commented):
It looks solid overall, just have some nitpicks.

zrml/swaps/src/lib.rs
```diff
@@ -103,19 +103,28 @@ mod pallet {
...
     impl<T: Config> Pallet<T> {
         /// Clean up the pool of a resolved market.
-        #[pallet::weight(T::WeightInfo::admin_clean_up_pool())]
+        ///
+        /// # Weight
+        ///
+        /// Complexity: `O(1)` if the market is scalar, `O(n)` where `n` is the number of
+        /// assets in the pool if the market is categorical.
+        #[pallet::weight(
+            T::WeightInfo::admin_clean_up_pool_cpmm_categorical(T::MaxAssets::get() as u32)
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in c97b636d898536640d272dc48614d6c826d52779

zrml/swaps/src/lib.rs
```diff
@@ -103,19 +103,28 @@ mod pallet {
...
     >>::Balance;
+    type MarketIdOf<T> = <<T as Config>::MarketCommons as MarketCommonsPalletApi>::MarketId;
 
     const MIN_BALANCE: u128 = CENT;
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
         /// Clean up the pool of a resolved market.
-        #[pallet::weight(T::WeightInfo::admin_clean_up_pool())]
+        ///
```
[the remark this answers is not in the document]
maltekliemann: Mitigated in c97b636d898536640d272dc48614d6c826d52779

zrml/swaps/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: Fixed in bb645b37c5781a8569d6687428d72b11791c6adf

zrml/swaps/src/benchmarks.rs
```diff
@@ -145,14 +150,18 @@ fn bench_create_pool<T: Config>(
 }
 
 benchmarks! {
-    admin_clean_up_pool {
+    admin_clean_up_pool_cpmm_categorical {
+        // We're excluding the case of two assets, which would leave us with only one outcome
+        // token, which makes no sense in the context of prediction markets.
+        let a in 3..T::MaxAssets::get().into();
```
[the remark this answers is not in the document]
maltekliemann: Unfortunately, we have to call `create_market`, which will error if we only provide one outcome token. I've updated the comment to make the problem a bit clearer. Assuming that `market-commons` does its job correctly, the case `a = 2` cannot occur in nature, even when you're using the swaps pallet "solo".

Of course, we could go full hog and start mutating the market after creation to only have one asset, then deploy a pool, and go from there, but I would suggest we don't do that.

The fact that `predicion-markets` and `swaps` are coupled in this way though this function is a code smell that we may want to eliminate someday by only passing the `primitives` of the market that are actually needed for cleaning up the pool. The fact that we can't benchmark this function for `a = 2` is another manifestation of that smell.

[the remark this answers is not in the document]
maltekliemann: Mitigated in 0ed5f91cca048512e19671465cfc49d5c0f7e6d9

zrml/swaps/src/benchmarks.rs
maltekliemann: Mitigated in 643812487ea202067721351d6381c38c699dd910

zrml/swaps/src/benchmarks.rs
maltekliemann: Mitigated in 1b280a1d152f2d2cc48622826130a8d9da1e846b

zrml/swaps/src/benchmarks.rs
maltekliemann: I think the difference was surprisingly large, like 10%. I believe Chris and I had a discussion about this. The reason seems to be that the vector needs to be deserialized, so it's `O(len)`. Nevertheless, I argued against parametrizing the benchmark by the asset count, since that would mean basically parametrizing every extrinsic which reads a pool.

Definitely something to keep in mind, though.

zrml/swaps/src/benchmarks.rs
maltekliemann: Mitigated in 5ab83c028b221e8df1eaefee3159e17729602796

zrml/swaps/src/benchmarks.rs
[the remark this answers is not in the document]
sea212: In that case I am fine with leaving the benchmark as it is.

zrml/swaps/src/benchmarks.rs
sea212: I think it makes sense to do that for structures that can grow quite big. I am fine with leaving it out now, we over-estimate which is fine and we can respect decoding in a future optimization run.

sea212 (approved):
LGTM

Chralt98 (approved):
Yeah, looks pretty solid now.

PR #789, opened 2022-09-09
Add and use production profile

### LINKED ISSUE #788: Cargo rustc production profile
I noticed while benchmarking, that [Moonbeam uses a production profile](https://github.com/PureStake/moonbeam/blob/eef06f9aba68412370130f2170bafdb45f4bd526/Cargo.toml#L44-L53) for optimise further than the default `--release` flag.

https://doc.rust-lang.org/cargo/reference/profiles.html

[Polkadot has the same approach](https://github.com/paritytech/polkadot/blob/87d6cfc7c0531d888898c9bbd1a1e299538a546b/Cargo.toml#L185). 

**Proposal profile production**: 
- [x] lto = true
  - > LTO can produce better optimized code, using whole-program analysis, at the cost of longer linking time.
- [x] codegen-units = 1
  - > When a crate is split into multiple codegen units, LLVM is able to process them in parallel. Increasing parallelism may speed up compile times, but may also produce slower code. Setting this to 1 may improve the performance of generated code, but may be slower to compile.
- [x] incremental = false
  -  > The incremental setting controls the [-C incremental flag](https://doc.rust-lang.org/rustc/codegen-options/index.html#incremental) which controls whether or not incremental compilation is enabled. Incremental compilation causes rustc to save additional information to disk which will be reused when recompiling the crate, improving re-compile times. The additional information is stored in the target directory.
  
This slows compile time down a significant amount, but it's even more optimised for production and we should use the `dev` or `test` profile for development purpose.

It's interesting to use the production profile later for the benchmarks.

**Solution:**

In the root `Cargo.toml` file we should add the following.

```toml
[profile.production]
inherits = "release"
lto = true
codegen-units = 1
incremental = false

[profile.release]
opt-level = 3
# Zeitgeist runtime requires unwinding.
panic = "unwind"
```

For the default `--profile=release` or `--release` the `opt-level` and `panic`[ is the same like here specified](https://doc.rust-lang.org/cargo/reference/profiles.html#release). But just in case, we absolutely would like to make sure, that the production profile inherits the full optimisation, because we don't know what is changed as default in the future.

Chralt98:
Fixes #788

https://nnethercote.github.io/perf-book/build-configuration.html

I look forward to the benchmark results on our production machine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,10 @@ pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branc
 parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
 
 [profile.production]
-inherits = "release"
-lto = true
 codegen-units = 1
 incremental = false
+inherits = "release"
+lto = true
 
 [profile.release]
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ sha2 = { opt-level = 3 }
 sha3 = { opt-level = 3 }
 smallvec = { opt-level = 3 }
 snow = { opt-level = 3 }
-substrate-bip39 = {opt-level = 3}
+substrate-bip39 = { opt-level = 3 }
 twox-hash = { opt-level = 3 }
 uint = { opt-level = 3 }
 wasmi = { opt-level = 3 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,14 @@ resolver = "2"
 [patch."https://github.com/purestake/moonbeam"]
 pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
 parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
+
+[profile.production]
+inherits = "release"
+lto = true
+codegen-units = 1
+incremental = false
+
+[profile.release]
+opt-level = 3
+# Zeitgeist runtime requires unwinding.
+panic = "unwind"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,62 @@ resolver = "2"
 pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
 parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
 
+# The list of dependencies below (which can be both direct and indirect dependencies) are crates
+# that are suspected to be CPU-intensive, and that are unlikely to require debugging (as some of
+# their debug info might be missing) or to require to be frequently recompiled. We compile these
+# dependencies with `opt-level=3` even in "dev" mode in order to make "dev" mode more usable.
+# The majority of these crates are cryptographic libraries.
+#
+# If you see an error mentioning "profile package spec ... did not match any packages", it
+# probably concerns this list.
+#
+# This list is ordered alphabetically.
+[profile.dev.package]
+blake2 = { opt-level = 3 }
+blake2-rfc = { opt-level = 3 }
+blake2b_simd = { opt-level = 3 }
+chacha20poly1305 = { opt-level = 3 }
+cranelift-codegen = { opt-level = 3 }
+cranelift-wasm = { opt-level = 3 }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -48,3 +48,14 @@ resolver = "2"
 [patch."https://github.com/purestake/moonbeam"]
 pallet-author-mapping = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
 parachain-staking = { git = "https://github.com/zeitgeistpm/moonbeam", branch = "downgrade-staking-and-mapping" }
+
```
sea212, line 51: Can we suppress compiling debug symbols for some of the heavy crates that are very unlikely to be the cause of an error, like [Paritytech](https://github.com/paritytech/polkadot/blob/09f340c928ea324105b39b957dc155cbd29b173d/Cargo.toml#L117-L171) did?

Cargo.toml
```diff
@@ -48,3 +48,14 @@ resolver = "2"
...
+[profile.production]
+codegen-units = 1
+incremental = false
```
sea212, line 54: Why is `incremental` set to false?

sea212 (changes requested):
Good changes that improve the production build :+1: . 
The following files should also use production builds:
- *scripts/build-parachain-spec.sh*
- *.github/workflows/docker-hub.yml* (set `PROFILE`)
- *.github/workflows/docker-hub-parachain.yml*) (set `PROFILE`)

Cargo.toml
Chralt98: > The incremental setting controls the [-C incremental flag](https://doc.rust-lang.org/rustc/codegen-options/index.html#incremental) which controls whether or not incremental compilation is enabled. Incremental compilation causes rustc to save additional information to disk which will be reused when recompiling the crate, improving re-compile times. The additional information is stored in the target directory.

It's default by the release profile [here](https://doc.rust-lang.org/cargo/reference/profiles.html#release), but Moonbeam wrote it still inside the configuration [here](https://github.com/PureStake/moonbeam/blob/eef06f9aba68412370130f2170bafdb45f4bd526/Cargo.toml#L44-L53)

Cargo.toml
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/789/commits/9907667a5bf0e5ddadb1af4e91cb2bd37db853e8

sea212:
@Chralt98 Waiting with the review until the formatting errors are resolved.

scripts/benchmarks/run_benchmarks.sh
```diff
@@ -16,28 +16,28 @@ source ./scripts/benchmarks/configuration.sh
 
 # Standalone benchmarks
 
-cargo build --release --features=runtime-benchmarks --bin=zeitgeist
+cargo build --profile=production --features=runtime-benchmarks --bin=zeitgeist
```
vivekvpandya, line 19: would not this increase compile time, if I just want to run benchmarks on local machine? 
Is it possible to make this override with an simple environment variable ?

Chralt98: When you use `run_benchmarks.sh` it has a longer compilation time now, but with more optimizations for the benchmarks. For what purpose do you want to run the benchmarks on your local machine? Just to check if they are running? I use this script for it: 
```bash
./target/release/zeitgeist benchmark pallet --chain dev --execution=native --wasm-execution=compiled --heap-pages=4096 --pallet "*" --extrinsic "*" --steps 1 --repeat 0 --detailed-log-output
```

vivekvpandya: sure I can use these commands directly.

Chralt98:
> cargo build --production should not work, it should use cargo build --profile production

Good catch. 4775a818c96ada49062fdf3389350749ae1093ca

scripts/benchmarks/run_benchmarks.sh
sea212: Actually I think Viveks suggestion is good. Having an ENV variable that is set to `production` by default is useful.

sea212: AFAIK this approach will always override the variable. The correct approach in bash to use an ENV variable or a default value in case it is not set is as follows: `PROFILE="${PROFILE-production}"`

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/789/commits/84f78ae280125da3cfc97bfff149e0ceacceb047

sea212 (approved):
Nice, looking forward to see the next benchmark results in comparison!

vivekvpandya (approved):

PR #796, opened 2022-09-14
Increase MaxCategories to 64

### LINKED ISSUE #791: [Runtime] Increase prediction market max categories
Set `MAX_CATEGORIES` to `64`. This change was planned in #595, but went under in the corresponding PR.

sea212:
Closes #791


maltekliemann (approved):
Yeah, that checks out! :thumbsup:

PR #802, opened 2022-09-22
Change `AdvisoryBond` slash percentage to 10%

maltekliemann:


zrml/prediction-markets/src/lib.rs
```diff
@@ -947,6 +947,10 @@ mod pallet {
         #[pallet::constant]
         type AdvisoryBond: Get<BalanceOf<Self>>;
 
+        /// The percentage of the advisory bond that gets slashed when a market is rejected.
+        #[pallet::constant]
+        type AdvisoryBondSlashPercentage: Get<Percent>;
```
Chralt98, line 952: In `admin_destroy_market` it's okay to slash `100%`? I am fine with this, because sudo action should be heavy.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1488,17 +1492,21 @@ mod pallet {
...
                 creator,
-                T::AdvisoryBond::get(),
+                advisory_bond_slash_amount,
             );
             T::AssetManager::unreserve_named(
                 &Self::reserve_id(),
                 Asset::Ztg,
                 creator,
-                T::OracleBond::get(),
+                T::OracleBond::get() + advisory_bond_unreserve_amount,
```
Chralt98, line 1509:
```suggestion
                T::OracleBond::get().saturating_add(advisory_bond_unreserve_amount),
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1488,17 +1492,21 @@ mod pallet {
         ) -> Result<Weight, DispatchError> {
             ensure!(market.status == MarketStatus::Proposed, Error::<T>::InvalidMarketStatus);
             let creator = &market.creator;
+            let advisory_bond_slash_amount =
+                T::AdvisoryBondSlashPercentage::get().mul_floor(T::AdvisoryBond::get());
+            let advisory_bond_unreserve_amount =
+                T::AdvisoryBond::get() - advisory_bond_slash_amount;
```
Chralt98, line 1498:
```suggestion
                T::AdvisoryBond::get().saturating_sub(advisory_bond_slash_amount);
```

I know it can't produce an underflow, but just to be safe for future devs, who will modify the code.

docs/changelog_for_devs.md
```diff
@@ -1,3 +1,8 @@
+# v0.3.6
+
+- Changed the slash percentage of the `AdvisoryBond` from 100% to 10% by
```
Chralt98, line 3:
```suggestion
- Changed the `reject_market` slash percentage of the `AdvisoryBond` from 100% to 10% by
```

Chralt98 (dismissed):
LGTM

zrml/prediction-markets/src/lib.rs
maltekliemann: Yes, I think so. There's no point in calling `admin_destroy_market` on a market that's proposed because you can achieve the same result by rejecting the market. And when the market is no longer proposed, then the `AdvisoryBond` has already been released.

Chralt98 (approved):

vivekvpandya (approved):
Change looks great! However just missing context, why this was decided to change?

maltekliemann:
> Change looks great! However just missing context, why this was decided to change?

Oh, sorry, didn't catch this question! The Advisory Committee noticed that in most cases, slashing the entire bond seemed to harsh. Most markets are 90% good, but have small flaws. Slashing the equivalent of 40$ for a small error seems unreasonable.

Issue #792, opened 2022-09-14
Investigation: How to safely remove outcome assets ("garbage collect") over multiple blocks?


Chralt98:
**Problem**

Zeitgeist should provide a fast environment, in which outcome assets get created and unused ones get efficiently destroyed.

Currently Substrate uses a [Base-16 Modified Merkle Tree ("Trie") data structure](https://github.com/paritytech/trie) above a key-value database. That's why each storage read/write, in this case the deletion of an outcome asset, takes O(log(n)), where n is the number of nodes in the merkle tree. Therefore storage accesses are more expensive than a non-tree key-value database.
However, to save balances for traders in a prediction market, we use [a StorageDoubleMap from `orml_tokens`](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/src/lib.rs#L369-L385). To properly delete all balances for a specific currency/asset, we are left with the usage of the storage API.

The current challenge is, that the `orml_tokens` `StorageDoubleMap` maps from `AccountId` to `CurrencyId`, which requires us to iterate over all accounts, which is unsustainable. That's why [we tried to cache all accounts, which have a specific currency id](https://github.com/zeitgeistpm/zeitgeist/pull/755). This adds extra storage and expensive reads and writes and will not scale well in the long run, when the storage layers are not switched. There is no change in sight, which will allow us to have a `StorageDoubleMap` from `CurrencyId` to `AccountId`, [because `orml_tokens` is optimised for the query of one specific `AccountId` rather than to get one specific `CurrencyId`](https://github.com/open-web3-stack/open-runtime-module-library/issues/802).

[...]

PR #806, opened 2022-09-28
Remove destroying outcome assets when market is destroyed or resolved

vivekvpandya:


Chralt98:
Starting review now..

zrml/prediction-markets/src/lib.rs
```diff
@@ -221,11 +179,8 @@ mod pallet {
         /// market to resolved.
         ////
         #[pallet::weight((T::WeightInfo::admin_move_market_to_resolved_overhead()
-            .saturating_add(T::WeightInfo::internal_resolve_categorical_reported(
-                4_200,
-                4_200,
-                T::MaxCategories::get().into()
-            ).saturating_sub(T::WeightInfo::internal_resolve_scalar_reported())
+            .saturating_add(T::WeightInfo::internal_resolve_categorical_reported()
+            .saturating_sub(T::WeightInfo::internal_resolve_scalar_reported())
```
Chralt98, line 183: Do you know why this is `saturating_sub` here?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -225,26 +225,13 @@ fn setup_resolve_common_scalar_after_dispute<T: Config>(
...
-        let b in 0..10;
-        // c = num. asset types
-        let c in (T::MinCategories::get().into())..T::MaxCategories::get().into();
-        // Complexity: c*a + c*b ∈ O(a)
-        let c_u16 = c.saturated_into();
-        let (caller, market_id) = setup_resolve_common_categorical::<T>(a, b, c_u16)?;
-
-        for i in 0..c.min(T::MaxDisputes::get()) {
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
```
Chralt98, line 229: Just sample values, that don't influence the benchmark weights?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -225,26 +225,13 @@ fn setup_resolve_common_scalar_after_dispute<T: Config>(
 
 benchmarks! {
     admin_destroy_disputed_market{
```
Chralt98, line 227: You benchmark it in another PR? (after you corrected the benchmark for `on_resolution`), right?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -371,35 +352,37 @@ benchmarks! {
     }: { Pallet::<T>::handle_expired_advised_market(&market_id, market)? }
 
     internal_resolve_categorical_reported {
```
Chralt98, line 354: You benchmark it (verify the benchmark) in another PR?

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: yes.

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: yes.

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: yes.

zrml/prediction-markets/src/lib.rs
vivekvpandya: As per comment on admin_move_market_to_resolved_overhead() benchmark it uses scalar market hence we does not substract internal_resolve_scalar_reported() then weight of on_resolution() for scalar market will be counted twice, first in admin_move_market_to_resolved_overhead() and then again in internal_resolve_categorical_reported() :thinking:

zrml/prediction-markets/src/lib.rs
```diff
@@ -131,24 +121,9 @@ mod pallet {
...
-                    None
-                }
-            };
-            for asset in outcome_assets.into_iter() {
-                if let Some(total) = destroy_asset(asset) {
-                    total_accounts = total;
-                }
-            }
+            //NOTE: Currently we don't clean up outcome assets.
+            //TODO: Above needs more work on https://github.com/zeitgeistpm/zeitgeist/issues/101
```
Chralt98, line 125: I don't know, why the CI doesn't complain with this TODO comment, but the style is normally the following.

```suggestion
            // TODO(#792): Remove outcome assets for accounts! Delete asset of `orml_tokens` with storage migration.
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1905,18 +1802,7 @@ mod pallet {
...
-
-            if let Ok([local_total_accounts, local_total_asset_accounts, local_total_categories]) =
-                Self::manage_resolved_categorical_market(market, market_id, &resolved_outcome)
-            {
-                total_accounts = local_total_accounts.saturated_into();
-                total_asset_accounts = local_total_asset_accounts.saturated_into();
-                total_categories = local_total_categories.saturated_into();
-            }
-
+            //NOTE: Currently we don't clean up outcome assets.
```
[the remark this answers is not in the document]
Chralt98: I meant that we need to clean up the resolved outcome assets.

Chralt98:
Starting second review..

zrml/prediction-markets/src/lib.rs
Chralt98: Maybe reference this PR with the full link to it in this comment to, that we later have an idea, what code was deleted and can be probably reused https://github.com/zeitgeistpm/zeitgeist/pull/806

Maybe add  this comment also to on_idle or on_initialize.

zrml/prediction-markets/src/weights.rs
```diff
@@ -76,56 +76,34 @@ impl<T: frame_system::Config> WeightInfoZeitgeist for WeightInfo<T> {
...
-            .saturating_add((100_222_000 as Weight).saturating_mul(c as Weight))
-            .saturating_add(T::DbWeight::get().reads(7 as Weight))
-            .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
-            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(a as Weight)))
-            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(b as Weight)))
-            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(c as Weight)))
+    fn admin_destroy_disputed_market() -> Weight {
+        (161_945_000 as Weight)
+            .saturating_add(T::DbWeight::get().reads(6 as Weight))
+            .saturating_add(T::DbWeight::get().writes(5 as Weight))
```
Chralt98, line 86: Wow, what a difference to before!

zrml/prediction-markets/src/lib.rs
```diff
@@ -1905,18 +1803,8 @@ mod pallet {
...
+            //TODO: Above needs more work on https://github.com/zeitgeistpm/zeitgeist/issues/101
```
Chralt98, line 1807: Wow, what a lightweight `on_resolution` now!

zrml/prediction-markets/src/lib.rs
Chralt98: I see. Okay!

zrml/prediction-markets/src/benchmarks.rs
Chralt98: I think the reason `setup_resolve_common_categorical` has `total_accounts` etc. as parameter was, because of the dependence on the benchmark performance of for example `admin_destroy_market`. So it's basically irrelevant here what you put in. Maybe use obvious arbitrary numbers like `42` or I am thinking about removing the parameter from the function completely and reference this PR as comment in case of later usage with the garbage collection.

Chralt98 (commented):
We still have an unregarded case for `destroy_pool` in `admin_destroy_market`.

https://github.com/zeitgeistpm/zeitgeist/blob/710cb62ace72b6eb975bc538b35717c00cbd4b38/zrml/swaps/src/lib.rs#L1575-L1582

There we got a weight dependence, because we iterate again over all accounts... We should probably fix that too in this in this PR, while we are at it.

zrml/prediction-markets/src/weights.rs
vivekvpandya: Most likely this will change, as I have benchmaked them on a laptop which is not much powerful. once we are confident with reviews I can add benchmarks needed lable.

Chralt98: I mean, it's now not dependent on benchmark components. That's great!

vivekvpandya:
> We still have an unregarded case for `destroy_pool` in `admin_destroy_market`.
> 
> https://github.com/zeitgeistpm/zeitgeist/blob/710cb62ace72b6eb975bc538b35717c00cbd4b38/zrml/swaps/src/lib.rs#L1575-L1582
> 
> There we got a weight dependence, because we iterate again over all accounts... We should probably fix that too in this in this PR, while we are at it. @maltekliemann What do you think about that?

I am not much confidence about swap pallet's benchmarks so I request not to delay this PR because of it. Keep things short and we make faster progress.

zrml/prediction-markets/src/lib.rs
```diff
@@ -99,13 +93,9 @@ mod pallet {
             // TODO(#618): Not implemented for Rikiddo!
             T::DestroyOrigin::ensure_origin(origin)?;
 
-            let mut total_accounts = 0usize;
-            let mut share_accounts = 0usize;
             let market = T::MarketCommons::market(&market_id)?;
             ensure!(market.scoring_rule == ScoringRule::CPMM, Error::<T>::InvalidScoringRule);
             let market_status = market.status;
-            let outcome_assets = Self::outcome_assets(market_id, &market);
-            let outcome_assets_amount = outcome_assets.len();
```
sea212, line 108: The market will be deleted completely out of storage later and we won't know which outcome assets to delete in the future. I think this is fine though, since we can deduct which which markets got deleted (we got incremental market ids and some won't return a market -> delete markets) and then iterate through all accounts in a (potential multi-block) migration and delete outcome assets where the market id belongs to one of the deleted markets.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -225,26 +225,13 @@ fn setup_resolve_common_scalar_after_dispute<T: Config>(
...
-        let c in (T::MinCategories::get().into())..T::MaxCategories::get().into();
-        // Complexity: c*a + c*b ∈ O(a)
-        let c_u16 = c.saturated_into();
-        let (caller, market_id) = setup_resolve_common_categorical::<T>(a, b, c_u16)?;
-
-        for i in 0..c.min(T::MaxDisputes::get()) {
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
+        let categories = T::MaxCategories::get();
+        let (caller, market_id) = setup_resolve_common_categorical::<T>(total_accounts, accounts_with_assets, categories)?;
```
sea212, line 231: We don't need any outcome token holding account now, so we can remove the overhead of creating them in the benchmarks (which get executed thousands of times):
```suggestion
        let categories = T::MaxCategories::get();
        let (caller, market_id) = setup_resolve_common_categorical::<T>(0, 0, categories)?;
```

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -256,16 +243,10 @@ benchmarks! {
...
-        // c = num. asset types
-        let c in (T::MinCategories::get().into())..T::MaxCategories::get().into();
-        // Complexity: c*a + c*b ∈ O(a)
-
-        let c_u16 = c.saturated_into();
-        let (caller, market_id) = setup_resolve_common_categorical::<T>(a, b, c_u16)?;
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
+        let categories :u16 = T::MaxCategories::get().saturated_into();
+        let (caller, market_id) = setup_resolve_common_categorical::<T>(total_accounts, accounts_with_assets, categories)?;
```
sea212, line 249: We don't need any outcome token holding account now, so we can remove the overhead of creating them in the benchmarks (which get executed thousands of times):
```suggestion
        let categories = T::MaxCategories::get();
        let (caller, market_id) = setup_resolve_common_categorical::<T>(0, 0, categories)?;
```

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -225,26 +225,13 @@ fn setup_resolve_common_scalar_after_dispute<T: Config>(
...
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
+        let categories = T::MaxCategories::get();
+        let (caller, market_id) = setup_resolve_common_categorical::<T>(total_accounts, accounts_with_assets, categories)?;
+
+        let categories : u32 = categories.saturated_into();
+        for i in 0..categories.min(T::MaxDisputes::get()) {
             let origin = caller.clone();
             let disputes = crate::Disputes::<T>::get(market_id);
             let market = T::MarketCommons::market(&Default::default()).unwrap();
```
sea212, line 237: This will be taken care of in https://github.com/zeitgeistpm/zeitgeist/pull/812 , correct?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -371,35 +352,37 @@ benchmarks! {
...
-        let (_, market_id) = setup_resolve_common_categorical_after_dispute::<T>(a, b, c_u16)?;
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
+        let categories : u16 = T::MaxCategories::get().saturated_into();
+        let (_, market_id) =
+            setup_resolve_common_categorical_after_dispute::<T>(
+                total_accounts,
+                accounts_with_assets,
+                categories
+            )?;
```
sea212, line 363:
```suggestion
        let categories = T::MaxCategories::get();
        let (_, market_id) = setup_resolve_common_categorical_after_dispute::<T>(0, 0, categories)?;
```

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -371,35 +352,37 @@ benchmarks! {
...
-        for i in 0..c.min(d) {
+        let total_accounts = 1000;
+        let accounts_with_assets = 750;
+        let categories = T::MaxCategories::get();
+        let (caller, market_id) =
+            setup_resolve_common_categorical_after_dispute::<T>(
+                total_accounts,
+                accounts_with_assets,
+                categories,
+            )?;
```
sea212, line 382:
```suggestion
        let (caller, market_id) = setup_resolve_common_categorical_after_dispute::<T>(0, 0, categories)?;
```

sea212: I think we can remove the number of accounts from the helper function to avoid having to use `0, 0` as parameters always.

sea212 (changes requested):
The changes applied look fine, although what [Chris said about the swaps pallet](https://github.com/zeitgeistpm/zeitgeist/pull/806#pullrequestreview-1124830670) is correct. We have to delete the removal of pool shares there to avoid the iteration over every account.

zrml/prediction-markets/src/lib.rs
Chralt98: That sounds like a reasonable approach!

zrml/prediction-markets/src/benchmarks.rs
Chralt98: Or just use the following.

```rust
let categories: u16 = T::MaxCategories::get().saturated_into();
let (caller, market_id) = create_close_and_report_market::<T>(
    MarketCreation::Permissionless,
    MarketType::Categorical(categories),
    OutcomeReport::Categorical(categories.saturating_sub(1)),
)?;
```

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: I prefer to keep those parameters, as if we have better solution like using pallet-assets or anything else we can use it.

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: @Chralt98 ?

Chralt98: Correct 👍

zrml/prediction-markets/src/benchmarks.rs
Chralt98: I will take care of it in https://github.com/zeitgeistpm/zeitgeist/pull/812

zrml/swaps/src/lib.rs
```diff
@@ -1571,11 +1571,8 @@ mod pallet {
                 let amount = T::AssetManager::free_balance(asset, &pool_account);
                 T::AssetManager::slash(asset, &pool_account, amount);
             }
-            let pool_share_id = Self::pool_shares_id(pool_id);
-            let (_, liquidity_providers) =
-                T::AssetManager::accounts_by_currency_id(pool_share_id).unwrap_or((0usize, vec![]));
-            let _ =
-                T::AssetManager::destroy_all(pool_share_id, liquidity_providers.iter().cloned());
```
Chralt98, line 1578: Double check from @maltekliemann : No negative consequences here? I think the storage migration is possible the same way as you @sea212 explained [here](https://github.com/zeitgeistpm/zeitgeist/issues/792#issuecomment-1262115357) .

Chralt98 (commented):
Looks quite solid.

zrml/swaps/src/tests.rs
```diff
@@ -3080,8 +3080,6 @@ fn assert_all_parameters(
     assert_eq!(Currencies::free_balance(ASSET_B, &pai), pool_assets[1]);
     assert_eq!(Currencies::free_balance(ASSET_C, &pai), pool_assets[2]);
     assert_eq!(Currencies::free_balance(ASSET_D, &pai), pool_assets[3]);
-
-    assert_eq!(Currencies::total_issuance(psi), total_issuance);
```
sea212, line 3084: We mustn't remove that check, many tests rely on that. To fix the issue we have to fix the corresponding test (line 141+):
```rust
#[test]
fn destroy_pool_correctly_cleans_up_pool() {
        create_initial_pool_with_funds_for_alice(ScoringRule::CPMM, Some(0), true);
        let pool_id = 0;
        let alice_balance_before = [
            Currencies::free_balance(ASSET_A, &ALICE),
            Currencies::free_balance(ASSET_B, &ALICE),
            Currencies::free_balance(ASSET_C, &ALICE),
            Currencies::free_balance(ASSET_D, &ALICE),
        ];
        assert_ok!(Swaps::destroy_pool(pool_id));
        assert_err!(Swaps::pool(pool_id), crate::Error::<Runtime>::PoolDoesNotExist);
        // Ensure that funds _outside_ of the pool are not impacted!
        // TODO(#792): Remove pool shares.
        let total_pool_shares = Currencies::total_issuance(Swaps::pool_shares_id(0));
        assert_all_parameters(alice_balance_before, 0, [0, 0, 0, 0], total_pool_shares);
    });
}
```
I just changed the last three lines of that function

sea212 (changes requested):
Looks good overall with the exception that the current changes make many tests skip verification of a value.

sea212 (approved):
LGTM

Chralt98 (approved):
OK

maltekliemann:
> We still have an unregarded case for `destroy_pool` in `admin_destroy_market`.
> 
> https://github.com/zeitgeistpm/zeitgeist/blob/710cb62ace72b6eb975bc538b35717c00cbd4b38/zrml/swaps/src/lib.rs#L1575-L1582
> 
> There we got a weight dependence, because we iterate again over all accounts... We should probably fix that too in this in this PR, while we are at it. @maltekliemann What do you think about that?

> We still have an unregarded case for `destroy_pool` in `admin_destroy_market`.
> 
> https://github.com/zeitgeistpm/zeitgeist/blob/710cb62ace72b6eb975bc538b35717c00cbd4b38/zrml/swaps/src/lib.rs#L1575-L1582
> 
> There we got a weight dependence, because we iterate again over all accounts... We should probably fix that too in this in this PR, while we are at it. @maltekliemann What do you think about that?

We decided to remove basically all calls to `accounts_by_currency_id` and `destroy_all`, if I remember correctly to prevent. The lines in question destroy the liquidity tokens, which represent the LPs share of the pool. When the pool is destroyed, these are worthless, so keeping them around shouldn't be a problem. So my suggestion here would be to simply remove these calls.

Unless the fact that `destroy_pool` is only called in `admin_destroy_market` counts as a counter argument.

PR #807, opened 2022-09-28
Partially benchmark on_initialize (without on_resolution)

Chralt98:


zrml/prediction-markets/src/lib.rs
```diff
@@ -1225,6 +1235,15 @@ mod pallet {
                     },
                 );
 
+                if let Ok(weight) = open {
+                    total_weight = total_weight.saturating_add(weight);
+                } else {
```
vivekvpandya, line 1240: I don't understand this else part, so when open_market() fails then it goes to else here but then technically on_initialize() is failed? and why that means we need to charge for worst case? Same question for resolution and close cases.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -527,6 +535,145 @@ benchmarks! {
...
+                // noop, because weight is already measured somewhere else
+                Ok(())
+            },
+        ).unwrap();
+    }
+
+    on_initialize_top_overhead {
+        let now = 2u32;
+    }: {
+        if now <= 1u32 {}
```
vivekvpandya, line 643: isn't this a dead code as you initialize now = 2 above?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -527,6 +535,145 @@ benchmarks! {
             Ok(())
         })?;
     }: { Pallet::<T>::start_subsidy(&market_clone.unwrap(), market_id)? }
+
+    market_status_manager {
+        let b in 1..31;
+        let f in 1..31;
+        // let t in 0..10;
```
vivekvpandya, line 542: if not required please remove.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -527,6 +535,145 @@ benchmarks! {
...
+        for _ in 0..64 {
+            let _ = create_market_common::<T>(
+                MarketCreation::Permissionless,
+                MarketType::Categorical(T::MaxCategories::get()),
+                ScoringRule::CPMM
+            )?;
+        }
+
+        let block_number: T::BlockNumber = Zero::zero();
+        let last_time_frame: TimeFrame = Zero::zero();
```
vivekvpandya, line 554: not sure why last_time_frame is always zero?

vivekvpandya:
This seems to make weight for on_initialize() more precise :pray:

zrml/prediction-markets/src/benchmarks.rs
Chralt98: I would like ask, you, @maltekliemann and @sea212, if we like to benchmark the market_status_manager in the case, that the last time frame is far behind the current time frame [here](https://github.com/zeitgeistpm/zeitgeist/blob/30b7191666809c5f40293b0e815c9336d3a9ec91/zrml/prediction-markets/src/lib.rs#L2199-L2209) ?  I think as I benchmarked it currently, it's sufficient enough, because of this:

```rust
market_ids_per_time_frame_len = market_ids_per_time_frame_len
                    .saturating_add(market_ids_per_time_frame.len() as u32);
```

So if we have more than 1 iteration in the parent for-loop, when we simply put a higher benchmark component inside the weight function. 

If you agree, then I would delete the comments, as you suggested.

zrml/prediction-markets/src/benchmarks.rs
Chralt98: It's just an arbitrary [start value](https://github.com/zeitgeistpm/zeitgeist/blob/30b7191666809c5f40293b0e815c9336d3a9ec91/zrml/prediction-markets/src/lib.rs#L2200). More important is the `current_time_frame`, because my initial thought was to iterate only once over the parent for loop. So the `current_time_frame` should be `last_time_frame + 1 == current_time_frame` (assuming the for-loop is inclusive).

zrml/prediction-markets/src/benchmarks.rs
Chralt98: The initial thought was to simulate a near exact code as in the `on_initialize` to get an accurate weight representation. The purpose of this `if`, is that the comparison is benchmarked. I know, everybody assumes, that this comparison has a small effect on the weight, but again, I wanted to be as accurate as possible.

zrml/prediction-markets/src/lib.rs
Chralt98: Yeah, that's a good question. This is a conservative approach. We don't know when this failed [here](https://github.com/zeitgeistpm/zeitgeist/blob/30b7191666809c5f40293b0e815c9336d3a9ec91/zrml/prediction-markets/src/lib.rs#L2194-L2195) or [here](https://github.com/zeitgeistpm/zeitgeist/blob/30b7191666809c5f40293b0e815c9336d3a9ec91/zrml/prediction-markets/src/lib.rs#L2205-L2206). It could be in the last or first for-loop-iteration. The corrected weight is only returned in the `Ok(weight)` case, but not in early return errors. So it could be, that `market_status_manager` failed at the last for-iteration (thus taking lots of processing time).

maltekliemann:
Starting my review now...

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: this is interesting point, for my learning could you please explain substrate side issues for bench-marking nested for-loops in bit more detail?

Chralt98: Cool, that you ask. I did pick that up in this video [here](https://www.youtube.com/watch?v=Qa6sTyUqgek). I am not quite sure anymore, but I think the benchmarking pallet assumes, that each benchmark component scales linear with higher values. So for example, this is a weight function:

Weight_function = A*x + B*y

Assume x and y are the components. Now the benchmark is basically manipulating x while y is constant and after that manipulating y while x is constant.

But this weight function is not quadratic like this for example: A * x^2 + B * y

Substrates assumes always linear benchmarks.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -679,6 +686,151 @@ benchmarks! {
...
+        Pallet::<T>::resolution_manager(
+            now,
+            |market_id, market| {
+                // noop, because weight is already measured somewhere else
+                Ok(())
+            },
+        ).unwrap();
+    }
+
+    on_initialize_top_overhead {
```
maltekliemann, line 797: So, what's happening here is that you're benchmarking the top level of the function by recreating its layout but deleting certain parts of it, right? I don't know man... this looks like a nightmare to maintain.

It should be easier to do this by using dependency injection. You inject the managers which open/close/resolve markets. For benchmarking the top level component, you then inject noop managers which take (virtually) no computing time.

maltekliemann (changes requested):
I've just glanced over this PR. The solution for the top level of `on_initialize` scares me a little bit. I wouldn't go down that road unless necessary.

zrml/prediction-markets/src/benchmarks.rs
Chralt98: That's a good idea. Thanks a lot!

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -532,8 +538,9 @@ benchmarks! {
         let asset_accounts = 10u32;
         let d in 0..T::MaxDisputes::get();
 
-        let (caller, market_id) =
-            setup_resolve_common_scalar_after_dispute::<T>(total_accounts, asset_accounts)?;
+        let (caller, market_id) = setup_resolve_common_scalar_after_dispute::<T>(
+            total_accounts, asset_accounts
+        )?;
```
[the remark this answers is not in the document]
Chralt98: 60a00a2551240b808a6199e51a845ca9a3fc31dc

zrml/prediction-markets/src/benchmarks.rs
Chralt98: f7f47a5638c99e214ad2166c1d980e96f487bad9

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -679,6 +686,151 @@ benchmarks! {
             Ok(())
         })?;
     }: { Pallet::<T>::start_subsidy(&market_clone.unwrap(), market_id)? }
+
+    market_status_manager {
+        let b in 1..31;
+        let f in 1..31;
```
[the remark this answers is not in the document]
Chralt98: What do you exactly mean? The upper limit is set to 31.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -679,6 +686,151 @@ benchmarks! {
...
+                MarketCreation::Permissionless,
+                MarketType::Categorical(T::MaxCategories::get()),
+                ScoringRule::CPMM,
+                Some(MarketPeriod::Timestamp(range_start..range_end)),
+            )?;
+        }
+
+        let block_number: T::BlockNumber = Zero::zero();
+        let last_time_frame: TimeFrame = Zero::zero();
+        for i in 1..=b {
```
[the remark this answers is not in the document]
Chralt98: 5d484e6c2ee97db59b5d7bb78ee8c96e34c586ce Is this okay?

zrml/prediction-markets/src/lib.rs
```diff
@@ -2157,24 +2200,36 @@ mod pallet {
...
-            for market_id in MarketIdsPerBlock::get(block_number).iter() {
+            let market_ids_per_block = MarketIdsPerBlock::get(block_number);
+            let market_ids_per_block_len = market_ids_per_block.len() as u32;
+            for market_id in market_ids_per_block.iter() {
                 let market = T::MarketCommons::market(market_id)?;
                 mutation(market_id, market)?;
             }
             MarketIdsPerBlock::remove(block_number);
 
+            let mut market_ids_per_time_frame_len = 0u32;
```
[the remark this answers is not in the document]
Chralt98: 6264e43eaad2378b0b42ee455a1bbab35f7500e5

zrml/prediction-markets/src/lib.rs
```diff
@@ -2157,24 +2200,36 @@ mod pallet {
                 Query = BoundedVec<MarketIdOf<T>, CacheSize>,
             >,
         {
-            for market_id in MarketIdsPerBlock::get(block_number).iter() {
+            let market_ids_per_block = MarketIdsPerBlock::get(block_number);
+            let market_ids_per_block_len = market_ids_per_block.len() as u32;
```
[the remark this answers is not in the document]
Chralt98: 6264e43eaad2378b0b42ee455a1bbab35f7500e5

zrml/prediction-markets/src/lib.rs
```diff
@@ -2198,13 +2255,18 @@ mod pallet {
             MarketIdsPerReportBlock::<T>::remove(block);
 
             // Resolve any disputed markets.
-            for id in MarketIdsPerDisputeBlock::<T>::get(block).iter() {
+            let market_ids_per_dispute_block = MarketIdsPerDisputeBlock::<T>::get(block);
+            let market_ids_per_dispute_block_len = market_ids_per_dispute_block.len() as u32;
```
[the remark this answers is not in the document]
Chralt98: 6264e43eaad2378b0b42ee455a1bbab35f7500e5

zrml/prediction-markets/src/lib.rs
```diff
@@ -2183,13 +2238,15 @@ mod pallet {
...
-                return Ok(());
+                return Ok(0u64);
             }
 
             let block = now.saturating_sub(dispute_period);
 
             // Resolve all regularly reported markets.
-            for id in MarketIdsPerReportBlock::<T>::get(block).iter() {
+            let market_ids_per_report_block = MarketIdsPerReportBlock::<T>::get(block);
+            let market_ids_per_report_block_len = market_ids_per_report_block.len() as u32;
```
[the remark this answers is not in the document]
Chralt98: 6264e43eaad2378b0b42ee455a1bbab35f7500e5

zrml/prediction-markets/src/benchmarks.rs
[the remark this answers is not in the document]
Chralt98: f732286f54afe1cecb0057f5b0e4f273bc841d92

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -679,6 +686,151 @@ benchmarks! {
...
+        let range_start: MomentOf<T> = 100_000u64.saturated_into();
+        let range_end: MomentOf<T> = 1_000_000u64.saturated_into();
+        // ensure markets exist
+        for _ in 0..64 {
+            let _ = create_market_common::<T>(
+                MarketCreation::Permissionless,
+                MarketType::Categorical(T::MaxCategories::get()),
+                ScoringRule::CPMM,
+                Some(MarketPeriod::Timestamp(range_start..range_end)),
+            )?;
```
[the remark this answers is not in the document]
Chralt98: f732286f54afe1cecb0057f5b0e4f273bc841d92

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -51,7 +55,8 @@ fn create_market_common_parameters<T: Config>(
     permission: MarketCreation,
 ) -> Result<(T::AccountId, T::AccountId, MultiHash, MarketCreation), &'static str> {
     let caller: T::AccountId = whitelisted_caller();
-    T::AssetManager::deposit(Asset::Ztg, &caller, (u128::MAX).saturated_into()).unwrap();
+    T::AssetManager::deposit(Asset::Ztg, &caller, (100 * MinLiquidity::get()).saturated_into())
+        .unwrap();
```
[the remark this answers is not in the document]
Chralt98: I don't know why, but I did cargo fmt and it formatted in benchmarks now o.O? So it's changed now again.

zrml/prediction-markets/src/benchmarks.rs
Chralt98: 3aec79b4a0b530726c1af595056f6c832135ddeb

I actually saw an unused `on_initialize_resolve_overhead` and just used it.

zrml/prediction-markets/src/benchmarks.rs
Chralt98: Fixed with this 3aec79b4a0b530726c1af595056f6c832135ddeb

vivekvpandya (dismissed):

zrml/prediction-markets/src/benchmarks.rs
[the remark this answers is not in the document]
maltekliemann: I mean something like:

```rust
let f_max = 31;
let f in 1..f_max;

/* ... */

for _ in 0..f_max {

}
```

maltekliemann:
Starting my review now...

zrml/prediction-markets/src/benchmarks.rs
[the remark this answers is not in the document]
Chralt98: Unfortunately nothing can be above the benchmark component inside the bench setup.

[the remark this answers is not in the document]
maltekliemann: Yikes! Didn't realize that. Thanks.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1293,7 +1293,7 @@ mod pallet {
     }
 
     #[pallet::event]
-    #[pallet::generate_deposit(fn deposit_event)]
+    #[pallet::generate_deposit(pub(crate) fn deposit_event)]
```
maltekliemann, line 1296:
```suggestion
    #[pallet::generate_deposit(fn deposit_event)]
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1400,14 +1413,36 @@ mod pallet {
                     },
                 );
 
+                if let Ok(weight) = close {
+                    total_weight = total_weight.saturating_add(weight);
+                } else {
+                    // charge weight for the worst case
+                    total_weight = total_weight.saturating_add(
+                        T::WeightInfo::market_status_manager(CacheSize::get(), CacheSize::get()),
```
maltekliemann, line 1421: Of course, the worst case for `market_status_manager` is not `CacheSize::get()`; it's a very large difference between `current_time_frame` and `last_time_frame`, and each time frame holds `CacheSize::get()`. The larger the gap between the time frames, the larger the weight.

I don't think we should do this right now, but can you open an issue that we need to basically delay status changes until the next block if there are too many markets to handle? I would do it like this:

- Set a maximum weight for status changes `m`
- Benchmark the weight of a single status change `w`; if we need to handle more markets than `m / w` markets, we need to delay the status change until the next block

This basically boils down to having a priority system, where the weight `m` needs to be distributed between a couple of tasks (status changes, garbage collecting) and opening/closing is more important than resolving, which in turn is more important than garbage collection.

This shouldn't be difficult to implement, we just need to find the time to do it.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -783,6 +785,130 @@ benchmarks! {
...
+        for i in 1..=b {
+            <MarketIdsPerOpenBlock<T>>::try_mutate(block_number, |ids| {
+                ids.try_push(i.into())
+            }).unwrap();
+        }
+
+        let last_offset: TimeFrame = last_time_frame + 1.saturated_into::<u64>();
+        //* quadratic complexity should not be allowed in substrate blockchains
+        //* assume at first that the last time frame is one block before the current time frame
+        let t = 0;
```
maltekliemann, line 827: `t` is the number of additional frames?

Chralt98: `t` is the time frame span between last time frame and current time frame.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -679,6 +681,130 @@ benchmarks! {
...
+        for i in 1..=d {
+            // + 31 to not conflict with the markets of MarketIdsPerReportBlock
+            d_ids_vec.push((i + 31).into());
+        }
+        MarketIdsPerDisputeBlock::<T>::mutate(block_number, |ids| {
+            *ids = BoundedVec::try_from(d_ids_vec).unwrap();
+        });
+
+        // TODO(#730) this will get outdated with the merged PR
+        let now: T::BlockNumber = T::DisputePeriod::get() + block_number;
```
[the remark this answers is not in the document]
Chralt98: d532546c145700546272d1af3485035de43604e6

zrml/prediction-markets/src/lib.rs
```diff
@@ -1326,14 +1339,36 @@ mod pallet {
...
+                        total_weight.saturating_add(T::WeightInfo::market_resolution_manager(
+                            CacheSize::get(),
+                            CacheSize::get(),
+                        ));
+                }
+
                 LastTimeFrame::<T>::set(Some(current_time_frame));
+                total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
 
+                // hopefully the weight of the following is insignificant
```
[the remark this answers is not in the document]
Chralt98: 22fed0c2708cb22a336a574171f7f26e5da29617 just deleted the misleading comment. 

About your point: I benchmarked the `Ok(())` case with `on_initialize_resolve_overhead`. The only weight, that is missing is the following block:

```rust
Self::deposit_event(Event::BadOnInitialize);
log::error!("Block {:?} was not initialized. Error: {:?}", now, err);
TransactionOutcome::Rollback(err.into())
```

The rollback is just an enum variant return. The unknown for me is actually what happens in `with_transaction`, when we have the Rollback case. Documentation [here](https://paritytech.github.io/substrate/master/frame_support/storage/transactional/fn.with_transaction.html). In my eyes, it just looks like: Doing all the stuff in `with_transaction` and if there is no error, then actually apply those changes on the storage.

I mean the consumed weight is added anyways (total_weight accumulated in both cases Ok or Rollback). So we have a mostly accurate weight with the current solution.

zrml/prediction-markets/src/lib.rs
Chralt98: I see your point. I added your comment in this issue https://github.com/zeitgeistpm/zeitgeist/issues/817

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: Oh, that makes sense. So I guess it's fine the way it is :thumbsup:

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -783,6 +785,124 @@ benchmarks! {
...
+                Ok(())
+            })?;
+        }
+
+        let block_number: T::BlockNumber = Zero::zero();
+
+        let mut r_ids_vec = Vec::new();
+        for i in 1..=r {
+           r_ids_vec.push(i.into());
+        }
```
maltekliemann, line 878: [nitpick^2] I'd use the patterns from `d_ids_vec` here, as well (or vice versa - both are fine!)

zrml/prediction-markets/src/lib.rs
```diff
@@ -1385,6 +1389,10 @@ mod pallet {
                     },
                 );
 
+                total_weight = total_weight.saturating_add(open.unwrap_or_else(|_| {
+                    T::WeightInfo::market_status_manager(CacheSize::get(), CacheSize::get())
+                }));
```
maltekliemann, line 1394: [nitpick^2] Same thing here - I'd either use the `unwrap_or_else` pattern everywhere or nowhere in this function.

maltekliemann (approved):
:thumbsup:

vivekvpandya (approved):

PR #813, opened 2022-09-29
Fix on_resolution() benchmarks

vivekvpandya:


zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -397,47 +376,59 @@ benchmarks! {
...
-        let (caller, market_id) = setup_resolve_common_categorical_after_dispute::<T>(a, b, c_u16)?;
+        let (caller, market_id) =
+            setup_resolve_common_categorical::<T>(a, b, c_u16)?;
+        T::MarketCommons::mutate_market(&market_id, |market| {
+            let admin = account("admin", 0, 0);
+            market.dispute_mechanism = MarketDisputeMechanism::Authorized(admin);
+            Ok(())
+        })?;
 
         for i in 0..c.min(d) {
```
maltekliemann, line 387: I think this should iterate over `0..d` and dispute ~~`OutcomeReport::Categorical(i % c)`.~~ `OutcomeReport::Categorical(i % 2)` to maximize the number of correct reporters, which will result in addition `deposit` calls in `on_resolution`.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -397,47 +376,59 @@ benchmarks! {
...
+    } verify {
+        let market = T::MarketCommons::market(&market_id)?;
+        assert_eq!(market.status, MarketStatus::Resolved);
     }
 
     internal_resolve_scalar_reported {
         let total_accounts = 10u32;
         let asset_accounts = 10u32;
-        let (_, market_id) = setup_resolve_common_scalar_after_dispute::<T>(total_accounts, asset_accounts)?;
+    let (caller, market_id) = setup_resolve_common_scalar::<T>(total_accounts, asset_accounts)?;
```
maltekliemann, line 402:
```suggestion
        let (caller, market_id) = setup_resolve_common_scalar::<T>(total_accounts, asset_accounts)?;
```

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -397,47 +376,59 @@ benchmarks! {
...
+            let admin = account("admin", 0, 0);
+            market.dispute_mechanism = MarketDisputeMechanism::Authorized(admin);
+            Ok(())
+        })?;
         for i in 0..d {
-            let disputes = crate::Disputes::<T>::get(market_id);
             let origin = caller.clone();
-            let market = T::MarketCommons::market(&Default::default()).unwrap();
-            T::SimpleDisputes::on_dispute(&disputes, &market_id, &market)?;
+            Pallet::<T>::dispute(RawOrigin::Signed(origin).into(), market_id, OutcomeReport::Scalar(i.into()))?;
```
maltekliemann, line 424: You may want to add an `assert` which ensures that `d < range.end` where `range` is the scalar range.

```suggestion
            Pallet::<T>::dispute(
                RawOrigin::Signed(origin).into(),
                market_id,
                OutcomeReport::Scalar(i.into()),
            )?;
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1911,6 +1911,7 @@ mod pallet {
             };
             let clean_up_weight = Self::clean_up_pool(market, market_id, &resolved_outcome)?;
             total_weight = total_weight.saturating_add(clean_up_weight);
+            //TODO: following call should return weight consumed by it.
```
maltekliemann, line 1914: You should add an issue to this TODO so that we can properly track it.

maltekliemann (changes requested):

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: range is u128::MAX so hopefully not needed, and to get it from market_type need to unfold enum.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -347,62 +323,82 @@ benchmarks! {
...
-            let market = T::MarketCommons::market(&Default::default()).unwrap();
-            T::SimpleDisputes::on_dispute(&disputes, &market_id, &market)?;
+            Pallet::<T>::dispute(
+                RawOrigin::Signed(origin).into(),
+                market_id,
+                OutcomeReport::Categorical(i.saturated_into::<u16>()),
+            )?;
         }
     }: {
         let market = T::MarketCommons::market(&market_id)?;
```
Chralt98, line 357: Why do you call the market inside the effective benchmark?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -347,62 +323,82 @@ benchmarks! {
...
-        let disputes = crate::Disputes::<T>::get(market_id);
-        T::SimpleDisputes::on_resolution(&disputes, &market_id, &market)?
+        Pallet::<T>::on_resolution(&market_id, &market)?;
+    } verify {
+        let market = T::MarketCommons::market(&market_id)?;
```
Chralt98, line 360: The market should be queried in the setup {} (first) section one time.

Chralt98 (commented):
You should maximise for the number of correct_reporters as @maltekliemann pointed out. I wouldn't do a benchmark component out of it, I would just use the worst case for the benchmark. 
Apart from that, since there are many `unreserve_named`, `slash_reserved_named `, `deposit` I would probably verify and go in heavy paths and check for the differences (to get the worst case).

Last but not least: `clean_up_pool` depends on the number of assets of the associated pool. https://github.com/zeitgeistpm/zeitgeist/blob/b09f44bed399f8b37153e92a77a549f5a86deed8/zrml/swaps/src/lib.rs#L1449-L1458 

I give you an example how to setup the pool:

https://github.com/zeitgeistpm/zeitgeist/blob/2033329f866af2981c83686f79074f9e3790305d/zrml/prediction-markets/src/benchmarks.rs#L230-L261

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: in verify section I just wanted to assert that after call to on_resolution() market status is properly updated.

Chralt98: OK

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+        MarketCreation::Permissionless,
+        MarketType::Categorical(categories.saturated_into()),
+        ScoringRule::CPMM,
+        None,
+    )?;
+
+    let max_swap_fee: BalanceOf<T> = MaxSwapFee::get().saturated_into();
+    let min_liquidity: BalanceOf<T> = MinLiquidity::get().saturated_into();
+    let _ = Call::<T>::buy_complete_set { market_id, amount: min_liquidity }
+        .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
```
maltekliemann, line 214: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+        RawOrigin::Signed(caller.clone()).into(),
+        market_id,
+        max_swap_fee,
+        min_liquidity,
+        weights,
+    )
+    .unwrap();
+
+    let _ = Call::<T>::admin_move_market_to_closed { market_id }
+        .dispatch_bypass_filter(T::CloseOrigin::successful_origin())?;
```
maltekliemann, line 227: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+    let outcome = OutcomeReport::Categorical(1u16);
+    let _ = Call::<T>::report { market_id, outcome }
+        .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
```
maltekliemann, line 230: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+        MarketCreation::Permissionless,
+        MarketType::Scalar(0u128..=u128::MAX),
+        ScoringRule::CPMM,
+        None,
+    )?;
+
+    let max_swap_fee: BalanceOf<T> = MaxSwapFee::get().saturated_into();
+    let min_liquidity: BalanceOf<T> = MinLiquidity::get().saturated_into();
+    let _ = Call::<T>::buy_complete_set { market_id, amount: min_liquidity }
+        .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
```
maltekliemann, line 247: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
         market_id,
-        OutcomeReport::Scalar(1u128),
+        max_swap_fee,
+        min_liquidity,
+        weights,
     )
     .unwrap();
+
+    let _ = Call::<T>::admin_move_market_to_closed { market_id }
+        .dispatch_bypass_filter(T::CloseOrigin::successful_origin())?;
```
maltekliemann, line 262: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+    let outcome = OutcomeReport::Scalar(u128::MAX);
+    let _ = Call::<T>::report { market_id, outcome }
+        .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
```
maltekliemann, line 265: The `let _ =` pattern should be removed here. You're using the question mark operator, so you're already using the result.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -486,65 +523,98 @@ benchmarks! {
...
     }
 
     internal_resolve_categorical_disputed {
         // d = num. disputes
         let d in 0..T::MaxDisputes::get();
 
         let categories = T::MaxCategories::get();
-        let (caller, market_id) =
-            setup_resolve_common_categorical_after_dispute::<T>(0, 0, categories)?;
+        let (caller, market_id) = setup_reported_categorical_market_with_pool::<T>(categories.into())?;
```
maltekliemann, line 546:
```suggestion
        let (caller, market_id) =
            setup_reported_categorical_market_with_pool::<T>(categories.into())?;
```

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+    let weight_len: usize = MaxRuntimeUsize::from(categories).into();
+    let weights = vec![MinWeight::get(); weight_len];
+    Pallet::<T>::deploy_swap_pool_for_market(
+        RawOrigin::Signed(caller.clone()).into(),
+        market_id,
+        max_swap_fee,
+        min_liquidity,
+        weights,
+    )
+    .unwrap();
```
maltekliemann, line 224: You made a great point in some other PR that our helper functions are returning results, so we shouldn't use `unwrap()` here, and use the question mark operator instead.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -215,18 +198,72 @@ fn setup_resolve_common_scalar<T: Config>(
...
+    let weights = vec![MinWeight::get(); weight_len];
+    Pallet::<T>::deploy_swap_pool_for_market(
         RawOrigin::Signed(caller.clone()).into(),
         market_id,
-        OutcomeReport::Scalar(1u128),
+        max_swap_fee,
+        min_liquidity,
+        weights,
     )
     .unwrap();
```
maltekliemann, line 259: Same thing here: Let's not use `unwrap()` in a function which returns a `Result`.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -486,65 +523,98 @@ benchmarks! {
...
         let total_accounts = 10u32;
         let asset_accounts = 10u32;
-        let (_, market_id) =
-            setup_resolve_common_scalar_after_dispute::<T>(total_accounts, asset_accounts)?;
+        let (caller, market_id) = setup_reported_scalar_market_with_pool::<T>()?;
+        generate_accounts_with_assets::<T>(
+            total_accounts,
+            asset_accounts,
+            Asset::ScalarOutcome(market_id, ScalarPosition::Long),
+        )?;
```
maltekliemann, line 578: Why are we generating assets here? I wouldn't expect the resolution of a scalar market to depend on the number of accounts that hold the assets.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -486,65 +523,98 @@ benchmarks! {
...
     }: {
         let market = T::MarketCommons::market(&market_id)?;
```
maltekliemann, line 580: I think you want this to be in the scope above, otherwise you're benchmarking the `market` call, as well.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -486,65 +523,98 @@ benchmarks! {
...
+        T::MarketCommons::mutate_market(&market_id, |market| {
+            let admin = account("admin", 0, 0);
+            market.dispute_mechanism = MarketDisputeMechanism::Authorized(admin);
+            Ok(())
+        })?;
+        generate_accounts_with_assets::<T>(
+            total_accounts,
+            asset_accounts,
+            Asset::ScalarOutcome(market_id, ScalarPosition::Long),
+        )?;
```
maltekliemann, line 602: Same question here: What do we need these accounts for?

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -486,65 +523,98 @@ benchmarks! {
...
-            let market = T::MarketCommons::market(&Default::default()).unwrap();
-            T::SimpleDisputes::on_dispute(&disputes, &market_id, &market)?;
+            Pallet::<T>::dispute(
+                RawOrigin::Signed(origin).into(),
+                market_id,
+                OutcomeReport::Scalar(i.into())
+            )?;
         }
     }: {
         let market = T::MarketCommons::market(&market_id)?;
```
maltekliemann, line 613: This call should go into the scope above.

maltekliemann (changes requested):

zrml/prediction-markets/src/benchmarks.rs
Chralt98: Why not? Isn't it working or is it just a better style?

zrml/prediction-markets/src/benchmarks.rs
maltekliemann: I guess this is debatable since the call doesn't return a `DispatchResultWithPostInfo`, so consider this a nitpick.

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: So error is batter than painc right? as it can prpogate error from its origin, for example in this if something goes wrong in deploy_swap...() using ? is better than just unwrap() which may point that line 217 fails :thinking:

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: I was under impression of clean_up_pool() to be used from swap() however we have disabled removing account from swaps. So not required.

zrml/prediction-markets/src/benchmarks.rs
maltekliemann: Sure. It's a nitpick, but you're still _assuming_ that that's the case. Not important, though.

Chralt98 (changes requested):
Overall the `on_resolution` benchmarks look quite good. Just a small request change for ignoring scalar markets with clean_up_pool.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -454,11 +486,17 @@ benchmarks! {
             Ok(())
         })?;
 
+        let market = T::MarketCommons::market(&market_id)?;
+        if let MarketType::Scalar(range) = market.market_type {
+            assert!((d as u128) < *range.end());
```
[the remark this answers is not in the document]
vivekvpandya: https://github.com/zeitgeistpm/zeitgeist/pull/813#discussion_r983685224

Chralt98 (approved):
OK.

zrml/prediction-markets/src/benchmarks.rs
maltekliemann: Sorry, I guess the question wasn't totally clear. Even if we didn't disable the removing accounts, why do we need this when we're dealing with scalar markets (where we don't need to destroy accounts)?

vivekvpandya: yes I was not aware about that swaps does not remove accounts for scalar market. It was brought to my attention by @Chralt98 and I have already removed this code. Does it need anything more?

maltekliemann: No, it's all good. Just wanted to clarify what I meant.

Chralt98:
I will wait for the CI.

Chralt98 (approved):
OK again.

maltekliemann (approved):
:thumbsup:

zrml/prediction-markets/src/benchmarks.rs
maltekliemann: Whenever possible, I would use results. Delegates the responsibility for handling the error to the caller, allowing maximum flexibility and keeps the call stack clean.

PR #818, opened 2022-10-04
Fix `swaps` benchmark failures

maltekliemann:
When running the benchmarks of `swaps`, the ~~`pool_exit`~~ `pool_exit_with_exact_pool_amount` benchmark throws `MaxOutRatio` and `swap_exact_amount_out` throws an error because the maximum allowed total weight is exceeded. Both are caused by the recent increase of `MaxCategories`.

This PR fixes that, in addition to correcting some outstanding formatting errors in `swaps/benchmarks.rs`.

### DIFF
--- a/zrml/swaps/src/benchmarks.rs
+++ b/zrml/swaps/src/benchmarks.rs
@@ -37,7 +37,7 @@ use frame_system::RawOrigin;
 use orml_traits::MultiCurrency;
 use sp_runtime::traits::{SaturatedConversion, Zero};
 use zeitgeist_primitives::{
-    constants::BASE,
+    constants::{BASE, CENT},
     traits::Swaps as _,
     types::{
         Asset, Deadlines, Market, MarketCreation, MarketDisputeMechanism, MarketPeriod,
@@ -190,8 +190,7 @@ benchmarks! {
             pool.pool_status = PoolStatus::Closed;
             Ok(())
         });
-
-}: admin_clean_up_pool(RawOrigin::Root, market_id, OutcomeReport::Categorical(0))
+    }: admin_clean_up_pool(RawOrigin::Root, market_id, OutcomeReport::Categorical(0))
 
     admin_clean_up_pool_cpmm_scalar {
         let caller: T::AccountId = whitelisted_caller();
@@ -231,7 +230,6 @@ benchmarks! {
             pool.pool_status = PoolStatus::Closed;
             Ok(())
         });
-
     }: admin_clean_up_pool(RawOrigin::Root, market_id, OutcomeReport::Scalar(33))
 
     end_subsidy_phase {
@@ -253,11 +251,7 @@ benchmarks! {
         let amount = T::MinSubsidy::get();
 
         // Create b accounts, add MinSubsidy base assets and join subsidy
-        let accounts = generate_accounts_with_assets::<T>(
-            b,
-            a.saturated_into(),
-            amount,
-        ).unwrap();
+        let accounts = generate_accounts_with_assets::<T>(b, a.saturated_into(), amount).unwrap();
 
         // Join subsidy with each account
         for account in accounts {
@@ -284,11 +278,8 @@ benchmarks! {
         let amount = T::MinSubsidy::get();
 
         // Create a accounts, add MinSubsidy base assets and join subsidy
-        let accounts = generate_accounts_with_assets::<T>(
-            a,
-            min_assets_plus_base_asset,
-            amount,
-        ).unwrap();
+        let accounts =
+            generate_accounts_with_assets::<T>(a, min_assets_plus_base_asset, amount).unwrap();
 
         // Join subsidy with each account
         for account in accounts {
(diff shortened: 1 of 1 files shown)


Chralt98:
Mhhh, when I run the benchmark from the current `main` branch of `pool_exit` I don't get this error? 

```bash
./target/release/zeitgeist benchmark pallet --chain=dev --steps=1 --repeat=1 --pallet=zrml_swaps --extrinsic='pool_exit' --execution=wasm --wasm-execution=compiled --heap-pages=4096 --output=./zrml/swaps/src/weights2.rs --template=./misc/weight_template.hbs
```

Chralt98:
Mhh I wonder why this wasn't caught by the CI process. 

`cargo test --features runtime-benchmarks -p zrml-swaps` does not fail for the `impl_benchmark_test_suite`.

`test benchmarks::benchmark_tests::test_benchmarks ... ok` 

I assumed in my review, that the mentioned test would catch it, but since the mock and the real runtime parameters are so different, we didn't notice it.

Haha, here we have a classic example of mock failures. Nice, that we see know, that using mock values for benchmarks is probably not sufficient enough. I will create a CI script to solve that problem for the future.

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,7 +386,7 @@ benchmarks! {
             false,
             None,
         );
-        let asset_amount: BalanceOf<T> = BASE.saturated_into();
+        let asset_amount: BalanceOf<T> = CENT.saturated_into();
```
Chralt98, line 389: Now for the release runtime parameters we are below the following.

```rust
bmul(asset_balance.saturated_into(), T::MaxOutRatio::get().saturated_into())?.saturated_into()
```

Nice 💪 . So the reason for this failure was, that the `asset_amount` in this benchmark was set too high, for the actual runtime parameters.

zrml/swaps/src/benchmarks.rs
```diff
@@ -560,8 +558,8 @@ benchmarks! {
         .saturated_into();
         let weight_out = T::MinWeight::get();
         let weight_in = 4 * weight_out;
-        let mut weights = vec![weight_in; asset_count as usize];
-        weights[asset_count as usize - 1] = weight_out;
+        let mut weights = vec![weight_out; asset_count as usize];
+        weights[0] = weight_in;
```
Chralt98, line 562: This change seems complicated. We have now `asset_count - 1` times `MinWeight` and for the first weight `4 * MinWeight`. Instead of `asset_count - 1` times `4 * MinWeight` and for the last weight `MinWeight`. Can you elaborate on that a little bit?

Chralt98 (approved):
I did not detect any release benchmark failures in `zrml_swaps` now anymore after your changes.

I will change the CI to avoid this failures in the future.

maltekliemann:
> I see you mean `swap_exact_amount_out_cpmm` fails with `MaxTotalWeight` and 
`pool_exit_with_exact_pool_amount` fails with `MaxOutRatio`.

Apologies for the unclear description.

> Mhh I wonder why this wasn't caught by the CI process.
> 
> `cargo test --features runtime-benchmarks -p zrml-swaps` does not fail for the `impl_benchmark_test_suite`.
> 
> `test benchmarks::benchmark_tests::test_benchmarks ... ok`
> 
> I assumed in my review, that the mentioned test would catch it, but since the mock and the real runtime parameters are so different, we didn't notice it.
> 
> Haha, here we have a classic example of mock failures. Nice, that we see know, that using mock values for benchmarks is probably not sufficient enough. I will create a CI script to solve that problem for the future.

Does running `cargo test --features=runtime-benchmarks` actually execute the runtime benchmarks? I would be pretty surprised about that.

> I did not detect any release benchmark failures in `zrml_swaps` now anymore after your changes.
> 
> I will change the CI to avoid this failures in the future.

Sounds like a plan! :thumbsup:

zrml/swaps/src/benchmarks.rs
maltekliemann: Sure. This test raises `MaxTotalWeight`, which means that the total weight (which is the sum of all weights) is too large. So we need to make some weights smaller.

Previously, the weight vector was initialized as `[4, ..., 4]` and then the last weight, which is the weight of the outgoing asset, was set to `1`. This results in a large weight. But we only need the first (ingoing) and last (outgoing) asset for this test, so we don't care about all the weights in the middle.

Now we initialize the vector as `[1, ..., 1]` and only change the first weight (of the ingoing asset) to `4`, which results in a much lower weight, which, fortunately, is smaller than the maximum total weight.

zrml/swaps/src/benchmarks.rs
```diff
@@ -395,7 +386,7 @@ benchmarks! {
             false,
             None,
         );
-        let asset_amount: BalanceOf<T> = BASE.saturated_into();
+        let asset_amount: BalanceOf<T> = CENT.saturated_into();
         let pool_amount = 0u32.into();
     }: _(RawOrigin::Signed(caller), pool_id, assets[0], asset_amount, pool_amount)
```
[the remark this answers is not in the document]
maltekliemann: Yikes! Fixed in 9d040b0ea2e9d5c5f92b28c881b753144d2f9c50.

Chralt98:
> Does running cargo test --features=runtime-benchmarks actually execute the runtime benchmarks? I would be pretty surprised about that.

Yes, but it uses the configuration parameters of the provided mock ident from below.

https://github.com/zeitgeistpm/zeitgeist/blob/cadca835826737725f24b798612a27a9f2b095b9/zrml/swaps/src/benchmarks.rs#L607

You can see the execution of the benchmark tests below.

`test benchmarks::benchmark_tests::test_benchmarks ... ok`

This output happens at the bottom of each pallet test run.

Chralt98 (approved):
I checked the current extrinsics of swaps again against the real runtime and it works.

OK 👍

sea212 (approved):

PR #823, opened 2022-10-06
Remove toolchain: input so that rust-toolchain file will be used

vivekvpandya:
As  per https://github.com/marketplace/actions/rust-toolchain#the-toolchain-file
May close https://github.com/zeitgeistpm/zeitgeist/issues/780
As per discussion on https://lightrun.com/answers/actions-rs-toolchain-support-new-toml-format-for-rust-toolchain currently actions-rs has some issues using toolchain files. So used suggested solution there.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -33,10 +33,10 @@ jobs:
         uses: actions/checkout@v2
         if: github.event_name == 'workflow_dispatch'
 
-# TODO(#XXX): `rustup show` installs the rustc version specified in the `rust-toolchain` file
-# according to https://rust-lang.github.io/rustup/overrides.html. We don't use actions-rs due to the
-# lack of support of `rust-toolchain` files with TOML syntax:
-# https://github.com/actions-rs/toolchain/issues/126.
+      # TODO(#839): `rustup show` installs the rustc version specified in the `rust-toolchain` file
+      # according to https://rust-lang.github.io/rustup/overrides.html. We don't use actions-rs due
+      # to the lack of support of `rust-toolchain` files with TOML syntax:
+      # https://github.com/actions-rs/toolchain/issues/126.
       - name: Install rust toolchain
         run: rustup show
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -36,7 +36,6 @@ jobs:
       - name: Install rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
           target: wasm32-unknown-unknown
           default: true
           override: true
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -34,13 +34,7 @@ jobs:
         if: github.event_name == 'workflow_dispatch'
 
       - name: Install rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          target: wasm32-unknown-unknown
-          default: true
-          override: true
-          profile: minimal
-          components: rustfmt, clippy
+        run: rustup show
 
       - name: Cache dependencies
         uses: Swatinem/rust-cache@v1
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -33,6 +33,10 @@ jobs:
         uses: actions/checkout@v2
         if: github.event_name == 'workflow_dispatch'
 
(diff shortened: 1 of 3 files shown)


Chralt98 (commented):
Ok, the "rustup show" looks like a dirty hot fix 😂. As I understand https://github.com/marketplace/actions/rust-toolchain#the-toolchain-file correctly, they just say, that the `toolchain` input can be removed to use [this file](https://github.com/zeitgeistpm/zeitgeist/blob/main/rust-toolchain). 

> In these cases the toolchain can be named in the project's directory in a file called rust-toolchain.toml or rust-toolchain.

When leaving out `toolchain` in the workflow files doesn't work, maybe rename the root file to `rust-toolchain.toml` and try again.

vivekvpandya:
> Ok, the "rustup show" looks like a dirty hot fix joy. As I understand https://github.com/marketplace/actions/rust-toolchain#the-toolchain-file correctly, they just say, that the `toolchain` input can be removed to use [this file](https://github.com/zeitgeistpm/zeitgeist/blob/main/rust-toolchain).
> 
> > In these cases the toolchain can be named in the project's directory in a file called rust-toolchain.toml or rust-toolchain.
> 
> When leaving out `toolchain` in the workflow files doesn't work, maybe rename the root file to `rust-toolchain.toml` and try again.

Not sure why that will work? I see moonbeam also using the hack "rustup show" https://github.com/PureStake/moonbeam/blob/master/.github/workflows/build.yml

Chralt98:
Ok, so you tried it without the toolchain input? I just thought: Why shouldn't it work, when `actions-rs/toolchain` directly explains, that it works with a toolchain file. But sure, we can go with `rustup show`. I asked, because this little hack could be misunderstood later.

Chralt98:
I researched a little bit and found this issue comment https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659 . So yeah, sounds like this hack is used at the moment.

Chralt98 (dismissed):
OK

maltekliemann:
I don't get it. You provide this link: https://github.com/marketplace/actions/rust-toolchain#the-toolchain-file. Why use this workaround?

vivekvpandya:
> I don't get it. You provide this link: https://github.com/marketplace/actions/rust-toolchain#the-toolchain-file. Why use this workaround?

It does not work as advertise for now, there is a bug and not fixed yet.

maltekliemann:
This issue? https://github.com/actions-rs/toolchain/issues/126

vivekvpandya:
> This issue? [actions-rs/toolchain#126](https://github.com/actions-rs/toolchain/issues/126)

Yes it seems so, here is relevant PR https://github.com/actions-rs/toolchain/pull/166

.github/workflows/benchmark.yml
```diff
@@ -33,15 +33,12 @@ jobs:
         uses: actions/checkout@v2
         if: github.event_name == 'workflow_dispatch'
 
+# TODO(#XXX): `rustup show` installs the rustc version specified in the `rust-toolchain` file
+# according to https://rust-lang.github.io/rustup/overrides.html. We don't use actions-rs due to the
+# lack of support of `rust-toolchain` files with TOML syntax:
+# https://github.com/actions-rs/toolchain/issues/126.
```
maltekliemann, line 39: Can you please fill in the XXX with the issue for this, and add this to the other two places where these changes are made? Oh, also I forgot to add indentation to the comment.

```suggestion
      # TODO(#XXX): `rustup show` installs the rustc version specified in the `rust-toolchain` file
      # according to https://rust-lang.github.io/rustup/overrides.html. We don't use actions-rs due
      # to the lack of support of `rust-toolchain` files with TOML syntax:
      # https://github.com/actions-rs/toolchain/issues/126.
```

maltekliemann (changes requested):

.github/workflows/benchmark.yml
vivekvpandya: Oh I missed that :sweat:

vivekvpandya: Should we create issue on our git repo?

Chralt98 (approved):
OK

.github/workflows/benchmark.yml
maltekliemann: No worries. Yeah, just so we don't forget about removing the workaround when we get the chance.

maltekliemann (approved):
:thumbsup:

PR #811, opened 2022-09-28
Upgrade codebase to polkadot-v0.9.23

sea212:
Closes #809
Closes #715 

This PR updates the dependencies.

**Additional features**
- Use original repositories instead of forks (previously many forks from Moonbeam were used)
- Use references in `get_spot_price` and corresponding RPCs
- Implement Moonbeam's VRFs (generate randomness) in the client, add it to `author_rotateKeys` rpc and to `pallet-author-mapping`
- Build benchmarks from parachain modules only when parachain modules are activated by another feature
- Use `jsonrpsee` instead of `jsonrpc_core`
- Use patched Purestake libraries to ensure all modules are fetched from original repositories and make `try-runtime` work for `pallet-author-mapping`

**Notes**
Executing `try-runtime` will show decoding errors for `pallet-author-mapping`. This happens after `pre_upgrade` and `on_runtime_upgrade` and can be ignored - the storage is correctly migrated, as the `try-runtime` checks prove. In addition, the runtime upgrade was tested on a Battery Station clone. Unfortunately nodes have to upgrade their clients to be able to determine the author of the first block after the runtime upgrade.

PR #829, opened 2022-10-06
Bump versions to v0.3.6

maltekliemann:


sea212 (approved):
The versions in *Cargo.lock* should also updated. I'll add those in https://github.com/zeitgeistpm/zeitgeist/pull/811.

maltekliemann:
> The versions in _Cargo.lock_ should also updated. I'll add those in #811.

My bad, should have known that.

PR #832, opened 2022-10-12
Fix new clippy lints

Chralt98:


node/Cargo.toml
```diff
@@ -10,14 +10,20 @@ pallet-transaction-payment = { branch = "moonbeam-polkadot-v0.9.19", git = "http
 pallet-transaction-payment-rpc = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
 pallet-transaction-payment-rpc-runtime-api = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
 sc-basic-authorship = { branch = "moonbeam-polkadot-v0.9.19", git = "https://github.com/purestake/substrate" }
-sc-cli = { branch = "moonbeam-polkadot-v0.9.19", features = ["wasmtime"], git = "https://github.com/purestake/substrate" }
+sc-cli = { branch = "moonbeam-polkadot-v0.9.19", features = [
```
vivekvpandya, line 13: Is this because of newer version of taplo ?

Chralt98: Maybe, I just did `taplo fmt`

maltekliemann: You may have to update taplo. This shouldn't happen with the current config. I think taplo just ignores config options that it doesn't know.

Chralt98: I updated via `cargo install taplo-cli --locked` and still get the same result after `taplofmt`

maltekliemann: Yeah, but if you format `main`, you get no diff. We allow expanding arrays even if they are too long, so taplo doesn't fix the line cited above. I guess this is somewhat close to actually being a bug. Report it, if you like.

Chralt98: OK, I reverted the changes here 1eb464da5aa71adc1d9076053ef51ac2ab70a885

sea212 (approved):

PR #834, opened 2022-10-13
Add extrinsics for editing markets

### LINKED ISSUE #742: [prediction-markets] Allow metadata changes in pending advised markets
The purpose of this is to allow users to make changes according to the Advisory Committee's comments.

vivekvpandya:
This PR should close #742


maltekliemann:
Starting my review now...

maltekliemann (changes requested):
Nice, I like the thorough testing. Some things, though:

- The changelog for developers needs to be updated.
- I think you should also test that `request_edit` fails if the market is not `Proposed` (I may have missed that)

I'm also a bit worried about letting people change the market period. This creates an exploit for getting around slashing: Let's say I create a market which I don't know how to fix, I get asked by the committee to edit and so I set the market period to something like `current_block..next_block`. And then the market expires and I get my deposit back. I admit this is a very contrived example. But still...

I guess one solution is to release the advisory bond when an edit request is made, but that opens up a whole new can of worms. The alternative is to slash the advisory bond of expired markets. Seems fair to me. I'll bring this up at the Advisory Committee meeting tomorrow and let you know what the others think.

Chralt98:
I wait for the apply of the review suggestion from Malte.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -389,7 +403,66 @@ benchmarks! {
...
+                Pallet::<T>::calculate_time_frame_of_moment(range_end),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
+        for i in 0..m {
+            MarketIdsPerOpenTimeFrame::<T>::try_mutate(
+                Pallet::<T>::calculate_time_frame_of_moment(range_start),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
```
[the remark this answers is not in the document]
vivekvpandya: Yes I think we don't need `MarketIdsPerOpenTimeFrame` as it is related to pool deployment. 
The market period  is already timestamp based.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -410,9 +410,11 @@ benchmarks! {
 
         let market_type = MarketType::Categorical(T::MaxCategories::get());
         let dispute_mechanism = MarketDisputeMechanism::SimpleDisputes;
-        let scoring_rule =ScoringRule::CPMM;
-        let range_end = T::MaxSubsidyPeriod::get();
-        let range_start = T::MinSubsidyPeriod::get();
+        let scoring_rule = ScoringRule::CPMM;
+        // let range_end = T::MaxSubsidyPeriod::get();
+        // let range_start = T::MinSubsidyPeriod::get();
```
maltekliemann, line 625:
```suggestion
```

zrml/prediction-markets/src/benchmarks.rs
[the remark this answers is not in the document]
maltekliemann: And what do we need `MarketIdsPerCloseTimeFrame` for?

maltekliemann (changes requested):

zrml/prediction-markets/src/lib.rs
```diff
@@ -639,28 +634,64 @@ mod pallet {
...
+            let sender = ensure_signed(origin)?;
+            ensure!(
+                MarketIdsForEdit::<T>::contains_key(market_id),
+                Error::<T>::MarketEditNotRequested
+            );
+            let old_market = T::MarketCommons::market(&market_id)?;
+            ensure!(old_market.creator == sender, Error::<T>::EditorNotCreator);
+            ensure!(old_market.status == MarketStatus::Proposed, Error::<T>::InvalidMarketStatus);
+
+            Self::clear_auto_close(&market_id)?;
```
maltekliemann, line 667: When we change `market.period`, we also need to take changes to `period.start()` into account - in other words, we need to call `clear_auto_open(&market_id)`.

maltekliemann (changes requested):

zrml/prediction-markets/src/lib.rs
```diff
@@ -282,25 +280,35 @@ mod pallet {
...
         pub fn request_edit(
             origin: OriginFor<T>,
             #[pallet::compact] market_id: MarketIdOf<T>,
-            edit_reason: EditReason,
+            edit_reason: Vec<u8>,
         ) -> DispatchResult {
+            let edit_reason: EditReason<T> = edit_reason
+                .try_into()
+                .map_err(|_| Error::<T>::EditReasonLengthExceedsMaxEditReasonLen)?;
             T::ApproveOrigin::ensure_origin(origin)?;
```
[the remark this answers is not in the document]
vivekvpandya: I thought only advisory committee can request changes in market? But I have no issues adding one more Origin type.

Chralt98:
Starting my review now..

zrml/prediction-markets/src/lib.rs
vivekvpandya: I am bit confuse, AFAIU we can deploy pool for active markets only, so not sure how auto_open storage will have anything for market which is in proposed state?

Chralt98 (commented):
Cool, Vivek, with your help, the markets can now be edited in the `Proposed` state. As long as the markets are in the proposed state I feel confident, that this enhancement is safe.

Apart from Malte's suggestion, I've some smaller fixes. PS: And I love the tests too!

zrml/prediction-markets/src/lib.rs
```diff
@@ -131,6 +137,7 @@ mod pallet {
             };
             if market_status == MarketStatus::Proposed {
                 slash_market_creator(T::AdvisoryBond::get());
+                MarketIdsForEdit::<T>::remove(market_id);
```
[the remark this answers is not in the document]
vivekvpandya: you mean to create a new case for Proposed market in benchmark `admin_destroy_market()` ?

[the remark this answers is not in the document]
vivekvpandya: if we do that how to use it with .max() for weight above?

zrml/prediction-markets/src/lib.rs
maltekliemann: Yes, you guys are right, `MarketIdsPerOpen*` is only modified when a pool is deployed, so we're all good. :thumbsup:

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: That's corrrect, but the Advisory Committee has multiple origins to allow maximum configurability. For example, the `ApproveOrigin` currently requiries one member, while the `RejectOrigin` requires at least 50% of votes. So adding a new origin for requesting edits seems the reasonable thing to do.

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
Chralt98: Oh this was just an open thought. Because I don't think, that remove reads the value of the storage, it's okay to ignore it in the benchmarks. But if anyone of you know why, and that this remove actually should be regarded in the benchmarks, feel free to correct me. 

So the question is:

MarketIdsForEdit = [0, 1, 2]

MarketIdsForEdit::remove(3)

Does this different in case of the weight than this MarketIdsForEdit::remove(1)? 

Especially when behind 1 is a big list of length MaxEditReasonLen.

You could create a benchmark path for Proposed markets, yeah, but I don't think that it exceeds the weight of  reported or disputed markets. And because of each market status other than reported and disputed it takes the Default weight, so the maximum weight of reported and disputed I could live with as it is now. The admin_destroy_market is only callable by DestroyOrigin anyways, so it's even more  irrelevant. So, forget my comment. I just wanted to figure out, what actually happens for remove in the cases explained above.

Chralt98 (commented):
Just a quick explanation of the events and the extrinsics added to the changelog_for_devs file is missing.

Chralt98 (dismissed):
OK

maltekliemann:
Starting my review now...

zrml/prediction-markets/src/lib.rs
```diff
@@ -351,6 +362,40 @@ mod pallet {
...
+        /// Edit requested to proposed market from advisory committee.
+        /// It requires market_id and edit_reason (a `Vec<u8>`).
+        /// NOTE: Can only be called by the `ApproveOrigin`.
+        /// # Weight
+        ///
+        /// Complexity: `O(1)`
+        #[pallet::weight((
+                T::WeightInfo::request_edit(edit_reason.len() as u32),
+                Pays::No
+            ))]
```
maltekliemann, line 374:
```suggestion
        #[pallet::weight((
            T::WeightInfo::request_edit(edit_reason.len() as u32),
            Pays::No,
        ))]
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1492,10 +1492,10 @@ mod pallet {
         MarketReported(MarketIdOf<T>, MarketStatus, Report<T::AccountId, T::BlockNumber>),
         /// A market has been resolved \[market_id, new_market_status, real_outcome\]
         MarketResolved(MarketIdOf<T>, MarketStatus, OutcomeReport),
-        /// A proposed market has been requested edit by advisor. \[market_id, edit_reason]
+        /// A proposed market has been requested edit by advisor. \[market_id, edit_reason\]
         MarketRequestedEdit(MarketIdOf<T>, EditReason<T>),
-        /// A proposed market has been edited by a user. \[market_id, editor\]
-        MarketEdited(MarketIdOf<T>, <T as frame_system::Config>::AccountId),
+        /// A proposed market has been edited by the market creator \[market_id\]
+        MarketEdited(MarketIdOf<T>),
```
maltekliemann, line 1498: Chris made a good point about the event composition which I didn't catch earlier. I think it may be necessary to actually add the `MarketOf<T>` parameter here to make the indexer's job a little easier.

```suggestion
        MarketEdited(MarketIdOf<T>, MarketOf<T>),
```

maltekliemann (changes requested):
~~I'm also having trouble finding a test for the scenario where someone other than the market creator attempts to edit a market.~~ Nevermind, found it.

zrml/prediction-markets/src/lib.rs
```diff
@@ -639,28 +634,64 @@ mod pallet {
...
+            let ids_amount: u32 = Self::insert_auto_close(&market_id)?;
 
             Self::deposit_event(Event::MarketCreated(market_id, market_account, market));
 
             Ok(Some(T::WeightInfo::create_market(ids_amount).saturating_add(extra_weight)).into())
         }
 
+        #[pallet::weight(T::WeightInfo::edit_market(CacheSize::get()))]
+        #[transactional]
+        pub fn edit_market(
```
maltekliemann, line 646: Can you add a docstring here, as well?

zrml/prediction-markets/src/lib.rs
```diff
@@ -639,28 +640,82 @@ mod pallet {
...
+        /// * `period`: MarketPeriod to edit market.
+        /// * `deadlines`: Deadlines to edit market.
+        /// * `metadata`: MultiHash metadata to edit market.
+        /// * `market_type`: MarketType to edit market.
+        /// * `dispute_mechanism`: MarketDisputeMechanism to edit market.
+        /// * `scoring_rule`: ScoringRule to edit market.
+        ///
+        /// # Weight
+        ///
+        /// Complexity: `O(1)`
```
maltekliemann, line 667:
```suggestion
        /// Complexity: `O(n)`, where `n` is the number of markets which end at the same time as
        /// the market before the edit.
```

maltekliemann (approved):
:thumbsup: Just one correction in the documentation, no need to do a re-review.

Chralt98 (approved):
OK

PR #835, opened 2022-10-13
Add reject_reason in MarketRejected event

### LINKED ISSUE #795: When the advisory board rejects a market there is no way to give feedback to the market creator
This is frustrating to market creators

vivekvpandya:
This PR is intended to close #795


zrml/prediction-markets/src/lib.rs
```diff
@@ -907,12 +909,13 @@ mod pallet {
         pub fn reject_market(
             origin: OriginFor<T>,
             #[pallet::compact] market_id: MarketIdOf<T>,
+            reject_reason: RejectReason,
```
sea212, line 912: Maybe it makes sense to use `Vec` here like in [pallet_bounties](https://github.com/paritytech/substrate/blob/0ee03277c33b6334ddba7434e014fa637dcb6107/frame/bounties/src/lib.rs#L816-L822).
The app that crafts the transaction can include a vector of arbitrary size anyways, only when the transaction is executed on-chain and the parameter is [decoded the size is checked](https://github.com/paritytech/substrate/blob/a47f200eebeb88a5bde6f1ed2be9728b82536dde/primitives/core/src/bounded/bounded_vec.rs#L279-L291). Consequently it is the same like including a `Vec` as a parameter, except the following:
1. We can use a custom error that can be pinned exactly to the location where it happened.
2. Using a `Vec` makes crafting the parameter easier for the application

I propose we use `Vec`, convert it in the dispatchable, create a custom error and create a new `Config` item, that contains the maximum length `MaxRejectReasonLen` and that can be queried by the app that crafts the transaction.

zrml/prediction-markets/src/tests.rs
```diff
@@ -2491,8 +2499,10 @@ fn reject_market_fails_on_approved_market() {
...
         simple_create_categorical_market(MarketCreation::Advised, 0..1, ScoringRule::CPMM);
         assert_ok!(PredictionMarkets::approve_market(Origin::signed(SUDO), 0));
+        let reject_reason: RejectReason =
+            vec![0; 1024].try_into().expect("BoundedVec conversion failed");
         assert_noop!(
-            PredictionMarkets::reject_market(Origin::signed(SUDO), 0),
+            PredictionMarkets::reject_market(Origin::signed(SUDO), 0, reject_reason),
             Error::<Runtime>::InvalidMarketStatus
         );
     });
```
sea212, line 2508: We don't have that yet but I think it would be nice to see that events are thrown properly, i.e. that the rejection event contains all expected elements. Would also be nice to see that the extrinsic fails when the reason is too long.

[Here is some example to check events](https://github.com/paritytech/substrate/blob/0ee03277c33b6334ddba7434e014fa637dcb6107/frame/balances/src/tests.rs#L733-L756) or in case you want to reset event history and check the order of all new events, [take a look at this snippet](https://github.com/paritytech/substrate/blob/5a8f3e0a9cb45198b67c4eeba20062085a9381ce/frame/balances/src/tests.rs#L758-L852)

sea212 (commented):
I think we could go with the current implementation, however there are some potential improvements that we should discuss.

sea212:
An update of *docs/changelog_for_devs.md* is also required.

maltekliemann (changes requested):
Looks good. :thumbsup:

Added a couple of nitpicks and a question about the benchmarking. I think it would be prudent to parametrize the benchmark of `reject_market` by the length of `reject_reason`.

Also, we should add a note to `docs/changelog_for_devs.md`.

zrml/prediction-markets/src/lib.rs
```diff
@@ -907,12 +908,16 @@ mod pallet {
...
             #[pallet::compact] market_id: MarketIdOf<T>,
+            reject_reason: Vec<u8>,
         ) -> DispatchResultWithPostInfo {
             T::RejectOrigin::ensure_origin(origin)?;
             let market = T::MarketCommons::market(&market_id)?;
             let open_ids_len = Self::clear_auto_open(&market_id)?;
             let close_ids_len = Self::clear_auto_close(&market_id)?;
-            Self::do_reject_market(&market_id, market)?;
+            let reject_reason = reject_reason
+                .try_into()
```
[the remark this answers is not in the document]
vivekvpandya: isn't that O(1). As per following because Vec's len() is O(1).
```rust
  27 impl<T, S: Get<u32>> TryFrom<Vec<T>> for BoundedVec<T, S> {                                                                                  
  26     type Error = Vec<T>;                                                                                                                     
  25     fn try_from(t: Vec<T>) -> Result<Self, Self::Error> {                                                                                    
  24         if t.len() <= Self::bound() {                                                                                                        
  23             // explicit check just above                                                                                                     
  22             Ok(Self::unchecked_from(t))                                                                                                      
  21         } else {                                                                                                                             
  20             Err(t)                                                                                                                           
  19         }                                                                                                                                    
  18     }                                                                                                                                        
  17 } 
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -978,21 +979,38 @@ mod pallet {
         /// and `m` is the number of market ids,
         /// which close at the same time as the specified market.
         #[pallet::weight((
-            T::WeightInfo::reject_market(CacheSize::get(), CacheSize::get()),
+            T::WeightInfo::reject_market(
+                CacheSize::get(),
+                CacheSize::get(),
+                T::MaxRejectReasonLen::get()
```
maltekliemann, line 985:
```suggestion
                reject_reason.len() as u32,
```

This ensures that a non-advisor calling this function with a huge vector pays a large fee up front.

docs/changelog_for_devs.md
```diff
@@ -1,3 +1,9 @@
+# v0.3.7
+
+- `reject_market` extrinsic now requires `reject_reason` parameter which is `Vec<u8>`.
+   The config constant `MaxRejectReasonLen` defines maximum length of above parameter.
+   `MarketRejected` event also contains `reject_reason` so that it can be cached for market creator.
```
maltekliemann, line 5: Can you run `prettier -w docs/changelog_for_devs.md` on this?

maltekliemann (changes requested):
Tiny change request, then we're good to go. :thumbsup:

maltekliemann (approved):
:thumbsup:

What do you guys think about the range of `r` in the benchmarks? If we go up to 1024, then we have four million executions of this particular benchmark. I have no clue if that's too much.

maltekliemann (approved):
:thumbsup:

sea212 (approved):
LGTM

sea212:
> What do you guys think about the range of `r` in the benchmarks? If we go up to 1024, then we have four million executions of this particular benchmark. I have no clue if that's too much.

This is fine @maltekliemann . The `steps` option in the benchmark command will limit the number of iterations significantly. Currently we use `steps = 10` for Zeitgeist pallets, which results in `r` taking the values `i * r_max/10`, where `i` is the current iteration. Given the current configuration we have 4000 benchmark runs per extrinsic.

Issue #822, opened 2022-10-06
[pallet_collective] Benchmark failure for `propose_proposed`


Chralt98:
Running following command:

```bash
./target/release/zeitgeist benchmark pallet --chain dev --execution=wasm --wasm-execution=compiled --heap-pages=4096 --pallet "pallet_collective" --extrinsic "propose_proposed" --steps 1 --repeat 0
```

I get `Error: Input("DuplicateProposal")`. It seems like a configuration value is not set correctly for `pallet_collective`.

I investigate some time for it and also to check inside the CI, that all release benchmarks work correctly.

- [x] Correct pallet_collective config parameters

So it fails [here](https://github.com/paritytech/substrate/blob/174735ea1bb5fc4513519c45181d8df63d86f613/frame/collective/src/lib.rs#L520-L523) (for v0.9.19). And [here](https://github.com/paritytech/substrate/blob/174735ea1bb5fc4513519c45181d8df63d86f613/frame/collective/src/benchmarking.rs#L186-L201) is the benchmark. We only add a benchmark for the `AdvisoryCommittee` [here](https://github.com/zeitgeistpm/zeitgeist/blob/cadca835826737725f24b798612a27a9f2b095b9/runtime/common/src/lib.rs#L1116). The `MaxProposals` for the AdvisoryCommittee are set to `300` on [battery-station](https://github.com/zeitgeistpm/zeitgeist/blob/cadca835826737725f24b798612a27a9f2b095b9/runtime/battery-station/src/parameters.rs#L61) and [main-net](https://github.com/zeitgeistpm/zeitgeist/blob/cadca835826737725f24b798612a27a9f2b095b9/runtime/zeitgeist/src/parameters.rs#L61).

Since the creation of proposals produce [hashes with `u8`](https://github.com/paritytech/substrate/blob/174735ea1bb5fc4513519c45181d8df63d86f613/frame/collective/src/benchmarking.rs#L199), the number of `300` for `MaxProposals` is too high, because `u8` will only produce unique hashes for `255` proposals. 

Created an issue for that [here](https://github.com/paritytech/substrate/issues/12433).
 
- [x] Update the CI to quick check benchmarks

I would suggest the following command:

```bash
./target/release/zeitgeist benchmark pallet --chain dev --execution=native --wasm-execution=compiled --heap-pages=4096 --pallet "*" --extrinsic "*" --steps 1 --repeat 0 --detailed-log-output
```

sea212:
[...]

PR #836, opened 2022-10-14
Improve logging of try-runtime makefile targets

sea212:
Using native execution leads to more specific error descriptions and a backtrace that pinpoints the line that caused the error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,24 @@ check:
 check-dummy:
 	BUILD_DUMMY_WASM_BINARY= cargo check
 
+# Pseudo private target is invoked by public targets for different chains
+--try-runtime:
+	RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace \
+		cargo run \
+		--features=parachain,try-runtime \
+		--bin=zeitgeist \
+		try-runtime \
+		--execution=Native \
+		--chain=${TRYRUNTIME_CHAIN} \
+		on-runtime-upgrade \
+		live \
+		--uri=${TRYRUNTIME_URL}
+
 try-runtime-upgrade-battery-station:
-	cargo run --bin=zeitgeist --features=parachain,try-runtime try-runtime --execution=Native on-runtime-upgrade live --uri wss://bsr.zeitgeist.pm:443
+	@$(MAKE) TRYRUNTIME_CHAIN="battery_station" TRYRUNTIME_URL="wss://bsr.zeitgeist.pm:443" -- --try-runtime
 
 try-runtime-upgrade-zeitgeist:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,11 +15,24 @@ check:
 check-dummy:
 	BUILD_DUMMY_WASM_BINARY= cargo check
 
+# Pseudo private target is invoked by public targets for different chains
+--try-runtime:
+	RUST_LOG=runtime=trace,try-runtime::cli=trace,executor=trace \
+		cargo run \
```
Chralt98, line 21: Why do you use the `dev` profile in this case? (no --release) Is it, because you would like to allow debug_assertions? I could imagine, that the runtime speed might be too slow for complex storage migrations?

Makefile
```diff
@@ -15,11 +15,24 @@ check:
...
+		--bin=zeitgeist \
+		--features=parachain,try-runtime \
+		try-runtime \
+		--execution=Native \
```
Chralt98, line 25: Yeah with `Native` we have no `wasm::stripped` cases anymore, but more helpful error messages instead. 👍 I noticed it while benchmarking stuff. That's why I used this (`./target/release/zeitgeist benchmark pallet --chain dev --execution=native --wasm-execution=compiled --heap-pages=4096 --pallet "*" --extrinsic "*" --steps 1 --repeat 0 --detailed-log-output`) command to quick check the benchmarks. https://github.com/zeitgeistpm/zeitgeist/issues/822

Makefile
```diff
@@ -15,11 +15,24 @@ check:
...
+		live \
+		--uri=${TRYRUNTIME_URL}
+
 try-runtime-upgrade-battery-station:
-	cargo run --release --bin=zeitgeist --features=parachain,try-runtime try-runtime on-runtime-upgrade live --uri wss://bsr.zeitgeist.pm:443
+	@$(MAKE) TRYRUNTIME_CHAIN="battery_station_staging" TRYRUNTIME_URL="wss://bsr.zeitgeist.pm:443" -- --try-runtime
 
 try-runtime-upgrade-zeitgeist:
-	cargo run --release --bin=zeitgeist --features=parachain,try-runtime try-runtime on-runtime-upgrade live --uri wss://zeitgeist-rpc.dwellir.com
+	@$(MAKE) TRYRUNTIME_CHAIN="zeitgeist_staging" TRYRUNTIME_URL="wss://zeitgeist-rpc.dwellir.com:443" -- --try-runtime
```
Chralt98, line 35: Can you briefly explain, why and what the effect of using `staging` chains means? So I think staging is a pre-release chain specification?

Chralt98 (commented):
Just a couple of questions.

PR #838, opened 2022-10-16
Transform scalar markets from integer to fixed point numbers

### LINKED ISSUE #800: Allow negative numbers and fractions in scalar markets
Pretty much immediately after launch it seemed to me that we need more flexibility for scalar markets. If you bet on the change of some value in percentage ("How much will Tesla stock price change in Q3?"), then you need negative numbers, unless you want to do something absolutely terrible like: "0 to 100 means -100% to 0% and 100 to 200 corresponds to 0% to 100%."

Something similar goes for having fractions. Otherwise, if you want precision up to 0.1%, you have to do something like: "0 to 1000 means 0% to 100% (in equal increments)."

So, let's replace the `u128` type used to bet on scalar markets with:

```rust
enum ScalarValue {
    Integer(i128),
    Fraction(i128, u128),
}
```

We can also use a full-blown implementation of on-chain rational numbers for the last variant, but I think that's overkill, since we do almost no math with the scalar value.

**Edit.** Looking at this now, how about we replace it with `typedef ScalarFraction = (i128, u128)`?

maltekliemann:
Fixes #800. This PR consists of a storage migration only. That may seem confusing.

The storage migration transforms the numbers used in scalar markets to fixed point numbers by multiplying them by `BASE`. This basically changes the resolution of the market. If the market's values were USD before, they are now fixed point numbers with ten digits after the decimal point, still representing USD.

No additional changes to the business logic are required. Resolving scalar markets has worked before, it still works now. We're just transforming the markets that are currently on chain to a higher precision.


Chralt98:
There is only a conflict with the migrations in frame_executive::Executive. So I will start my review now.

Chralt98:
I am thinking about the current maximal price we could derive.

`340282366920938463463374607431768211455 / 10000000000 approx.= 3.40 × 10^28` 

Ok, seems big enough.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -15,259 +15,328 @@
...
     fn pre_upgrade() -> Result<(), &'static str> {
+        // Check that no saturation occurs.
+        for (market_id, market) in T::MarketCommons::market_iter().drain() {
+            if let MarketType::Scalar(range) = market.market_type {
+                assert!(
+                    range.end().checked_mul(BASE).is_some(),
+                    "TransformScalarMarketsToFixedPoint: Arithmetic overflow when transforming \
+                     market {:?}",
+                    market_id,
+                );
```
Chralt98, line 158: Nice overflow check here. 🥇 Edge case checker medal

I don't think this hits, but I wonder what we should do if this would happen in an imperfect world.

zrml/prediction-markets/src/migrations.rs
```diff
@@ -15,259 +15,328 @@
...
 
-        log::info!("Completed storage migration of MigrateMarketIdsPerBlockStorage");
-        total_weight
+            let market_after = Markets::<Runtime>::get(market_id).unwrap();
+            assert_eq!(market_after, market);
+
+            let disputes_after = Disputes::<Runtime>::get(market_id);
+            assert_eq!(disputes_after.len(), 1);
+            assert_eq!(disputes_after[0], dispute);
+        });
```
Chralt98, line 315: I don't really get it. Where do you skip what and how here?

Chralt98 (changes requested):
Nice, a useful migration, Malte, for higher precision by a redenomination of scalar markets.

We should speak to the front end and in the changelog_for_devs file about this. I could imagine, that without the notice, the front end might show very high USD prices. So for every `OutcomeReport::Scalar` or `MarketType::Scalar` input for extrinsics, this needs to be regarded by the front-end.

On our node side, each storage element with `OutcomeReport::Scalar` or `MarketType::Scalar` needs to be transformed. You did it for `Markets` in market-commons. You did it for `Disputes` in prediction-markets.

I see the following missing.

- `AuthorizedOutcomeReports` in authorized contains `OutcomeReport`
- `Votes` in court contains `OutcomeReport`

zrml/prediction-markets/src/migrations.rs
maltekliemann: To be honest, this test (almost) doesn't really do anything useful because someone could create a market that saturates between us running the test and deploying the update. If the PR wasn't already reviewed, I would remove that test.

If this test did discover any markets that saturate, we would just go through with the migration anyways. The on-chain democracy can decide whether or not they think it's a good idea to do this. The effect on the market is straightforward: The scalar range is restricted, which is good for the longs and bad for the shorts.

But there's one thing that shouldn't happen: Our scalar resolution assumes that `range.start() != range.end()`. I don't the time look it up, but I'm assuming that if `range.start() == range.end()`, then LONG and SHORT redeem for 1 ZTG, which is bad.

zrml/prediction-markets/src/migrations.rs
maltekliemann: Good lord, that diff... 😐

Yeah, that test name is terrible. What I'm testing here is that the migration is noop if the version numbers are incorrect.

zrml/prediction-markets/src/migrations.rs
Chralt98: > because someone could create a market that saturates between us running the test and deploying the update

But all other existing markets are checked. 

> we would just go through with the migration anyways. 

Doesn't `pre_upgrade` run before the `on_runtime_upgrade` and stops the migration if one assertion fails? 

> The effect on the market is straightforward: The scalar range is restricted, which is good for the longs and bad for the shorts.

The upper bound is limited. In case of an overflow, wouldn't be prices lower then and therefore it might be better for shorts instead of longs? 

> but I'm assuming that if range.start() == range.end(), then LONG and SHORT redeem for 1 ZTG, which is bad.

OK, so we should check for that?

maltekliemann: I've already added a check for `range.start() != range.end()`. Should have had that from the start, instead of checking for saturation.

The `pre_upgrade`/`post_upgrade` methods are only part of the `OnRuntimeUpgrade` trait `try-runtime` is enabled. They are only meant for testing purposes.

maltekliemann: > The upper bound is limited. In case of an overflow, wouldn't be prices lower then and therefore it might be better for shorts instead of longs?

Not really, the price resolution is based on where the bounds of the range are. So let's say you own a SHORT token for a market of range `[100, 200]`. Let's say this market would resolve to 175. That means your SHORT token would be worth .25 ZTG. Now suppose the the range is cut to `[100, 150]`. Now the market still resolves to 175, but that's the same as the market resolving to `150`, which means SHORT is worthless.

On the other hand, LONG would have been worth .75 ZTG before the change and 1 ZTG after.

Chralt98: Ok, I read about how scalar markets work and understand the math behind it now. Long is just the percentage from the start of the range to the regarded (maybe resolved or current) price on the range. Short is just 1 minus the Long price.

> So let's say you own a SHORT token for a market of range [100, 200]. Let's say this market would resolve to 175. That means your SHORT token would be worth .25 ZTG.

`long_price = (range_price - range_start) / (range_end - range_start)`

`long_price = (175 - 100) / (200 - 100) = 75 / 100 = 0.75`

`short_price = 1 - long_price = 1 - 0.75 = 0.25`

> Now suppose the the range is cut to [100, 150]. Now the market still resolves to 175, but that's the same as the market resolving to 150, which means SHORT is worthless.

I see you are right 👍  Thanks for the explanation again.

zrml/prediction-markets/src/migrations.rs
Chralt98: Just a dumb question: Where do you set the version numbers in this test case?

Chralt98: Because you did not use `set_up_chain();` ?

Chralt98: That's hard to understand, but because it's only a transitory test it's okay.

zrml/prediction-markets/src/migrations.rs
vivekvpandya: This is very informative discussion for me. :pray: 
if range.start() == range.end() then wouldn't there be some divide by zero math error in out code base somewhere, as chris's formula above?

zrml/prediction-markets/src/migrations.rs
```diff
@@ -15,259 +15,403 @@
...
+{
+    fn on_runtime_upgrade() -> Weight
     where
         T: Config,
     {
-        let mut total_weight = T::DbWeight::get().reads(1);
-        let storage_version = utility::get_on_chain_storage_version_of_market_commons_pallet();
-        if storage_version != MARKET_COMMONS_REQUIRED_STORAGE_VERSION {
-            log::info!("Skipping updates of markets; prediction-markets already up to date");
+        // TODO Use dependency pallets instead.
```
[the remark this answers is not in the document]
maltekliemann: No, sorry, just forgot to remove that.

maltekliemann:
> Is it wise to update some tests like following, just to show that to get actual value 0 and 100 we need to multiply it with BASE ?
> 
> ```rust
> simple-disputes/src/tests.rs
> 37:    market_type: MarketType::Scalar(0..=100),
> 79:            MarketDispute { at: 0, by: 0, outcome: OutcomeReport::Scalar(0) },
> 80:            MarketDispute { at: 0, by: 0, outcome: OutcomeReport::Scalar(20) },
> ```

Sorry, not quite sure what you mean. I'm always for having more tests, but doesn't this test everything necessary? https://github.com/zeitgeistpm/zeitgeist/blob/e0aa57f51d7ad599aebe07d4419ddafabed2f8bc/zrml/prediction-markets/src/migrations.rs#L319

zrml/prediction-markets/src/migrations.rs
maltekliemann: Yeah, the point is that `setup_chain` creates a situation in which the update is applied. By default, the pallet versions are so that the update will not trigger, so not using `setup_chains` means that update doesn't happen. I've made the test a little clearer, hope that helps.

Chralt98 (changes requested):
Brilliant example for a storage migration. `migrations.rs` archived ✅ for my first own storage migration.

Just a small change requests for the weights of the storage writes.

Because of the transitory behaviour of storage migrations, but they have a huge impact on the chain, they are so dangerous. I don't mean yours in particular here, but just wanted to point that out in general. I know, each new feature in a blockchain could be full of flaws, but with a storage migration, one can literally unconsciously ignore assumptions, that were made earlier in the runtime logic.

runtime/common/src/lib.rs
```diff
@@ -74,10 +68,7 @@ macro_rules! decl_common_types {
             frame_system::ChainContext<Runtime>,
             Runtime,
             AllPalletsWithSystem,
-            (
-                zrml_prediction_markets::migrations::UpdateMarketsForDeadlines<Runtime>,
-                zrml_prediction_markets::migrations::MigrateMarketIdsPerBlockStorage<Runtime>,
-            )
```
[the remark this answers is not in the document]
maltekliemann: > Ok, now I ask 😁: Are these migrations already executed? Migrations get executed when a runtime upgrade happens. So with every release cycle. So Harald makes the `0.3.6` release from [this commit](https://github.com/zeitgeistpm/zeitgeist/commit/8c6395fc152f04b49e1e6866ae0bbfbb8df5dfd3) (which includes the deleted migrations here), right?
> 
> And this PR is part of `0.3.7`. So we collect all coming storage migrations into this `frame_executive` for `0.3.7`, okay.

Correct.

> Another question: The second `type Executive` is when the feature parachain is not activated. So, it is required for each migrations (which has nothing to do with the parachain stuff like pallet_parachain_staking) to pass it in too? So this is the standalone mode, in case Zeitgeist looses the parachain slot? So it's better for each migration to execute it for the standalone mode too. But, when Zeitgeist always would have a parachain slot, this would be not necessary?

First of all, nothing speaks against making sure that the migrations execute on the standalone version, as well. If a migration for, say, `market-commons` was only executed on parachains, I would be very confused. We should also try to treat the development of Zeitgeist as if other developers are using our code to run chains, so executing the upgrade on standalone networks is often necessary.

maltekliemann:
Migration passes `try-runtime` test on Battery Station.

vivekvpandya:
> > Is it wise to update some tests like following, just to show that to get actual value 0 and 100 we need to multiply it with BASE ?
> > ```rust
> > simple-disputes/src/tests.rs
> > 37:    market_type: MarketType::Scalar(0..=100),
> > 79:            MarketDispute { at: 0, by: 0, outcome: OutcomeReport::Scalar(0) },
> > 80:            MarketDispute { at: 0, by: 0, outcome: OutcomeReport::Scalar(20) },
> > ```
> 
> Sorry, not quite sure what you mean. I'm always for having more tests, but doesn't this test everything necessary?
> 
> https://github.com/zeitgeistpm/zeitgeist/blob/e0aa57f51d7ad599aebe07d4419ddafabed2f8bc/zrml/prediction-markets/src/migrations.rs#L319-L374

no what I mean is when we want to write 100 , we have to write 100 * BASE, so if we can just update the test to show that ?
Just a suggestion!

maltekliemann:
> no what I mean is when we want to write 100 , we have to write 100 * BASE, so if we can just update the test to show that ? Just a suggestion!

Oh, I understand. I was going to do that, but then didn't because I didn't want to do this:

```
fn f(x: i32) -> i32 {
    x * 100
}

#[test]
fn test_f() {
    let value = 123;
    assert_eq!(f(value), value * 100);
}
```

This basically means that we test `f` against the logic of `f`. In some tests, it's not bad to automate a little, but in this case, I thought it was cleaner to specify the exact values that we want.

Chralt98 (dismissed):
OK

vivekvpandya (approved):

Chralt98 (approved):
OK

PR #841, opened 2022-10-20
Update weights

sea212:
This PR adds weights generated by Zeitgeist's benchmark machine. Two benchmarks could not be executed:
1. `pallet-parachain-staking`: Due to invalid verification logic the benchmarks failed. [A patch was applied](https://github.com/zeitgeistpm/moonbeam/commit/42fdf1a90294c49ddbf8d0204ef002b2914a886f), whose code is provided in a later version of `pallet-parachain-staking`.
2. `pallet-author-inherent`: The benchmark returns an error. Even when building the latest version of Moonbeam, which is from the same org (Purestake) like Nimbus (that contains `pallet-author-inherent`), it fails. A fix was not supplied yet. Our weights currently contain a copy of the weights they provided. This is fine, since this extrinsic is only executed once, fees are not charged and it only need a couple of microseconds to execute (even if we would be off by a margin, a stronger CPU could still process the block in time easily). @Chralt98 This affects https://github.com/zeitgeistpm/zeitgeist/pull/825, as it will fail now. I propose instead of benchmarking all pallets, the CI executes the `scripts/benchmark/run_benchmarks.sh` while overriding the ENV variables used in `scripts/benchmark/configuration.sh` to use low steps & repetitions. This required adjusting the `configuration.sh` script.

The benchmark `pallet-parachain-staking` is currently happening, the weight file will be provided later that day. At that point, the PR should be ready for review.


Chralt98:
> I propose instead of benchmarking all pallets, the CI executes the scripts/benchmark/run_benchmarks.sh while overriding the ENV variables used in scripts/benchmark/configuration.sh to use low steps & repetitions.

Addressed in https://github.com/zeitgeistpm/zeitgeist/pull/825/commits/98834b9e028518e0281ebace446371924d6de022

runtime/common/src/weights/frame_system.rs
```diff
@@ -49,7 +49,7 @@ use frame_support::{
 pub struct WeightInfo<T>(PhantomData<T>);
 impl<T: frame_system::Config> frame_system::weights::WeightInfo for WeightInfo<T> {
     fn remark(_b: u32) -> Weight {
-        (50_310_000 as Weight)
+        (0 as Weight)
```
sea212, line 52:
```suggestion
        (50_310_000 as Weight)
```

sea212:
Should be ready for review now

vivekvpandya (approved):

Chralt98 (approved):
WOW, cool, when looking at the unchanged logic I can notice an approx. 40% difference in weights with this production profile.

About `author_inherent`, stack trace [here](https://github.com/zeitgeistpm/zeitgeist/actions/runs/3290124026/jobs/5422542060) 

So [this file](https://github.com/zeitgeistpm/zeitgeist/blob/sea212-update-weights/runtime/common/src/weights/pallet_author_inherent.rs) is unchanged ok.
It's the benchmark `kick_off_authorship_validation`.

It fails [here](https://github.com/PureStake/nimbus/blob/51eaa2cdc905c8f2cbf6d468c82e194d91e13ffc/pallets/author-inherent/src/lib.rs#L138-L142). The [`T::SlotBeacon::slot()`](https://github.com/PureStake/nimbus/blob/51eaa2cdc905c8f2cbf6d468c82e194d91e13ffc/nimbus-primitives/src/lib.rs#L73-L104) seems to return zero, although it was set in the benchmarks [here](https://github.com/PureStake/nimbus/blob/51eaa2cdc905c8f2cbf6d468c82e194d91e13ffc/pallets/author-inherent/src/benchmarks.rs#L27).

`set_block_number` has an empty default implementation [here](https://paritytech.github.io/substrate/master/src/sp_runtime/traits.rs.html#1893). 
So I assume the problem here is, that [`set_block_number`](https://github.com/PureStake/nimbus/blob/51eaa2cdc905c8f2cbf6d468c82e194d91e13ffc/nimbus-primitives/src/lib.rs#L91) was not properly implemented on the moonbeam nimbus side.

> [This is useful in case the block number provider is different than System](https://paritytech.github.io/substrate/master/src/sp_runtime/traits.rs.html#1891)

So the fix is to implement `set_block_number` for `BlockNumberProvider`:

```rust
impl BlockNumberProvider {
      fn current_block_number() -> Self {
	      frame_system::Pallet<Config>::block_number()
      }

      fn set_block_number(slot: u32) -> Self {
	      frame_system::Pallet<Config>::set_block_number(slot)
      }
}
```

PR #846, opened 2022-10-25
Filter dangerous calls

sea212:
Filters potentially dangerous calls on Zeitgeist mainnet. I left out Battery Station because we only have the superuser there and might want to experiment with those functions.

To block:

`advisoryCommitteeCollective` && `council` && `technicalCommittee`
  - `setMembers` (mandatory) - Membership is managed by the respective Membership instance. Directly setting the members desynchronizes both pallets membership view.
  
`balances`:
  - `setBalance` (optional) - Balances should not be set. All newly generated tokens be minted by well know and approved processes, like staking. However, this could be used in some cases to fund system accounts like the parachain sorveign account in case something goes terribly wrong (like a hack that draws the funds from such an account, see Maganta hack). Invoking this function one can also easily mess up consistency in regards to reserved tokens and locks.
  - `forceTransfer` (optional) -- There should be no reason to force an account to transfer funds.
  
`assetManager`:
  - `updateBalance` (optional) - see reason for `balance.setBalance`
  
  
`dmpQueue`:
  - `serviceOverweight` (mandatory) - Executing this call with a maliciously crafted XCM can halt the chain.

`proxy`: 
  - `anonymous` (mandatory) - Makes `reserve` inconsistent, see https://github.com/paritytech/substrate/blob/37cca710eed3dadd4ed5364c7686608f5175cce1/frame/proxy/src/lib.rs#L270-L271
  - `kill_anonymous` (mandatory) - See `anonymous`

`system`:
  - `killPrefix` (mandatory) - Some "waste" storage will never impact proper operation. Cleaning up storage should be done by pallets or independent migrations.
  - `killStorage` (mandatory) - See `system.killPrefix`
  - `setCode` (mandatory) - A parachain uses parachain_system to enact and authorized a runtime upgrade. This ensure proper synchronization with the relay chain. Calling `setCode` will wreck the chain.
  - `setCodeWithoutChecks` (mandatory) - See `system.setCode`
  - `setStorage` (mandatory) -- Setting the storage directly is a dangerous operation that can lead to an inconsistent state. There might be scenarios where this is helpful, however, a well reviewed migration is better suited for that.
   
   
`vesting`:
  - `forceVestedTransfer` (optional) - There should be no reason to force vested transfer.

### DIFF
--- a/runtime/zeitgeist/src/lib.rs
+++ b/runtime/zeitgeist/src/lib.rs
@@ -98,6 +98,17 @@ pub struct IsCallable;
 // dispute mechanism.
 impl Contains<Call> for IsCallable {
     fn contains(call: &Call) -> bool {
+        use cumulus_pallet_dmp_queue::Call::service_overweight;
+        use frame_system::Call::{
+            kill_prefix, kill_storage, set_code, set_code_without_checks, set_heap_pages,
+            set_storage,
+        };
+        use orml_currencies::Call::update_balance;
+        use pallet_balances::Call::{force_transfer, set_balance};
+        use pallet_collective::Call::set_members;
+        use pallet_proxy::Call::{anonymous, kill_anonymous};
+        use pallet_vesting::Call::force_vested_transfer;
+
         use zeitgeist_primitives::types::{
             MarketDisputeMechanism::{Court, SimpleDisputes},
             ScoringRule::RikiddoSigmoidFeeMarketEma,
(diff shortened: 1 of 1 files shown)


vivekvpandya:
Looks good, just there is some format issues as indicated by CI. Also is this exhaustive list from all pallets being used in our runtime?

maltekliemann:
Starting my review now...

maltekliemann (commented):
No complaints about the code. However, a few notes on the changes made:

- You may want to have force_transfer available for something like this: https://kusama.polkassembly.io/referendum/240
- Regarding `anonymous` and `kill_anonymous`, if we're going to go down that road we should also check that nobody has already created pure proxies. I'm not too worried about making the reserved amount inconsistent. I don't think you can steal funds, because calling remove_proxy means you immediately lose control over the account. I'm just wondering why it is considered intended behavior to call remove_proxy from an anonymous account.
- I'm on the fence regarding `set_storage`. I could name one case off the top of my head where using `set_storage` to change an approved advised market using governance might be a good idea (market closes way too early). Doing that via a storage migration is strange, because we're adding a change that is very specific to the Zeitgeist chain to the codebase, and we're lumping it in with a bunch of other changes made for the next update. On the other hand, I agree that calling `set_storage` isn't exactly an awesome way of handling the problem either.

Furthermore, a general comment. With the exception of the `pallet-proxy` calls, these are all admin functions, so the main use of this PR is to limit the power one has through referenda and council motions in case of a hostile takeover, correct? I think I would find the failure of governance to prevent this takeover to be more concerning.

sea212:
@vivekvpandya Yes, this list should be exhaustive.

@maltekliemann Thanks for opening this discussion. The initial idea was that some calls are just plain wrong in our context and should never be executed, like `system.setCode` and `system.setCodeWithoutChecks`, because those are used for standalone chains. Those could be called accidentally and just wreck our chain (`parachainSystem.authorizeUpgrade` and `parachainSystem.enactAuthoritzedUpgrade` should be used for that). In addition, every collective has a `setMembers` call, which is invalid because it will desync the memberlist between the `membership` instance of the collective and the collective itself (`pallet-membership` handles the members of collectives). All the others functions are debatable, I just scanned through everything and left my thoughts about those.

In regards to all other functions, the idea is not to reduce the damage in case someone overtakes the chain - once this happened, a runtime upgrade can be deployed instantaneously that allows the attackers to do anything. The idea behind those restrictions is to avoid bringing the chain to an inconsistent/corrupt state. Disabling those calls should force the caller (most of the time global token holder vote) to use a less potentially harmful approach, for example use a treasury proposal instead of `force_transfer` out of the treasury. Not at all times are all the implications clear, especially for operations such as directly setting the storage or removing storage.

I think there is no way to check if anonymous/pure proxy accounts have been created other than checking the events, because the storage entries do not indicate that. I am fine to leave that one out.

In regards to `set_storage`, I am not convinced to keep it. The question is, why is it possible in the first place to get into a state where a root call of `set_storage` is necessary? Maybe it is not avoidable. I agree that migrations that handle such specific cases are nonsense. However, in that case the pallet should provide proper privileged functions to modify the storage as intended, without any potential unforeseen side effects that are caused through direct storage manipulation.

maltekliemann:
I agree with the sentiment that `set_storage` is a bad idea. Making the business logic more robust to solve problems which would otherwise require a storage migration sounds like a plan. Same goes for disabling the membership calls. I'm okay with disabling `force_transfer`, as well.

I guess the big question is what "invalid state" actually means. Due to the global nature of the substrate storage system, any pallet can break the specification of any other pallet in some form of another. The `remove_proxies` is a great example.

maltekliemann (approved):

vivekvpandya (approved):

sea212:
@maltekliemann What I meant by "invalid state" is if the state is not updated completely as required by the respective module. For example, if we increase the active collator set size using governance, it will set a storage value for the number of active collators. But in addition to this, it will also rearrange the bottom and top collator storage and do some other stuff required for proper operation. Setting those value directly would lead to an inconsistent state, because the variable containing the active collator set size was updated but other dependent storage was not.

PR #860, opened 2022-10-31
Allow MDMs to directly resolve the market 

### LINKED ISSUE #851: ZIP-0 Let the MDM control the resolution
Coupled with https://github.com/zeitgeistpm/zeitgeist/issues/738 

For `authorized` allow a correction period, so that the authority is able to correct mistakes (fat-fingered wrong outcome).

Chralt98:


Chralt98:
Because each MDM needs to resolve in the future, `on_initialize` is required for this. When we handle the timing in each MDM, we might end up with the situation, that we duplicate code (essentially the current handling of `MarketIdsPerDisputeBlock` in each MDM). So, I am in favour of using the current pm functionality and allow the MDM to control the `MarketIdsPerDisputeBlock` storage. https://github.com/zeitgeistpm/zeitgeist/pull/862

For `authorized` we have a correction period, which is much smaller than the dispute period. So we shouldn't really execute `on_resolution` with the extrinsic `authorize_market_outcome`, because there is the possibility, that the authority made a mistake.

PR #866, opened 2022-11-03
Update MarketIdsPer*Block benchmarks

Chralt98:
I noticed, that we forgot to update a specific benchmark case as we restructured the resolution manager.


vivekvpandya (approved):
Thanks for fixing this!

sea212 (approved):

PR #871, opened 2022-11-11
Run benchmark tests in separate environments

maltekliemann:
@Chralt98 and I noticed that failing on benchmark tests can cause failures in other benchmark tests, which seemed to be either a bug or some undefined behavior: https://github.com/zeitgeistpm/zeitgeist/pull/682#discussion_r1002726731

It is actually caused by `impl_benchmark_test_suite!` executing all tests in the same test environment, which means that state is not reset if a test fails, which can cause subsequent tests to fail if they are based on the assumption that we're starting with a blank chain state: https://paritytech.github.io/substrate/master/src/frame_benchmarking/lib.rs.html#1257-1283

We can fix this by putting the `impl_benchmark_test_suite!` call _inside_ the `benchmarks` macro invocation: https://paritytech.github.io/substrate/master/src/frame_benchmarking/lib.rs.html#1285-1325


sea212 (approved):
Good catch guys.

Chralt98 (approved):
Wow, thanks for the deep investigation. That's useful!

PR #874, opened 2022-11-11
Remove `report_todo` from `rustfmt` config

maltekliemann:


sea212 (approved):
Finally no more `report_todo` spam during compilation. Now the structure of todo comments is not enforced anymore, correct? Is there no alternative, i.e. do we have to keep order now by convention?

maltekliemann:
> Finally no more `report_todo` spam during compilation. Now the structure of todo comments is not enforced anymore, correct? Is there no alternative, i.e. do we have to keep order now by convention?

There doesn't seem to be an alternative: https://github.com/rust-lang/rustfmt/pull/5357. I've added checking for unnumbered TODOs to the checklist for reviews: https://github.com/zeitgeistpm/zeitgeist/pull/895.

PR #877, opened 2022-11-15
Unreserve dispute bonds when destroying a market

### LINKED ISSUE #686: [prediction-markets] `admin_destroy_market` doesn't clear dispute bonds
It is possible to destroy a market in `Disputed` state using `admin_destroy_market`, but this will not result in the dispute bonds being released.

This is a bug, but it would be very unusual for a market to be deleted when it is already in `Disputed` state. So I'm putting low priority.

maltekliemann:
Fixes #686. Strictly speaking, we may have some markets on Battery Station which were destroyed in disputed state (haven't checked yet), so this would require a migration. But since this storage isn't really critical, we'll just leave it there. If it ever bothers us, we can remove it using governance.


maltekliemann:
Note that this doesn't require changes to the benchmarks, since we're already considering the amount of disputes.

zrml/prediction-markets/src/lib.rs
```diff
@@ -168,6 +168,18 @@ mod pallet {
...
             let (ids_len, disputes_len) = Self::clear_auto_resolve(&market_id)?;
+            // `Disputes` is emtpy unless the market is disputed, so this is just a defensive
+            // check.
+            if market.status == MarketStatus::Disputed {
+                for (index, dispute) in Disputes::<T>::take(market_id).iter().enumerate() {
+                    T::AssetManager::unreserve_named(
+                        &Self::reserve_id(),
+                        Asset::Ztg,
+                        &dispute.by,
+                        default_dispute_bond::<T>(index),
```
sea212, line 179: Do we already track storing the bonded amount (In case the dispute amount changes)?

sea212 (approved):
LGTM

zrml/prediction-markets/src/lib.rs
maltekliemann: No, that's somewhere in ZIP-0, I believe.

zrml/prediction-markets/src/tests.rs
```diff
@@ -427,6 +428,64 @@ fn admin_destroy_market_correctly_slashes_permissionless_market_disputed() {
...
+                account,
+                SENTINEL_AMOUNT,
+            ));
+        };
+        set_up_account(&CHARLIE);
+        set_up_account(&DAVE);
+
+        let balance_free_before_charlie = Balances::free_balance(&CHARLIE);
+        let balance_free_before_dave = Balances::free_balance(&DAVE);
+        assert_ok!(PredictionMarkets::admin_destroy_market(Origin::signed(SUDO), 0));
```
Chralt98, line 469: A check that `Disputes` at market id 0 is actually empty could be useful here. But not so important after ZIP-0.

Chralt98 (dismissed):

zrml/prediction-markets/src/tests.rs
maltekliemann: [f6bbfb3](https://github.com/zeitgeistpm/zeitgeist/pull/877/commits/f6bbfb3a3371d3b0774f01b5d49763d842433c34)

Chralt98:
Starting my review now..

Chralt98 (approved):
OK, although I am in favour of completely moving the dispute bond to `simple-disputes`, if not remove `simple-disputes` completely.

maltekliemann:
> OK, although I am in favour of moving the dispute bond to `simple-disputes`, if not remove `simple-disputes` completely.

Sure, but right now, we're still using the old system.

PR #878, opened 2022-11-21
Add with_fees boolean parameter for Swaps getSpotPrices()

### LINKED ISSUE #840: Rpc calls getSpotPrice(s) take swap fees into consideration by default.
The rpc calls to fetch spot prices take fees into consideration when calculating prices, while the indexer does not. This leads to some inconsistency in results when implementing a transparent call in the new sdk that can work with both indexer or rpc.

Could we add a optional flag to the calls to include fees or not? Making it an option so it can be omitted defeault being not including fees when calculating prices?
```with_fees: Option<bool>```

### PR DESCRIPTION
**Pull request by vivekvpandya**
Fixes #840

### CODE REVIEW

File: zrml/swaps/src/lib.rs [hunk 1]
```diff
@@ -1434,7 +1440,17 @@ mod pallet {
...
+                    T::RikiddoSigmoidFeeMarketEma::price(*pool_id, balance_out, &balances)?;
+                if with_fees {
+                    Ok(price_with_fee.saturated_into())
+                } else {
+                    let fee_pct = T::RikiddoSigmoidFeeMarketEma::fee(*pool_id)?.saturated_into();
+                    let fee_plus_one = BASE.saturating_add(fee_pct);
+                    let price_without_fee: u128 =
+                        bdiv(price_with_fee.saturated_into(), fee_plus_one)?;
+                    Ok(price_without_fee.saturated_into())
+                }
```
maltekliemann, line 1453: I think all these changes regarding Rikiddo should be reverted. It doesn't really matter what the Rikiddo behavior is right now, and it seems to me like this calculation is not correct.

@sea212 It seems like the term _fee_ is used as a name for `min(min_revenue, initial_fee, eta(r))`, where `eta` is the sigmoid function (ignoring the fact that `r` is not always well-defined), so if you multiply this with the total issuance of all outcome tokens, you get the exponent in the cost/price formula. It seems like it is used here as something like an approximation of the actual "fee" of Rikiddo. This seems off.

I'm fairly sure that one shouldn't even speak of a "fee" when dealing with LMSR and its variants. For example, in terms of LS-LMSR, if you buy from the market and then immediately sell back to the market, you make no profit or loss. This wouldn't be the case if there are fees. On low volume trades, the same should be true for Rikiddo. Have you or @numacoding ever defined an actual fee for Rikiddo in abstract terms?

zrml/swaps/src/tests.rs
```diff
@@ -689,25 +689,26 @@ fn most_operations_fail_if_pool_is_clean() {
...
+#[test_case(_3, _3, _100, _100, 0, 10_000_000_000, 10_000_000_000)]
+#[test_case(_3, _3, _100, _150, 0, 6_666_666_667, 6_666_666_667)]
+#[test_case(_3, _4, _100, _100, 0, 13_333_333_333, 13_333_333_333)]
+#[test_case(_3, _4, _100, _150, 0, 8_888_888_889, 8_888_888_889)]
+#[test_case(_3, _6, _125, _150, 0, 16_666_666_667, 16_666_666_667)]
+#[test_case(_3, _6, _125, _100, 0, 25_000_000_000, 25_000_000_000)]
+#[test_case(_3, _3, _100, _100, _1_10, 11_111_111_111, 10_000_000_000)]
+#[test_case(_3, _3, _100, _150, _1_10, 7_407_407_408, 66_666_666_67)]
+#[test_case(_3, _4, _100, _100, _1_10, 14_814_814_814, 13_333_333_333)]
+#[test_case(_3, _4, _100, _150, _1_10, 9_876_543_210, 88_888_888_89)]
```
maltekliemann, line 701:
```suggestion
#[test_case(_3, _4, _100, _150, _1_10, 9_876_543_210, 8_888_888_889)]
```

zrml/swaps/src/tests.rs
```diff
@@ -689,25 +689,26 @@ fn most_operations_fail_if_pool_is_clean() {
...
-#[test_case(_3, _6, _125, _150, _1_10, 18_518_518_519)]
-#[test_case(_3, _6, _125, _100, _1_10, 27_777_777_778)]
+#[test_case(_3, _3, _100, _100, 0, 10_000_000_000, 10_000_000_000)]
+#[test_case(_3, _3, _100, _150, 0, 6_666_666_667, 6_666_666_667)]
+#[test_case(_3, _4, _100, _100, 0, 13_333_333_333, 13_333_333_333)]
+#[test_case(_3, _4, _100, _150, 0, 8_888_888_889, 8_888_888_889)]
+#[test_case(_3, _6, _125, _150, 0, 16_666_666_667, 16_666_666_667)]
+#[test_case(_3, _6, _125, _100, 0, 25_000_000_000, 25_000_000_000)]
+#[test_case(_3, _3, _100, _100, _1_10, 11_111_111_111, 10_000_000_000)]
+#[test_case(_3, _3, _100, _150, _1_10, 7_407_407_408, 66_666_666_67)]
```
maltekliemann, line 699:
```suggestion
#[test_case(_3, _3, _100, _150, _1_10, 7_407_407_408, 6_666_666_667)]
```

maltekliemann (changes requested):
You should also add a note to the changelog.

zrml/swaps/src/lib.rs
sea212: The first usage of the term "fee" was defined in the [Rikiddo research report from Numa](https://docs.google.com/document/d/1GF29lC2t987Knk57RZPRLOjNTdI2G-q9gs3_nufRCss/edit).

It is actually `min(min_revenue, initial_fee + eta(r))`. `min_revenue` is what Numa calls "minimal revenue coefficient (omega)".

It is some kind of approximation. This part definitely has to be improved, but for now we can leave it like that and raise an issue to handle that imo. (since Rikiddo is not active).
I think in a path-independent algorithm like LS-LMSR the assumption that there is no profit or loss is correct, because the amount of base currency to pay for buying `C(q0) -> C(q1)` is `C(q1) - C(q0)` and the amount received for selling `C(q1) -> C(q0)` is `C(q1) - C(q0)`. However, with Rikiddo this does not hold, because the "fee" parameter can vary in between two trades and change `C(q)`. It depends on additional data, I think it would be cleaner if this would be indicated in the function, like `C(q, r)`.
Although the question was raised what kind of implications the loss of the path-independence property has, it was never properly examined.

maltekliemann: Thanks for the answer. Can you provide access to the research report. I seem to be locked out. :(

Indeed, we should start writing `C(q, r)` or `C_r(q)` (note: `C_r`, with fixed `r` is actually path-independent). As you say, for path-independent buying and then selling results in an exact break-even (because of, well, path-independence). Which means that there are no fees involved.

My issue here is that if the buy _and_ the sell price go up, then whenever you sell to the market, the market creators pay _you_ a fee. That just feels wrong. I think that 3.3.1 of _A Practical Liquidity-Sensitive Automated Market Maker_ provides a somewhat servicable solution to this problem.

Chralt98: @vivekvpandya so maybe a TODO with an issue number would be appropriate.

zrml/swaps/src/lib.rs
```diff
@@ -1444,9 +1460,13 @@ mod pallet {
                 .saturated_into();
                 let fee_pct = T::RikiddoSigmoidFeeMarketEma::fee(*pool_id)?.saturated_into();
                 let fee_plus_one = BASE.saturating_add(fee_pct);
-                let price_with_fee: u128 =
-                    bmul(fee_plus_one, bmul(price_with_inverse_fee, fee_plus_one)?)?;
-                Ok(price_with_fee.saturated_into())
+                let price_without_fee: u128 = bmul(price_with_inverse_fee, fee_plus_one)?;
```
Chralt98: Ok, I understand now. Because inside the Rikiddo price we already have fees included, we need to multiply by `fee_plus_one` to get the fees out of the term.

```text
price_with_inverse_fee = 1 ZTG / (price + (1 + fee))
price_without_fee = (1 ZTG * (1 + fee)) / (price + (1 + fee)) = 1 ZTG / price
```

Chralt98 (commented):
Aren't the RPC calls defined here? https://github.com/zeitgeistpm/zeitgeist/blob/50339918748e597c6e13018dc97c4ccd4d001ef2/zrml/swaps/rpc/src/lib.rs#L40-L201

vivekvpandya:
@Chralt98 https://github.com/zeitgeistpm/zeitgeist/pull/878/files#diff-188fb32939fe9cc9416edd214b3ab316def04705fad5aa33b9130b8d49220244R1415 TODO added here.

Chralt98:
> Aren't the RPC calls defined here?

Completely my fault. I overlooked your changes made to this file.

zrml/swaps/src/lib.rs
```diff
@@ -1455,10 +1462,14 @@ mod pallet {
...
-                let price_with_fee: u128 = bmul(fee_plus_one, price_without_fee)?;
-                Ok(price_with_fee.saturated_into())
+                if with_fees {
+                    let fee_pct = T::RikiddoSigmoidFeeMarketEma::fee(*pool_id)?.saturated_into();
+                    let fee_plus_one = BASE.saturating_add(fee_pct);
+                    let price_with_fee: u128 = bmul(fee_plus_one, price_without_fee)?;
+                    Ok(price_with_fee.saturated_into())
+                } else {
+                    Ok(price_without_fee.saturated_into())
+                }
```
Chralt98, line 1472: May I ask you, why you still regard with_fees here for Rikiddo?

vivekvpandya: Just missed that, removed on local machine, I will push after sometime along with formatting issue fix.

zrml/swaps/runtime-api/src/lib.rs
```diff
@@ -31,6 +31,11 @@ sp_api::decl_runtime_apis! {
     {
         fn pool_shares_id(pool_id: PoolId) -> Asset<SerdeWrapper<MarketId>>;
         fn pool_account_id(pool_id: &PoolId) -> AccountId;
-        fn get_spot_price(pool_id: &PoolId, asset_in: &Asset<MarketId>, asset_out: &Asset<MarketId>) -> SerdeWrapper<Balance>;
+        fn get_spot_price(
+            pool_id: &PoolId,
+            asset_in: &Asset<MarketId>,
+            asset_out: &Asset<MarketId>,
+            with_fees: bool
```
maltekliemann, line 38:
```suggestion
            with_fees: bool,
```

maltekliemann (dismissed):

Chralt98 (dismissed):

Chralt98 (approved):

maltekliemann (approved):

PR #886, opened 2022-11-25
Update weights

sea212:


scripts/benchmarks/configuration.sh
```diff
@@ -43,4 +43,4 @@ else
 fi
 export EXECUTION="${EXECUTION:-wasm}"
 export ADDITIONAL="${ADDITIONAL:-}"
-export ADDITIONAL_FEATURES="${ADDITIONAL_FEATURES:-}"
\ No newline at end of file
+export ADDITIONAL_FEATURES="${ADDITIONAL_FEATURES:-with-global-disputes}"
```
Chralt98, line 46: Wow, nice, that you caught that too! So this release build needs `with-global-disputes` feature for battery station.

```suggestion
# TODO(#848) Delete default `with-global-disputes`
export ADDITIONAL_FEATURES="${ADDITIONAL_FEATURES:-with-global-disputes}"
```

zrml/swaps/src/weights.rs
```diff
@@ -80,45 +80,51 @@ impl<T: frame_system::Config> WeightInfoZeitgeist for WeightInfo<T> {
...
+            .saturating_add(T::DbWeight::get().writes((67 as Weight).saturating_mul(a as Weight)))
+    }
+    // Storage: Swaps PoolsCachedForArbitrage (r:7 w:6)
+    fn apply_to_cached_pools_noop(a: u32) -> Weight {
+        (0 as Weight)
+            // Standard Error: 7_000
+            .saturating_add((10_520_000 as Weight).saturating_mul(a as Weight))
+            .saturating_add(T::DbWeight::get().reads(1 as Weight))
+            .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(a as Weight)))
+            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(a as Weight)))
```
Chralt98, line 118: Hopefully the arbitrage doesn't stall our chain in the end. 🤞 I know the weight problem is known and I already had a couple of comments in the review about this, but just that we keep this in our minds if the chain stalls for unknown reasons.

Chralt98 (approved):
OK

maltekliemann (approved):

scripts/benchmarks/configuration.sh
sea212: I'll leave that out to be able to wrap up the release faster, since the comment is not mandatory

PR #887, opened 2022-11-26
Upgrade 0929

### LINKED ISSUE #869: Upgrade codebase to polkadot-v0.9.29
**Dependencies**
Cumulus: https://github.com/paritytech/cumulus/tree/polkadot-v0.9.29
Moonbeam: https://github.com/PureStake/moonbeam/tree/v0.27.2
Nimbus: https://github.com/PureStake/nimbus/tree/moonbeam-polkadot-v0.9.29
ORML: https://github.com/open-web3-stack/open-runtime-module-library/tree/polkadot-v0.9.29
Polkadot: https://github.com/paritytech/polkadot/tree/release-v0.9.29
Substrate: https://github.com/paritytech/substrate/tree/polkadot-v0.9.29

**Changes**
Cumulus: https://github.com/paritytech/cumulus/compare/polkadot-v0.9.26...polkadot-v0.9.29
Moonbeam: https://github.com/PureStake/moonbeam/compare/v0.26.1...v0.27.2
Nimbus: https://github.com/PureStake/nimbus/compare/moonbeam-polkadot-v0.9.26...moonbeam-polkadot-v0.9.29
ORML: https://github.com/open-web3-stack/open-runtime-module-library/compare/polkadot-v0.9.26...polkadot-v0.9.29
Polkadot: https://github.com/paritytech/polkadot/compare/release-v0.9.26...release-v0.9.29
Substrate: https://github.com/paritytech/substrate/compare/polkadot-v0.9.26...polkadot-v0.9.29
Substrate-Parachain: https://github.com/substrate-developer-hub/substrate-parachain-template/compare/polkadot-v0.9.26...polkadot-v0.9.29

vivekvpandya:
Closes #869


Chralt98:
Starting review now...

misc/frame_weight_template.hbs
```diff
@@ -44,22 +45,22 @@ impl<T: frame_system::Config> {{pallet}}::weights::WeightInfo for WeightInfo<T>
         {{~#each benchmark.components as |c| ~}}
         {{~#if (not c.is_used)}}_{{/if}}{{c.name}}: u32, {{/each~}}
     ) -> Weight {
-        ({{underscore benchmark.base_weight}} as Weight)
+        Weight::from_ref_time({{underscore benchmark.base_weight}} as u64)
```
Chralt98, line 48:
```suggestion
        Weight::from_ref_time({{underscore benchmark.base_weight}})
```

runtime/zeitgeist/src/parameters.rs
```diff
@@ -290,6 +292,8 @@ parameter_types! {
     /// Minimum amount of the multiplier. The test `multiplier_can_grow_from_zero` ensures
     /// that this value is not too low.
     pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000_000u128);
+    /// Maximum amount of the multiplier.
+    pub MaximumMultiplier: Multiplier = Bounded::max_value();
```
Chralt98, line 296: Can you send me a reference, where you found this configuration?

runtime/zeitgeist/src/parachain_params.rs
```diff
@@ -48,11 +48,12 @@ parameter_types! {
     // Cumulus and Polkadot
     pub Ancestry: MultiLocation = Parachain(ParachainInfo::parachain_id().into()).into();
     pub const RelayLocation: MultiLocation = MultiLocation::parent();
-    pub const RelayNetwork: NetworkId = NetworkId::Kusama;
-    pub const ReservedDmpWeight: Weight = MAXIMUM_BLOCK_WEIGHT / 4;
-    pub const ReservedXcmpWeight: Weight = MAXIMUM_BLOCK_WEIGHT / 4;
+    // Have to change "Any" to "Kusama" for mainnet once we have separate runtimes
+    pub const RelayNetwork: NetworkId = NetworkId::Any;
```
Chralt98, line 52: I don't get this. Can you explain why?

runtime/zeitgeist/src/parachain_params.rs
```diff
@@ -48,11 +48,12 @@ parameter_types! {
...
+    pub const ReservedDmpWeight: Weight = Weight::from_ref_time(MAXIMUM_BLOCK_WEIGHT.ref_time() / 4);
+    pub const ReservedXcmpWeight: Weight = Weight::from_ref_time(MAXIMUM_BLOCK_WEIGHT.ref_time() / 4);
```
Chralt98, line 54: Would prefer this way.

https://github.com/PureStake/moonbeam/blob/2c7c152cc805326014f6989f8502f66952d33731/runtime/moonbase/src/lib.rs#L619-L620

But your choice.

node/src/service/service_parachain.rs
```diff
@@ -261,10 +261,6 @@ where
         bool,
     ) -> Result<Box<dyn ParachainConsensus<Block>>, sc_service::Error>,
 {
-    if matches!(parachain_config.role, Role::Light) {
-        return Err("Light client not supported!".into());
-    }
-
```
Chralt98, line 267: Seems like a mistake here. Or is this by purpose?

zrml/swaps/src/tests.rs
```diff
@@ -3121,7 +3121,7 @@ fn on_idle_skips_arbitrage_if_price_does_not_exceed_threshold() {
         let pool_id = 0;
         // Force the pool into the cache.
         crate::PoolsCachedForArbitrage::<Runtime>::insert(pool_id, ());
-        Swaps::on_idle(System::block_number(), Weight::max_value());
+        Swaps::on_idle(System::block_number(), Weight::from_ref_time(u64::max_value()));
```
Chralt98, line 3124:
```suggestion
        Swaps::on_idle(System::block_number(), Weight::MAX);
```

https://paritytech.github.io/substrate/master/src/sp_weights/weight_v2.rs.html#77

zrml/swaps/src/tests.rs
```diff
@@ -3156,7 +3156,7 @@ fn on_idle_arbitrages_pools_with_mint_sell() {
 
         // Force arbitrage hook.
         crate::PoolsCachedForArbitrage::<Runtime>::insert(pool_id, ());
-        Swaps::on_idle(System::block_number(), Weight::max_value());
+        Swaps::on_idle(System::block_number(), Weight::from_ref_time(u64::max_value()));
```
Chralt98, line 3159:
```suggestion
        Swaps::on_idle(System::block_number(), Weight::MAX);
```

zrml/swaps/src/tests.rs
```diff
@@ -3241,8 +3241,8 @@ fn apply_to_cached_pools_only_drains_requested_pools() {
         let number_of_pools_to_retain: u32 = 3;
         Swaps::apply_to_cached_pools(
             pool_count.saturated_into::<u32>() - number_of_pools_to_retain,
-            |_| Ok(0),
-            Weight::max_value(),
+            |_| Ok(Weight::from_ref_time(0)),
+            Weight::from_ref_time(u64::max_value()),
```
Chralt98, line 3245:
```suggestion
            Weight::MAX,
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -2583,7 +2583,7 @@ mod pallet {
                 true
             };
 
-            let mut weight_basis = 0;
+            let mut weight_basis = Weight::from_ref_time(0);
```
Chralt98, line 2586:
```suggestion
            let mut weight_basis = Weight::zero();
```

Chralt98 (changes requested):
Back to the future of the Official Zeitgeist Forks. I cut this review here.

node/src/service/service_parachain.rs
vivekvpandya: AFAIR Role::Light was removed.

runtime/zeitgeist/src/parameters.rs
vivekvpandya: Based on https://github.com/paritytech/substrate/blob/master/frame/transaction-payment/src/lib.rs#L136

runtime/zeitgeist/src/parachain_params.rs
vivekvpandya: I think this is mistake. I don't remember now why I added this and from where I took inspiration. :sweat_smile: 
I will revert this

node/src/chain_spec/mod.rs
```diff
@@ -167,7 +170,7 @@ macro_rules! generate_generic_genesis_function {
                 system: $runtime::SystemConfig { code: wasm_binary.to_vec() },
                 technical_committee: Default::default(),
                 technical_committee_membership: $runtime::TechnicalCommitteeMembershipConfig {
-                    members: vec![].try_into().unwrap(),
+                    members: frame_support::bounded_vec![],
```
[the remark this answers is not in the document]
vivekvpandya: I don't recall now, but may be build with some features was not happy. I will revert this and frame_support include.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1026,7 +1026,7 @@ mod pallet {
                 return Ok(Some(T::WeightInfo::redeem_shares_scalar()).into());
             }
 
-            Ok(None.into())
+            Ok((Option::<Weight>::None, Pays::No).into())
```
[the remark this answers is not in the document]
vivekvpandya: Yes what you think is correct for syntax.

zrml/swaps/src/lib.rs
```diff
@@ -1249,19 +1249,18 @@ mod pallet {
...
-                })
-                .saturated_into(); // Saturates only if we have u32::MAX trades per block.
-            if pool_count == 0 {
+                });
+            if pool_count == Weight::zero() {
                 return weight;
             }
             Self::apply_to_cached_pools(
-                pool_count,
+                pool_count.ref_time().saturated_into(),
```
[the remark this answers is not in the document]
vivekvpandya: Could you please explain reason/benefit for your preference?

node/Cargo.toml
```diff
@@ -3,103 +3,102 @@ name = "zeitgeist"
...
 
 # Utility
 cfg-if = { version = "1.0.0" }
 clap = { version = "3.2.6", features = ["derive"] }
 hex-literal = { version = "0.3.4" }
-jsonrpsee = { version = "0.14.0", features = ["server"] }
+jsonrpsee = { version = "0.15.1", features = ["server"] }
 log = { optional = true, version = "0.4.17" }
-# TODO(#865): Remove in future Polkadot release
-tracing-core = "=0.1.26"
```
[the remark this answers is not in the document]
vivekvpandya: In this branch there are no migrations yet, it will be there once I merge main into this branch. But it does print lots of INFO and DEBUG traces so I think we are good now even without above explicit pinning of the dependency.

zrml/swaps/src/lib.rs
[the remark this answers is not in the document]
Chralt98: My suggestion was based on [this](https://github.com/zeitgeistpm/zeitgeist/blob/fbdf962d88fdfc8c30bb048d949ce00d5701d816/zrml/swaps/src/lib.rs#L1253-L1254). `overhead` is of type `Weight` and `extra_weight_per_pool.ref_time()` is of type `u64`.

runtime/common/src/lib.rs
```diff
@@ -1477,8 +1466,16 @@ macro_rules! create_runtime_api {
                     (weight, RuntimeBlockWeights::get().max_block)
                 }
```
sea212, line 1468: The implementation of `pallet_transaction_payment_rpc_runtime_api::TransactionPaymentCallApi` is missing. This was added between v0.9.26 and v0.9.29 and is a useful runtime call, since it allows us to query fee data using calls instead of extrinsics, as shown in the following example:
image
image

Here is the patch:
[implement_TransactionPaymentCallApi_v0929.zip](https://github.com/zeitgeistpm/zeitgeist/files/10353760/implement_TransactionPaymentCallApi_v0929.zip)

sea212 (changes requested):

node/Cargo.toml
```diff
@@ -3,103 +3,101 @@ name = "zeitgeist"
...
-nimbus-primitives = { branch = "moonbeam-polkadot-v0.9.26", default-features = false, git = "https://github.com/purestake/nimbus", optional = true }
-pallet-author-inherent = { branch = "moonbeam-polkadot-v0.9.26", default-features = false, git = "https://github.com/purestake/nimbus", optional = true }
-pallet-parachain-staking = { tag = "v0.26.1", git = "https://github.com/purestake/moonbeam", optional = true }
+moonbeam-vrf = { tag = "v0.27.2-a", git = "https://github.com/zeitgeistpm/moonbeam", optional = true }
+nimbus-consensus = { branch = "moonbeam-polkadot-v0.9.29", default-features = false, git = "https://github.com/zeitgeistpm/nimbus", optional = true }
+nimbus-primitives = { branch = "moonbeam-polkadot-v0.9.29", default-features = false, git = "https://github.com/zeitgeistpm/nimbus", optional = true }
+pallet-author-inherent = { branch = "moonbeam-polkadot-v0.9.29", default-features = false, git = "https://github.com/zeitgeistpm/nimbus", optional = true }
+pallet-parachain-staking = { tag = "v0.27.2-a", git = "https://github.com/zeitgeistpm/moonbeam", optional = true }
```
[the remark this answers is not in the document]
vivekvpandya: So once we patch substrate forked from their repo and use it only on ztg but not use it in other deps then there are two versions of substrate built every-time. I am not sure of patching can help here.

runtime/common/src/lib.rs
```diff
@@ -841,7 +838,7 @@ macro_rules! impl_config_traits {
             #[cfg(feature = "parachain")]
             type OnTimestampSet = ();
             #[cfg(not(feature = "parachain"))]
-            type OnTimestampSet = Aura;
+            type OnTimestampSet = ();
```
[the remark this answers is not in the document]
vivekvpandya: Thanks for noticing this, I think I need to investigate more here, I was facing https://github.com/paritytech/subport/issues/94
with ./scripts/benchmarks/quick_check.sh and hence did what was suggested there but your point is also correct.

node/src/command.rs
```diff
@@ -167,18 +167,23 @@ pub fn run() -> sc_cli::Result<()> {
...
                             )
                         }),
                         #[cfg(not(feature = "with-battery-station-runtime"))]
                         _ => panic!("{}", crate::BATTERY_STATION_RUNTIME_NOT_AVAILABLE),
                     }
                 }
+
+                BenchmarkCmd::Extrinsic(_cmd) => {
+                    Err("BenchmarkCmd Extrinsic is not yet supported".into())
+                }
```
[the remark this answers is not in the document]
vivekvpandya: oh sorry, I was using moonbeam and mangata-node as reference and in their implementation they have skipped this. So I did not dig more into this.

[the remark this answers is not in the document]
sea212: Please add this as it was part of the framework update to polkadot-v0.9.29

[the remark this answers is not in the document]
vivekvpandya: Already added https://github.com/zeitgeistpm/zeitgeist/pull/887/files#diff-8ce03a5a2b79f3021ff2d6322d13448db8cfd5fb9267517ab4b96d0d299ed45fR190 few days ago.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -682,14 +692,26 @@ benchmarks! {
...
             ScoringRule::CPMM,
-            Some(MarketPeriod::Timestamp(range_start..range_end)),
+            Some(period),
         )?;
+        let market = <zrml_market_commons::Pallet<T>>::market(&market_id)?;
+        let (range_start, range_end) = match market.period {
+            MarketPeriod::Timestamp(range) => (range.start, range.end),
+                _ => {
+                    panic!("MarketPeriod is block_number based");
+                }
```
Chralt98, line 713:
```suggestion
            _ => {
                panic!("MarketPeriod is block_number based");
            }
```

Chralt98 (commented):
Wow, what a lot of changes here for this upgrade PR. It's always hard to have a good overview of all together. I agree with Harald, that the pm-benchmark `pallet-aura` is not handled ideally, but because of a lack of a better approach, I would be okay with it.

scripts/benchmarks/configuration.sh
```diff
@@ -15,7 +15,7 @@ export FRAME_WEIGHT_TEMPLATE="./misc/frame_weight_template.hbs"
 # pallet_crowdloan_rewards benchmark lead to an error within the verify function (deprecated)
 export FRAME_PALLETS_PARACHAIN=( 
                 cumulus_pallet_xcmp_queue \
-                # pallet_author_inherent \
+                pallet_author_inherent \
```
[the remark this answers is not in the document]
vivekvpandya: Perhaps it was failing in previous releases? Need to check commit which commented out it. But with this upgrade Harald noticed that we can re-enable it.

node/Cargo.toml
```diff
@@ -3,103 +3,101 @@ name = "zeitgeist"
...
 parity-scale-codec = { optional = true, version = "3.0.0" }
-sc-chain-spec = { branch = "polkadot-v0.9.26", git = "https://github.com/paritytech/substrate", optional = true }
-sc-network = { branch = "polkadot-v0.9.26", git = "https://github.com/paritytech/substrate", optional = true }
-sc-tracing = { branch = "polkadot-v0.9.26", git = "https://github.com/paritytech/substrate", optional = true }
+sc-chain-spec = { branch = "moonbeam-polkadot-v0.9.29", git = "https://github.com/zeitgeistpm/substrate", optional = true }
+sc-network = { branch = "moonbeam-polkadot-v0.9.29", git = "https://github.com/zeitgeistpm/substrate", optional = true }
```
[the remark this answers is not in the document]
vivekvpandya: This seems interesting , I could not find much details so asked here https://github.com/AcalaNetwork/Acala/pull/2334#pullrequestreview-1261095286

node/src/command.rs
```diff
@@ -167,12 +173,72 @@ pub fn run() -> sc_cli::Result<()> {
...
+                                    Sr25519Keyring::Alice.to_account_id(),
+                                    BatteryStationED::get(),
+                                    false,
+                                )),
+                            ]);
+                            cmd.run(
+                                params.client,
+                                inherent_benchmark_data()?,
+                                Vec::new(),
+                                &ext_factory,
```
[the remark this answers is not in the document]
vivekvpandya: It is not a big effort, so we just added it. May not be useful right now but does not hurt to have it.

node/src/benchmarking.rs
```diff
@@ -25,18 +25,18 @@ use sp_inherents::{InherentData, InherentDataProvider};
 use sp_keyring::Sr25519Keyring;
```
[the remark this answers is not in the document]
vivekvpandya: IIUC https://github.com/paritytech/substrate/pull/11456/files#diff-2ed17e03389cef5a97d8a16fd718426df10ac9d21a705e3cf8ad7228a3ef6e81
all code in that file is for benchmark only.

node/Cargo.toml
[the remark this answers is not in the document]
vivekvpandya: okay based on reply from above comment, it seems like in our case those imports are found from sc_network crate. But we can also change.

utils/src/lib.rs
```diff
@@ -35,30 +35,7 @@ where
     let slot = timestamp / pallet_aura::Pallet::<T>::slot_duration();
     let digest = Digest { logs: vec![DigestItem::PreRuntime(AURA_ENGINE_ID, slot.encode())] };
     frame_system::Pallet::<T>::initialize(&block_number, &Default::default(), &digest);
+    // NOTE: pallet aura only needed with the `parachain` feature
```
Chralt98, line 38: I thought pallet aura is for standalone mode? 

> pallet aura only needed with the `parachain` feature

This indicates, that pallet aura runs with the parachain feature, which is not true, right?

node/Cargo.toml
```diff
@@ -3,103 +3,102 @@ name = "zeitgeist"
...
 # Utility
 cfg-if = { version = "1.0.0" }
-clap = { version = "3.2.6", features = ["derive"] }
+clap = { version = "3.2.20", features = ["derive"] }
 hex-literal = { version = "0.3.4" }
-jsonrpsee = { version = "0.14.0", features = ["server"] }
+jsonrpsee = { version = "0.15.1", features = ["server"] }
 log = { optional = true, version = "0.4.17" }
-# TODO(#865): Remove in future Polkadot release
-tracing-core = "=0.1.26"
```
Chralt98, line 102: For what moonbeam uses tracing-core = "0.1.29" https://github.com/PureStake/moonbeam/compare/v0.26.0...v0.27.2#diff-5dc0537169cb810865347de9185dae9ac7d231b7b5eb18eae2086c578dda133eR27 ?

zrml/prediction-markets/src/mock.rs
```diff
@@ -90,6 +90,7 @@ construct_runtime!(
         UncheckedExtrinsic = UncheckedExtrinsicTest<Runtime>,
     {
         Authorized: zrml_authorized::{Event<T>, Pallet, Storage},
+        Aura: pallet_aura::{Pallet, Storage, Config<T>},
```
Chralt98, line 93: > error: `Aura` does not have the std feature enabled, this will cause the `pallet_aura::GenesisConfig` type to be undefined.

node/Cargo.toml
vivekvpandya: I am not sure, we had used explicit version of tracing-core because traces in try-runtime was not displayed properly.
but now with this upgrade we no longer need this and it works fine.

utils/src/lib.rs
vivekvpandya: Thanks for correcting!

Chralt98 (commented):
`scripts/benchmarks/run_benchmarks` still has `parachain` feature [here](https://github.com/zeitgeistpm/zeitgeist/blob/5a7feaf7c83c10d2fe63288413e3c336aadda68c/scripts/benchmarks/run_benchmarks.sh#L37) When you want to have the benchmarks for prediction markets, you need to handle the case in this script too.

Exactly like for the `quick_benchmark.rs`

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -696,8 +696,8 @@ benchmarks! {
         <frame_system::Pallet<T>>::set_block_number(2_u64.saturated_into());
         let block = frame_system::Pallet::<T>::block_number();
         zeitgeist_utils::set_block_number_timestamp::<T>(block, start.saturated_into());
-        let range_start: MomentOf<T> = ((2_u64 + 100_000_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
-        let range_end: MomentOf<T> = ((2_u64 + 100_000_000_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
+        let range_start: MomentOf<T> = ((2_u64 + 100_00_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
+        let range_end: MomentOf<T> = ((2_u64 + 100_000_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
```
Chralt98, line 700: range_start should be smaller than range_end

Chralt98: You have a max period of `365 * 24  * 60 * 60 * (60000 / 12000) = 157_680_000`.

vivekvpandya:
> `scripts/benchmarks/run_benchmarks` still has `parachain` feature [here](https://github.com/zeitgeistpm/zeitgeist/blob/5a7feaf7c83c10d2fe63288413e3c336aadda68c/scripts/benchmarks/run_benchmarks.sh#L37) When you want to have the benchmarks for prediction markets, you need to handle the case in this script too.
> 
> Exactly like for the `quick_benchmark.rs`

This is not a problem right as it only runs FRAME_PALLETS with parachain feature and odes not run zrm_ pallets.

Chralt98:
> This is not a problem right as it only runs FRAME_PALLETS with parachain feature and odes not run zrm_ pallets.

Correct, my bad!

node/src/command.rs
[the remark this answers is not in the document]
sea212: Those are incremental changes that are slowly introduced by Parity. This way we avoid having to note all the changes that we don't integrate immediately and having to track them to eventually integrate them in a big chunk.

Chralt98:
Did you look into [this](https://github.com/zeitgeistpm/zeitgeist/pull/887#discussion_r1072825071) ? I know you might be a little annoyed. I understand that.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -761,7 +761,7 @@ benchmarks! {
         <frame_system::Pallet<T>>::set_block_number(2_u64.saturated_into());
         let block = frame_system::Pallet::<T>::block_number();
         let range_start: MomentOf<T> = (2_u64 * MILLISECS_PER_BLOCK as u64).saturated_into();
-        let range_end: MomentOf<T> = ((2_u64 + 100_00_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
+        let range_end: MomentOf<T> = ((2_u64 + 10_000_u64) * MILLISECS_PER_BLOCK as u64).saturated_into();
```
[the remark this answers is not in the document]
Chralt98: nitpicky, but it could be a deeper reason there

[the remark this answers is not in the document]
vivekvpandya: clippy did not complained about it.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -169,18 +169,22 @@ fn setup_redeem_shares_common<T: Config + pallet_timestamp::Config>(
...
         (market.deadlines.grace_period.saturated_into::<u32>() + 1) * MILLISECS_PER_BLOCK;
-    pallet_timestamp::Pallet::<T>::set_timestamp((end + grace_period).into());
+    let block = frame_system::Pallet::<T>::block_number();
+    zeitgeist_utils::set_block_number_timestamp::<T>(block, (end + grace_period).into());
     Call::<T>::report { market_id, outcome }
         .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
     Call::<T>::admin_move_market_to_resolved { market_id }
         .dispatch_bypass_filter(resolve_origin)?;
     Ok((caller, market_id))
 }
```
sea212, line 180: Thanks for applying the patch at https://github.com/zeitgeistpm/zeitgeist/pull/887#discussion_r1072825071 . I see now what was missing in the patch to make it complete. Some arguments against the approach to tightly couple Aura:

- The problem actually lies in the minimum timestamp period. It is too low in the benchmark environment, which hinders us from "skipping" blocks in terms of a timestamp delta. This is does not only affect Aura, it could also deny proper execution of other consensus systems.
- Currently we enforce that projects set the `parachain` feature when using PM in case they don't use Aura.
- Projects that have a standalone chain but don't use Aura (they could use Babe instead for example), won't be able to use the pallet benchmarks.

The solution suggested here does two things:
1. Create a custom `MinimumPeriod` implementation.
2. Within the `MinimumPeriod` implementation, figure out whether the node is in a benchmark environment or not and return an adequate value.

Unfortunately I am not aware of a direct way to query whether the node is in a benchmark environment, however after some extensive analysis of the pallet there is an indirect way. The [account whitelist can be fetched](https://github.com/paritytech/substrate/blob/d4384f242f48fd62516aac02c3f18b5a8c9e5793/frame/benchmarking/src/utils.rs#L281-L284), and this is something that [should only be set in a benchmark environment](https://github.com/paritytech/substrate/blob/c8653447fc8ef8d95a92fe164c96dffb37919e85/primitives/externalities/src/lib.rs#L289-L294). During initialization of the benchmarks, a couple of keys are added to the whitelist. We can use this as a measurement to figure out whether the node is in a benchmark environment, by checking whether the whitelist contains elements.

The solution is to define the `MinimumPeriod` of the timestamp pallet as such:
```rust
// Timestamp
/// Custom getter for minimum timestamp delta.
/// This ensures that consensus systems like Aura don't experience breaking assertions
/// in a benchmark environment
pub struct MinimumPeriod;
impl MinimumPeriod {
    /// Returns the value of this parameter type.
    pub fn get() -> u64 {
        #[cfg(feature = "runtime-benchmarks")]
        {
            use frame_benchmarking::{Benchmarking, benchmarking::get_whitelist};
            // Should that condition be true, we can assume that we are in a benchmark environment.
            if !get_whitelist().is_empty() { return u64::MAX; }
        }


        MILLISECS_PER_BLOCK as u64 / 2
    }
}
impl<I: From<u64>> frame_support::traits::Get<I> for MinimumPeriod {
    fn get() -> I {
        I::from(Self::get())
    }
}
impl frame_support::traits::TypedGet for MinimumPeriod {
    type Type = u64;
    fn get() -> u64 {
        Self::get()
    }
}
```

You can add https://github.com/sea212/zeitgeist as a remote and then merge the branch `sea212-fix-aura-slot-issue` into this PR. [Click here to see a changelog](https://github.com/zeitgeistpm/zeitgeist/compare/zeitgeistpm:zeitgeist:upgrade_0929...sea212:zeitgeist:sea212-fix-aura-slot-issue).

sea212: I am not sure if the changes (besides the Aura thing) within the PM benchmarks were mandatory, so feel free to revert some of the changes from my PR in regards to that.

sea212 (changes requested):
The PR is ready for merge, but the aura solution is still a needle in my approval finger.
This time I have found the root of the cause and provided a proper solution to fix it in the comment.

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: It feels like with this new solution it will always have CurrentSlot as zero. and that's how it prevents assertion failures.

vivekvpandya: I had initially though for making something like that but when I took look at other projects their approach seems better to me. :)

sea212 (approved):

sea212 (approved):
LGTM, great work Vivek :+1:

Chralt98 (approved):

PR #891, opened 2022-11-28
Add market_commons as tightly coupled dependency prediction_markets and swaps

### LINKED ISSUE #805: [PM][Swaps] Tightly couple market-commons pallet
The market-commons pallet was spawned in the past to avoid circular dependencies between the prediction markets and swaps pallet. Both pallets strictly depend on it, and when executing migrations, often they need access to its storage. This is a sign that market-commons should be tightly coupled to those pallets.

vivekvpandya:
Closes #805


vivekvpandya:
Closing this as I messed up with branch locally.

PR #894, opened 2022-12-01
Add base_asset to Market structure and make prediction_markets pallet use it instead of Asset::Ztg

### LINKED ISSUE #226: [PM]`PredictionMarkets` can use any approved token as a base currency
Currently, the `PredictionMarkets` pallet expects that only the native currency (ZTG) is being used as the base currency for all prediction markets to settle in. However, stablecoin like kUSD will also be a popular choice when creating markets, and should be able to be used as a base currency as well.

- There should be a list of approved tokens that can be used as base currency for prediction markets. The list of approved tokens can be changed by governance or the root origin. 
- We should add `USD` as a token type in the [`Asset`](https://github.com/zeitgeistpm/zeitgeist/blob/main/primitives/src/asset.rs#L22) enum, which currently is just stubbed.
- `PredictionMarkets` internal logic should be changed from assuming that all markets are using `ZTG` as the base currency, to using any of the approved assets.

vivekvpandya:
Fixes #226


zrml/prediction-markets/src/orml_asset_registry.rs
```diff
@@ -0,0 +1,268 @@
+// Copyright 2021 Centrifuge Foundation (centrifuge.io).
```
vivekvpandya, line 1: This code is almost taken as it is from Centrifuge repo. AFAIU we need to keep their license note right?

zrml/prediction-markets/src/tests.rs
```diff
@@ -340,6 +351,47 @@ fn create_market_fails_on_max_oracle_duration() {
...
+                deadlines,
+                gen_metadata(2),
+                MarketCreation::Permissionless,
+                MarketType::Categorical(2),
+                MarketDisputeMechanism::SimpleDisputes,
+                ScoringRule::CPMM,
+            ),
+            crate::Error::<Runtime>::InvalidBaseAsset,
+        );
+        // As per Mock asset_registry genesis ForeignAsset(100) has allow_in_pool set to true.
```
Chralt98, line 379: So `allow_in_pool` is meant to be a guarantee, that we allow the `true` assets as base asset in prediction markets? Shouldn't we rather add another boolean flag like `allow_as_base_asset`?

zrml/prediction-markets/src/orml_asset_registry.rs
Chralt98: Can you explain a little bit, what this file does and what it is needed for? Thanks!

zrml/market-commons/src/migrations.rs
```diff
@@ -14,30 +14,28 @@
...
+const MARKET_COMMONS_REQUIRED_STORAGE_VERSION: u16 = 4;
+const MARKET_COMMONS_NEXT_STORAGE_VERSION: u16 = 5;
+
 const MARKET_COMMONS: &[u8] = b"MarketCommons";
 const MARKETS: &[u8] = b"Markets";
-const MARKET_COMMONS_REQUIRED_STORAGE_VERSION: u16 = 2_u16;
-const MARKET_COMMONS_NEXT_STORAGE_VERSION: u16 = 3_u16;
 
-#[derive(Clone, Decode, Encode, Eq, PartialEq, RuntimeDebug, TypeInfo)]
+#[derive(Clone, Decode, Encode)]
```
Chralt98, line 38: Currently there are these derives for a market https://github.com/zeitgeistpm/zeitgeist/blob/f8f5c79c7c49c9011ab00ffa670d40dab48d44b7/primitives/src/market.rs#L30

Chralt98 (commented):
Essentially this is a great work to allow different base assets in prediction markets. I think about unintended consequences.

zrml/prediction-markets/src/mock.rs
```diff
@@ -397,6 +409,44 @@ impl ExtBuilder {
...
             .unwrap();
+        orml_tokens::GenesisConfig::<Runtime> {
+            balances: (0..69)
+                .into_iter()
+                .map(|idx| (idx, CurrencyId::ForeignAsset(100), INITIAL_BALANCE))
+                .collect(),
+        }
+        .assimilate_storage(&mut t)
+        .unwrap();
+        let custom_metadata = CustomMetadata { allow_in_pool: true, ..Default::default() };
```
maltekliemann, line 420: This should also be tagged with `#[cfg(feature = "parachainn")]`. Might be stylistically better to use https://docs.rs/cfg-if/0.1.2/cfg_if/ here.

zrml/prediction-markets/src/mock.rs
```diff
@@ -24,9 +24,10 @@
 use crate as prediction_markets;
 use frame_support::{
     construct_runtime, ord_parameter_types, parameter_types,
-    traits::{Everything, NeverEnsureOrigin, OnFinalize, OnInitialize},
+    traits::{Everything, GenesisBuild, NeverEnsureOrigin, OnFinalize, OnInitialize},
 };
 use frame_system::EnsureSignedBy;
+use orml_asset_registry::AssetMetadata;
```
maltekliemann, line 30: Should only be imported by `parachain` is activated.

maltekliemann (commented):
Just passing through and leaving a couple of hopefully useful comments. Do you understand what the purpose of using the `__private` prefix on the imports in the `orml_mock_registry` module is?

zrml/prediction-markets/src/lib.rs
```diff
@@ -600,7 +611,21 @@ mod pallet {
...
+            match base_asset {
+                Asset::Ztg => valid_base_asset = true,
+                #[cfg(feature = "parachain")]
+                Asset::ForeignAsset(fa) => {
+                    if let Some(metadata) = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)) {
+                        valid_base_asset = metadata.additional.allow_in_pool;
+                    }
+                }
+                _ => {}
+            }
```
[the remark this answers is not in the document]
vivekvpandya: I think we should add an error for UnregisteredAsset.

zrml/prediction-markets/src/orml_asset_registry.rs
vivekvpandya: No logical changes, just applied cargo fmt.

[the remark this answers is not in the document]
vivekvpandya: @Chralt98 this is required to implement traits on a mock structure to be used for AssetRegistry config item.
@maltekliemann should we move it in future when its required outside pm pallet?

zrml/market-commons/src/migrations.rs
vivekvpandya: but do we really need other traits derived ? I have taken reference from https://github.com/zeitgeistpm/zeitgeist/pull/732/files#diff-7eef70e0e24e29abbdd8ae4c5949b3d7633f1fa7b844e9b0e7ff3227cc344865R50 and it works fine in the test migration.

primitives/src/types.rs
```diff
@@ -121,3 +121,49 @@ pub struct ResultWithWeightInfo<R> {
...
+    PartialOrd,
+    TypeInfo,
+)]
+/// Custom XC asset metadata
+pub struct CustomMetadata {
+    /// XCM-related metadata.
+    pub xcm: XcmMetadata,
+
+    /// Whether an asset can be used in pools.
+    pub allow_as_base_asset: bool,
```
Chralt98, line 145: Now the description seems a little bit off.

zrml/market-commons/src/migrations.rs
Chralt98: I must admit that I don't know an answer to this question. So if you have the proof that it works, we should ignore this.

zrml/prediction-markets/src/lib.rs
```diff
@@ -599,7 +610,22 @@ mod pallet {
...
+                #[cfg(feature = "parachain")]
+                Asset::ForeignAsset(fa) => {
+                    if let Some(metadata) = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)) {
+                        Ok::<bool, Error<T>>(metadata.additional.allow_as_base_asset)
+                    } else {
+                        Err::<bool, Error<T>>(Error::<T>::UnregisteredForeignAsset)
+                    }
+                }
+                _ => Ok(false),
+            }?;
```
Chralt98, line 624: That looks kind of strange. If `valid_base_asset` contains the Error `UnregisteredForeignAsset`, what does happen when this reaches the below ensure check? I think this is quite confusing. 

```suggestion
            let mut valid_base_asset = false;
            match base_asset {
                Asset::Ztg => valid_base_asset = true,
                #[cfg(feature = "parachain")]
                Asset::ForeignAsset(fa) => {
                    let metadata = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)).ok_or(Error::<T>::UnregisteredForeignAsset);
                    valid_base_asset = metadata..additional.allow_as_base_asset;
                }
                _ => valid_base_asset = false,
            }?;
```

Chralt98 (commented):
Looks okay, when you ask me.

zrml/prediction-markets/src/lib.rs
Chralt98: I am sorry. I did not see [that.](https://github.com/zeitgeistpm/zeitgeist/pull/894#discussion_r1048946784) But the style with error seems still a little bit strange.

vivekvpandya: ? on line 622 will make function return error and it never reaches ensure on line 626. And test is updated to cover this case.

vivekvpandya: earlier it was like what @Chralt98 suggested but after @maltekliemann 's suggestion I changed it. I find both options fine.

Chralt98: It's just that I don't think that the following syntax is well understood at the first glance.

```rust
Ok::<bool, Error<T>>(true)
```

So I would like to stick with the "readable" version with `mut valid_base_asset`, but it's not that important. It just was so noticeable.

maltekliemann: Chris is absolutely right, that's not pretty. But if you just add a type hint to `valid_base_asset`, we can get rid of the other type hints.

```rust
            let valid_base_asset: bool = match base_asset {
                Asset::Ztg => Ok(true),
                #[cfg(feature = "parachain")]
                Asset::ForeignAsset(fa) => {
                    if let Some(metadata) = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)) {
                        Ok(metadata.additional.allow_as_base_asset)
                    } else {
                        Err(Error::<T>::UnregisteredForeignAsset)
                    }
                }
                _ => Ok(false),
            }?;
```

maltekliemann: You can also skip the entire result and just return an error where appropriate:

```rust
            let valid_base_asset = match base_asset {
                Asset::Ztg => true,
                #[cfg(feature = "parachain")]
                Asset::ForeignAsset(fa) => {
                    if let Some(metadata) = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)) {
                        metadata.additional.allow_as_base_asset
                    } else {
                        return Err(Error::<T>::UnregisteredForeignAsset.into());
                    }
                }
                _ => false,
            };
```

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
maltekliemann: I'd put it in `primitives/` now, but it's fine to keep it here, as well.

maltekliemann (changes requested):
Migration looks good. We need to fix how market bonds are unreserved and move a check to ensure that markets can't be edited in illegal ways.

zrml/prediction-markets/src/orml_asset_registry.rs
```diff
@@ -0,0 +1,265 @@
...
+                        location: &MultiLocation,
+                    ) -> Option<$asset_id> {
+                        for (curr_location, asset_id) in &self.location_to_asset {
+                            if curr_location == location {
+                                return Some(asset_id.clone());
+                            }
+                        }
+
+                        None
+                    }
```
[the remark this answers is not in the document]
vivekvpandya: Could you please explain more? location_to_asset is a Vec and function it iterating over it to find a match. :thinking:

zrml/market-commons/src/migrations.rs
```diff
@@ -14,3 +14,370 @@
...
+            assert_eq!(updated_market.creator, legacy_market.creator);
+            assert_eq!(updated_market.creation, legacy_market.creation);
+            assert_eq!(updated_market.creator_fee, legacy_market.creator_fee);
+            assert_eq!(updated_market.oracle, legacy_market.oracle);
+            assert_eq!(updated_market.metadata, legacy_market.metadata);
+            if let MarketType::Categorical(_) = legacy_market.market_type {
+                assert_eq!(updated_market.market_type, legacy_market.market_type);
+                assert_eq!(updated_market.report, legacy_market.report);
+                assert_eq!(updated_market.resolved_outcome, legacy_market.resolved_outcome);
+            };
```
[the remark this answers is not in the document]
vivekvpandya: took the code from similar old migration, but you are right this if let not relevant in this migration.

zrml/market-commons/src/migrations.rs
```diff
@@ -14,3 +14,370 @@
...
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn post_upgrade() -> Result<(), &'static str> {
+        use alloc::string::ToString;
+        use frame_support::traits::OnRuntimeUpgradeHelpersExt;
+        use scale_info::prelude::format;
+        let mut markets_count = 0_u32;
+        let legacy_markets_count_key = "legacy_markets_count_key".to_string();
+        for (market_id, updated_market) in storage_iter::<MarketOf<T>>(MARKET_COMMONS, MARKETS) {
```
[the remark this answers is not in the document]
vivekvpandya: okay, are there any other places in code too where we should use MarketCommonsPalletApi to make sure we are reading correct storage item? like `on_runtime_upgrade()` or `pre_upgrade()` ?

zrml/market-commons/src/migrations.rs
```diff
@@ -14,3 +14,370 @@
...
+            assert_eq!(
+                updated_market.base_asset,
+                Asset::Ztg,
+                "found unexpected base_asset in market. market_id: {:?}, base_asset: {:?}",
+                market_id,
+                updated_market.base_asset
+            );
+            let legacy_market: LegacyMarketOf<T> =
+                Self::get_temp_storage(&format!("{:?}", market_id.as_slice()))
+                    .expect("legacy market not found");
```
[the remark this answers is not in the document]
vivekvpandya: is it fine if I add market_id in panic message like below?
```rust
                   Self::get_temp_storage(&format!("{:?}", market_id.as_slice()))                                                       
                     .expect(&format!("legacy market not found for market_id {:?}", market_id)); 
```

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
maltekliemann: Yes, but `location_to_asset` is initialized as `Vec::new()` and it never changes, right?

zrml/prediction-markets/src/orml_asset_registry.rs
```diff
@@ -0,0 +1,265 @@
...
+//
+//
+// Centrifuge is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version (see http://www.gnu.org/licenses).
+// Centrifuge is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
```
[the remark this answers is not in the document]
maltekliemann: Sorry, correction: The copyright notice has to stay, of course. Only the license statement can be removed.

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
vivekvpandya: yes that is correct, but may be it will change in future PR when we add more tests. So the above code is correct right?

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
vivekvpandya: Yes , I think you already covered that https://github.com/zeitgeistpm/zeitgeist/pull/894#discussion_r1059453184

zrml/market-commons/src/migrations.rs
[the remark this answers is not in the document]
maltekliemann: Well, in `pre_upgrade` you're writing `LegacyMarket` objects, you don't have any other choice there. In the actualy migration, I think it's fine to use `put_storage_value` even when writing `Market` objects, because we're actually testing that code. My only concern is that tests shouldn't be "self-referential" (not sure if that's even the right word here). I believe one of my recent migrations had that problem.

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
maltekliemann: As far as I can tell, this is perfectly fine as a mock object. I was just wondering what the point of having `location_to_asset` was. It makes sense to keep it around though.

Did you ever figure out what the point of the `__private` prefix was?

maltekliemann (changes requested):
Thanks for merging the two migrations. Now that I see the diff, it's definitely the better approach. After an in-depth review, some additional problems have popped up:

- The base asset is used as collateral, so it must be used when buying/selling complete sets and when redeeming tokens. I've made a PR which (hopefully) covers all of this: https://github.com/zeitgeistpm/zeitgeist/compare/issue-226...issue-226-mkl-suggestion?diff=split. Please double check that it does before applying these changes.
- This oversight is not covered by our tests. That's very concerning. We need tests for the following scenarios:
  - Editing a market fails due to incorrect base asset (not registered/not allowed).
  - All tests that involve buying/selling complete sets and redeeming. You can probably use `test_case` to save _a lot_ of work and code duplications.
  - Destroying a market slashes the market account correctly.
  
  Let me know if you need any help with that.
- There are two open issues (see above) that should be resolved before we merge.

vivekvpandya:
> The base asset is used as collateral, so it must be used when buying/selling complete sets and when redeeming tokens. I've made a PR which (hopefully) covers all of this: https://github.com/zeitgeistpm/zeitgeist/compare/issue-226...issue-226-mkl-suggestion?diff=split. Please double check that it does before applying these changes.

I agree to this issue, I was thinking on same line so indeed we need a test for illustrating this fact.

zrml/prediction-markets/src/orml_asset_registry.rs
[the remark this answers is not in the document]
vivekvpandya: For `__private` it just seems some coding convention see https://stackoverflow.com/questions/73544273/having-both-public-and-private-code-shared-between-multiple-binaries-rust

Chralt98 (commented):
Apart from the comments of you, Malte, I could not find anything. @sea212 ready for your final review. But Harald, maybe you want to wait until the suggestions of Malte are applied.

maltekliemann (changes requested):
I'm still missing some tests here. Everything that has changed must be tested. So what did we change?

- `buy_complete_set`/`sell_complete_set` need to be tested. It's probably easiest to take the tests that already exist and use `test_case` to abstract the base asset.
- The asset used to collateralize complete sets is now not necessarily the same as the asset used for bonds. It would therefore be prudent to verify that whenever bonds

zrml/prediction-markets/Cargo.toml
```diff
@@ -64,7 +70,9 @@ std = [
     "frame-support/std",
     "frame-system/std",
     "orml-traits/std",
+    "orml-asset-registry/std",
```
sea212, line 73:
```suggestion
    "orml-asset-registry?/std",
```

zrml/market-commons/src/migrations.rs
```diff
@@ -14,3 +14,61 @@
...
+    pub fn get_on_chain_storage_version_of_swaps_pallet() -> StorageVersion {
+        let key = storage_prefix_of_swaps_pallet();
+        unhashed::get_or_default(&key)
+    }
+    #[allow(unused)]
+    pub fn put_storage_version_of_swaps_pallet(value: u16) {
+        let key = storage_prefix_of_swaps_pallet();
+        unhashed::put(&key, &StorageVersion::new(value));
+    }
+}
```
sea212, line 74: What do we need this for? I can't find where this is used.

```suggestion
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -48,6 +49,12 @@ mod pallet {
         Blake2_128Concat, BoundedVec, PalletId, Twox64Concat,
     };
     use frame_system::{ensure_signed, pallet_prelude::OriginFor};
+
+    #[cfg(feature = "parachain")]
+    use orml_traits::asset_registry::Inspect;
+    #[cfg(feature = "parachain")]
+    use zeitgeist_primitives::types::CustomMetadata;
```
sea212, line 56:
```suggestion
    #[cfg(feature = "parachain")]
    use {
        orml_traits::asset_registry::Inspect,
        zeitgeist_primitives::types::CustomMetadata,
    };
```

zrml/prediction-markets/src/orml_asset_registry.rs
```diff
@@ -0,0 +1,254 @@
...
+
+                thread_local! {
+                    pub static STATE: RefCell<
+                        RegistryState,
+                    > = RefCell::new(RegistryState::new());
+                }
+            }
+        }
+    };
+}
```
sea212, line 252: This is actually not necessary, we can achieve everything necessary with the asset registry pallet.
However, we still use the version `polkadot-v0.9.26` on asset registry, which unfortunately does not allow initializing genesis storage (we'd have to write raw keys and values...). So for now, this obviously is a won't fix. However, [from `polkadot-v0.9.29` onwards](https://github.com/open-web3-stack/open-runtime-module-library/compare/polkadot-v0.9.26...polkadot-v0.9.29#diff-973d213acaf59102ed6f6329665a78d77621db8e48f20758a6fdb9be6c75e675R114-R130), we can use the asset registry pallet for that. Can you create an issue to track that please?

zrml/market-commons/src/migrations.rs
vivekvpandya: If we keep it around it may be just hand for next time when we write migration. However I can remove it if you prefer that.

Chralt98: I think it's okay to leave it in, because of the common use. That's why we used a lot of `#[allow(unused)]`.

zrml/prediction-markets/src/orml_asset_registry.rs
Chralt98: https://github.com/zeitgeistpm/zeitgeist/issues/936

zrml/prediction-markets/src/tests.rs
```diff
@@ -2774,12 +3209,85 @@ fn edit_cycle_for_proposed_markets() {
...
+            Asset::ForeignAsset(100),
+            0,
+            CHARLIE,
+            MarketPeriod::Block(0..1),
+            get_deadlines(),
+            gen_metadata(2),
+            MarketType::Categorical(<Runtime as crate::Config>::MinCategories::get()),
+            MarketDisputeMechanism::SimpleDisputes,
+            ScoringRule::CPMM
+        ));
```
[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/894/commits/5fb067587c8e407b4508e7f9347f820e54cd72b8

zrml/prediction-markets/src/tests.rs
```diff
@@ -721,17 +784,24 @@ fn admin_destroy_market_correctly_cleans_up_accounts() {
...
         let pool_account = Swaps::pool_account_id(&pool_id);
         let market_account = MarketCommons::market_account(market_id);
-        let alice_ztg_before = AssetManager::free_balance(Asset::Ztg, &ALICE);
+        let alice_base_asset_before = AssetManager::free_balance(base_asset, &ALICE);
         assert_ok!(PredictionMarkets::admin_destroy_market(Origin::signed(SUDO), 0));
         assert_eq!(AssetManager::free_balance(Asset::CategoricalOutcome(0, 0), &pool_account), 0);
         assert_eq!(AssetManager::free_balance(Asset::CategoricalOutcome(0, 1), &pool_account), 0);
         assert_eq!(AssetManager::free_balance(Asset::CategoricalOutcome(0, 2), &pool_account), 0);
-        assert_eq!(AssetManager::free_balance(Asset::Ztg, &pool_account), 0);
+        assert_eq!(AssetManager::free_balance(base_asset, &pool_account), 0);
```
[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/894/commits/5fb067587c8e407b4508e7f9347f820e54cd72b8

Chralt98 (dismissed):

maltekliemann (dismissed):
Checks out from my side. Let's wait for @sea212's review.

@saboonikhil Do you need any additional info in the events to track balances of foreign assets that are traded on our chain?

saboonikhil:
> Checks out from my side. Let's wait for @sea212's review.
> 
> @saboonikhil Do you need any additional info in the events to track balances of foreign assets that are traded on our chain?

Just went through the changes.

Market structure already has `base_asset` parameter which now take its effect.

As of now, balances pallet records change in the Ztg balance for an account and tokens pallet for outcome assets, pool shares etc. Will changes to `base_asset`  also come under balances pallet?

vivekvpandya:
> > Checks out from my side. Let's wait for @sea212's review.
> > @saboonikhil Do you need any additional info in the events to track balances of foreign assets that are traded on our chain?
> 
> Just went through the changes.
> 
> Market structure already has `base_asset` parameter which now take its effect.
> 
> As of now, balances pallet records change in the Ztg balance for an account and tokens pallet for outcome assets, pool shares etc. Will changes to `base_asset` also come under balances pallet?

if base_asset is Ztg then it is pallet_balances, other wise it is orm_tokens.

saboonikhil:
> > > Checks out from my side. Let's wait for @sea212's review.
> > > @saboonikhil Do you need any additional info in the events to track balances of foreign assets that are traded on our chain?
> > 
> > 
> > Just went through the changes.
> > Market structure already has `base_asset` parameter which now take its effect.
> > As of now, balances pallet records change in the Ztg balance for an account and tokens pallet for outcome assets, pool shares etc. Will changes to `base_asset` also come under balances pallet?
> 
> if base_asset is Ztg then it is pallet_balances, other wise it is orm_tokens.

Alright, looking forward to see them on testnet 😎

zrml/prediction-markets/src/tests.rs
```diff
@@ -2909,99 +2909,6 @@ fn dispute_fails_unless_reported_or_disputed_market(status: MarketStatus) {
     });
 }
 
-#[test]
-fn it_resolves_a_disputed_market_to_default_if_dispute_mechanism_failed() {
```
Chralt98: Could you kindly explain why you removed this test?

zrml/prediction-markets/src/tests.rs
```diff
@@ -3977,78 +3884,6 @@ fn authorized_correctly_resolves_disputed_market() {
     });
 }
 
-#[test]
-fn on_resolution_defaults_to_oracle_report_in_case_of_unresolved_dispute() {
```
Chralt98: Could you kindly explain why you removed that test?

zrml/prediction-markets/src/tests.rs
vivekvpandya: main branch does not have these tests. https://github.com/zeitgeistpm/zeitgeist/commit/e991d9edbc76db6f2ecc2780a7caf4a4702cff6a#diff-ad4c4ea3d698c97bb435c01c3c30e79c5a60b34af1448fe64b7c87017054b4c6L2248

zrml/prediction-markets/src/tests.rs
```diff
@@ -337,15 +360,80 @@ fn create_market_fails_on_max_oracle_duration() {
...
+fn create_market_with_foreign_assets() {
     ExtBuilder::default().build().execute_with(|| {
-        simple_create_categorical_market(MarketCreation::Permissionless, 0..2, ScoringRule::CPMM);
+        let deadlines = Deadlines {
+            grace_period: <Runtime as crate::Config>::MaxGracePeriod::get(),
+            oracle_duration: <Runtime as crate::Config>::MaxOracleDuration::get(),
+            dispute_duration: <Runtime as crate::Config>::MaxDisputeDuration::get(),
+        };
+
+        // As per Mock asset_registry genesis ForeignAsset(420) has allow_as_base_asset set to false.
```
Chralt98, line 378: It could be useful here to have a method, which actually fills in the data to the mock asset registry, so that these comments are not required anymore. Just say: `I think it's fine`, if you don't like to change it here. I am okay with this too.

zrml/prediction-markets/src/tests.rs
```diff
@@ -2797,12 +3422,87 @@ fn edit_cycle_for_proposed_markets() {
     });
 }
 
+#[cfg(feature = "parachain")]
+#[test]
+fn edit_market_with_foreign_asset() {
```
Chralt98, line 3427: Ok, edit market check is here.

Chralt98 (commented):
Looks pretty good to allow different assets as the denominator in prediction markets.

zrml/prediction-markets/src/tests.rs
Chralt98: I see, I even think that I deleted them.. But it had the reason that those tests were duplicated. Thanks for the investigation.

Chralt98:
Really great work here Vivek! Love that.

zrml/market-commons/src/migrations.rs
sea212: We have exactly the same code in the prediction markets pallet. We should deduplicate it.

sea212 (dismissed):

maltekliemann (changes requested):
- It seems like the CI doesn't run the tests with `parachain`; it only runs `check`. I guess you could add
  ```
  cargo test -p zrml-prediction-markets --features parachain
  ```

  to `misc.sh` (not sure what the exact structure of those scripts is).
- One test is still failing.
- Is there a test for `create_market` failing if an invalid base asset is specified?
- There are some comments above that are still unaccounted for.

vivekvpandya:
> * Is there a test for `create_market` failing if an invalid base asset is specified?

please check test create_market_with_foreign_assets() it covers to failure where if a FA is not registered in AssetRegistry and if FA is registered but in metadata its not allowed to be used as base_asset.

zrml/prediction-markets/src/tests.rs
vivekvpandya: Its added in GenesisConfig, similar to how we add default balances to account, otherwise we need to mint balances in very test before we do anything. So I think this fine as comment will guide reader to mock registry.

zrml/prediction-markets/src/tests.rs
```diff
@@ -3859,17 +3880,24 @@ fn authorized_correctly_resolves_disputed_market() {
...
 
-        let charlie_balance = Balances::free_balance(&CHARLIE);
-        assert_eq!(charlie_balance, 1_000 * BASE + OracleBond::get());
-        let charlie_reserved_2 = Balances::reserved_balance(&CHARLIE);
+        if base_asset == Asset::Ztg {
+            let charlie_balance = AssetManager::free_balance(Asset::Ztg, &CHARLIE);
+            assert_eq!(charlie_balance, 1_000 * BASE + OracleBond::get());
+        } else {
+            let charlie_balance = AssetManager::free_balance(Asset::Ztg, &CHARLIE);
+            assert_eq!(charlie_balance, 1_000 * BASE + OracleBond::get());
```
Chralt98, line 3880: I see, the Bonds are always taken from ZTG, but not in the base asset.

Chralt98 (approved):
LGTM

scripts/tests/misc.sh
```diff
@@ -6,25 +6,35 @@ set -euxo pipefail
...
 . "$(dirname "$0")/aux-functions.sh" --source-only
 
-test_package_with_feature primitives default
-test_package_with_feature primitives std
+if [ -z "${1:-}" ]
+then
+    rustflags=""
+else
+    rustflags=$1
+fi
```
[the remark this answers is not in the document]
Chralt98: No then it's fine!

maltekliemann (commented):
Looks good now.

Just one question (I may have asked this before): Why are we using the `parachain` feature to begin with? **Edit.** Seems to me like `parachain` can be removed from `zrml_prediction_markets`. Only price to pay is longer compiler times since you need some XCM imports. Seems worth it, though. Am I wrong about this?

vivekvpandya:
> Looks good now.
> 
> Just one question (I may have asked this before): Why are we using the `parachain` feature to begin with? **Edit.** Seems to me like `parachain` can be removed from `zrml_prediction_markets`. Only price to pay is longer compiler times since you need some XCM imports. Seems worth it, though. Am I wrong about this?

Perhaps it would also bring in some testing time too as there will be XCM tests.
But I think that can be done in separate PR. I think it is better to conclude this PR at this point.

maltekliemann (approved):

PR #900, opened 2022-12-07
Tightly coupled market-commons with prediction markets.

### LINKED ISSUE #805: [PM][Swaps] Tightly couple market-commons pallet

vivekvpandya:
Fixes #805


maltekliemann (commented):
Double-booking the implementation of `MarketCommonsPalletApi` into the implementation of the `zrml_market_commons::Pallet::<T>` seems strange. Is there a particular reason why you did that, instead of replacing every `zrml_market_commons::Pallet::<T>` with `<zrml_market_commons::Pallet::<T> as MarketCommonsPalletApi>`?

vivekvpandya:
> Double-booking the implementation of `MarketCommonsPalletApi` into the implementation of the `zrml_market_commons::Pallet::<T>` seems strange. Is there a particular reason why you did that, instead of replacing every `zrml_market_commons::Pallet::<T>` with `<zrml_market_commons::Pallet::<T> as MarketCommonsPalletApi>`?

Frankly I did not get this idea when editing, I though for loose coupling we use MarketCommonsPalletApi and for tight coupling we directly use functions from pallet and don't need to import MarketCommonsPalletApi.

Chralt98:
Starting my review now..

maltekliemann:
> > Double-booking the implementation of `MarketCommonsPalletApi` into the implementation of the `zrml_market_commons::Pallet::<T>` seems strange. Is there a particular reason why you did that, instead of replacing every `zrml_market_commons::Pallet::<T>` with `<zrml_market_commons::Pallet::<T> as MarketCommonsPalletApi>`?
> 
> Frankly I did not get this idea when editing, I though for loose coupling we use MarketCommonsPalletApi and for tight coupling we directly use functions from pallet and don't need to import MarketCommonsPalletApi.

It's a matter of dependency. With loose coupling, you can _only_ access the config parameter `MarketCommons`, which implements `MarketCommonsPalletApi`. This allows you to use _any_ struct which implements that particular trait. With tight coupling, that config parameter is gone and we have to use the particular `zrml_market_commons` pallet, which happens to implement `MarketCommonsPalletApi`. In particular, the user has no choice in the matter. Before this change, they could have used their own "market manager". Now they _have_ to use ours.

But since the MDMs are still loosely coupled, I think we should move `MarketCommonsPalletApi` into the `primitives`. Is there any argument against this?

Chralt98:
You like to have the tight coupling of `zrml_swaps` in another PR?

vivekvpandya:
> You like to have the tight coupling of `zrml_swaps` in another PR?

yes if it is really required, as per discussion in one of the meeting we feel that swaps should not be tightly coupled with market-commons.

zrml/market-commons/src/lib.rs
```diff
@@ -152,10 +158,7 @@ mod pallet {
             <Markets<T>>::iter()
         }
 
-        fn market(
-            market_id: &Self::MarketId,
-        ) -> Result<Market<Self::AccountId, Self::BlockNumber, Self::Moment>, DispatchError>
-        {
+        fn market(market_id: &Self::MarketId) -> Result<MarketOf<T>, DispatchError> {
```
Chralt98, line 161: You can use this cool approach [here](https://github.com/zeitgeistpm/zeitgeist/blob/d6a3b8a5e1d056127c51f00f2f02be3409ebe742/zrml/market-commons/src/lib.rs#L181), [here](https://github.com/zeitgeistpm/zeitgeist/blob/d6a3b8a5e1d056127c51f00f2f02be3409ebe742/zrml/market-commons/src/lib.rs#L168), [here](https://github.com/zeitgeistpm/zeitgeist/blob/d6a3b8a5e1d056127c51f00f2f02be3409ebe742/zrml/market-commons/src/lib.rs#L238) and [here](https://github.com/zeitgeistpm/zeitgeist/blob/d6a3b8a5e1d056127c51f00f2f02be3409ebe742/zrml/market-commons/src/lib.rs#L156).

zrml/prediction-markets/src/lib.rs
```diff
@@ -122,10 +121,11 @@ mod pallet {
             // TODO(#618): Not implemented for Rikiddo!
             T::DestroyOrigin::ensure_origin(origin)?;
 
-            let market = T::MarketCommons::market(&market_id)?;
+            let market =
+                <zrml_market_commons::Pallet<T> as MarketCommonsPalletApi>::market(&market_id)?;
```
Chralt98, line 125: You can globally replace `<zrml_market_commons::Pallet<T> as MarketCommonsPalletApi>` by `<zrml_market_commons::Pallet<T>>` to shorten things up here.

```suggestion
            let market = <zrml_market_commons::Pallet<T>>::market(&market_id)?;
```

Chralt98 (changes requested):
I like the idea of the tightly couple.

zrml/prediction-markets/src/lib.rs
vivekvpandya: I wonder why clippy don't suggest that :thinking:

vivekvpandya:
> But since the MDMs are still loosely coupled, I think we should move MarketCommonsPalletApi into the primitives. Is there any argument against this?

so what is the motivation here? Just that above trait is imported from primitives and not from market_commons? I mean only implementer of that trait is MarketCommons so I am fine with any approach.

zrml/prediction-markets/src/lib.rs
maltekliemann: I wonder why I suggested it. My bad, of course we don't need the fully qualified syntax if there's no collisions between the function names.

maltekliemann:
> > But since the MDMs are still loosely coupled, I think we should move MarketCommonsPalletApi into the primitives. Is there any argument against this?
> 
> so what is the motivation here? Just that above trait is imported from primitives and not from market_commons? I mean only implementer of that trait is MarketCommons so I am fine with any approach.

Because the MDMs and `swaps` are still loosely coupled, people could potentially use them with some other struct that implements `MarketCommonsPalletApi`. If they wanted to do that, they would have to import `zrml_market_commons` to get the trait, which is an unnecessary import. Instead, they should be able to import the trait from the `primitives`.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -98,7 +98,7 @@ fn create_market_common<T: Config>(
         scoring_rule,
     }
     .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
-    let market_id = T::MarketCommons::latest_market_id()?;
+    let market_id = <zrml_market_commons::Pallet<T> as MarketCommonsPalletApi>::latest_market_id()?;
```
maltekliemann, line 101: This is inconsistent use of syntax. You use Chris' suggestion in `prediction-markets`, but my (bad) suggestion here.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2613,6 +2620,10 @@ mod pallet {
             T::PalletId::get().0
         }
 
+        fn market_account(market_id: MarketIdOf<T>) -> T::AccountId {
+            T::PalletId::get().into_sub_account_truncating(market_id.saturated_into::<u128>())
+        }
+
```
maltekliemann, line 2626: Why did you add this again?

```suggestion
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -2709,12 +2720,12 @@ mod pallet {
...
             outcome_report: &OutcomeReport,
         ) -> Result<Weight, DispatchError> {
-            let pool_id = if let Ok(el) = T::MarketCommons::market_pool(market_id) {
+            let pool_id = if let Ok(el) = <zrml_market_commons::Pallet<T>>::market_pool(market_id) {
                 el
             } else {
                 return Ok(T::DbWeight::get().reads(1));
             };
-            let market_account = T::MarketCommons::market_account(*market_id);
+            let market_account = Self::market_account(*market_id);
```
maltekliemann, line 2728:
```suggestion
            let market_account = <zrml_market_commons::Pallet<T>>::market_account(*market_id);
```

maltekliemann (changes requested):

zrml/prediction-markets/src/benchmarks.rs
vivekvpandya: oh missed this file.

zrml/prediction-markets/src/lib.rs
vivekvpandya: I think that now we have tight coupling we may perhaps able to remove `PredictionMarketsPalletId` from market_commons Config. however if we want to keep `market_account()` in the trait then I accept your suggested edit.

vivekvpandya: but I think we can't get rid of `PredictionMarketsPalletId` as we can't access prediction_market config in matket_commons.

Chralt98 (approved):
LGTM

maltekliemann (approved):
:thumbsup:

Issue #698, opened 2022-07-07
Should a non-oracle reported stake ZTG?


maltekliemann:
If the oracle fails to report and another user steps up to the plate, should they stake ZTG?

I think so. Although we don't want to discourage people from reporting the market, we definitely want to discourage people from reporting nonesense. This should not be an action that can be taken without consequence.

sea212:
I see your point and I agree. The reason why it is okay for now is that we still have the dispute mechanism. The question is what stake should we take? If it is too low, it doesn't make sense. If it is too high, it becomes exclusive. Another question is what should happen with the oracle stake (I guess slash is the way to go).

I will reduce the priority to medium since this is not pressing (this is a rare case and nothing really concerning happens) and since we have many other more pressing issues at high priority.

maltekliemann:
Two possible suggestions:

- Let the reporter stake `OracleBond`. If their report is correct, reward them (part of) the original oracle bond (which was slashed because the oracle did not report). If their report is incorrect, slash their oracle bond. This incentivizes making a report, but keeps people honest at the same time.
- Don't allow non-oracle reports and go straight to dispute if the oracle fails to report

I really like the second approach.

Chralt98:
Use the issue https://github.com/zeitgeistpm/zeitgeist/issues/868 for this. So I will remove this from the milestone.

Chralt98:
> Don't allow non-oracle reports and go straight to dispute if the oracle fails to report

We decided to allow a non-oracle report, because it's always good to have something we can fallback to. For this, we decided to reserve an `OutsiderBond`, when the oracle did not manage to report in time.

PR #903, opened 2022-12-09
[PM] Reserve `OutsiderBond` for report, when the oracle fails to report

### LINKED ISSUE #868: ZIP-0 Reserve an `OutsiderBond` for `report`, when the oracle fails to report
This is related to https://github.com/zeitgeistpm/zeitgeist/issues/698 . Don't forget to close #698, when this is resolved.

Slash the deposit, when the `resolved_outcome` is not the reported outcome in `on_resolution`.

Unreserve the `OutsiderBond` when the market resolves and [repatriate](https://paritytech.github.io/substrate/master/frame_support/traits/trait.ReservableCurrency.html#tymethod.repatriate_reserved) the `OracleBond` to the outsider.

Keep in mind that the `OutsiderBond` should be transferred to the right disputors like [here](https://github.com/zeitgeistpm/zeitgeist/blob/f8f5c79c7c49c9011ab00ffa670d40dab48d44b7/zrml/prediction-markets/src/lib.rs#L2341).

Chralt98:
Fixes #868. Fixes https://github.com/zeitgeistpm/zeitgeist/issues/698.


Chralt98:
Waiting for https://github.com/zeitgeistpm/zeitgeist/pull/892 to use the approach for tracking bonds.

runtime/battery-station/src/parameters.rs
```diff
@@ -183,6 +183,9 @@ parameter_types! {
     /// (Slashable) The orcale bond. Slashed in case the final outcome does not match the
     /// outcome the oracle reported.
     pub const OracleBond: Balance = 50 * CENT;
+    /// (Slashable) A bond for an outcome reporter, who is not the oracle.
+    /// Slashed in case the final outcome does not match the outcome by the outsider.
+    pub const OutsiderBond: Balance = 25 * CENT;
```
Chralt98, line 188: Need your opinion on this value. Open to discuss.

runtime/zeitgeist/src/parameters.rs
```diff
@@ -183,6 +183,9 @@ parameter_types! {
     /// (Slashable) The orcale bond. Slashed in case the final outcome does not match the
     /// outcome the oracle reported.
     pub const OracleBond: Balance = 200 * BASE;
+    /// (Slashable) A bond for an outcome reporter, who is not the oracle.
+    /// Slashed in case the final outcome does not match the outcome by the outsider.
+    pub const OutsiderBond: Balance = 100 * BASE;
```
Chralt98, line 188: Open to discuss!

zrml/prediction-markets/src/migrations.rs
```diff
@@ -63,84 +72,66 @@ pub struct OldMarket<AI, BN, M> {
...
-                    resolved_outcome: old_market.resolved_outcome,
-                    dispute_mechanism: old_market.dispute_mechanism,
-                    deadlines: old_market.deadlines,
-                    bonds: MarketBonds { creation, oracle },
-                };
-                (key, new_market)
-            })
-            .collect::<Vec<_>>();
+        let mut translated = 0u64;
+        zrml_market_commons::Markets::<T>::translate::<OldMarketOf<T>, _>(|_key, old_market| {
```
Chralt98, line 102: I am not sure, if I should rather use [translate_values](https://github.com/paritytech/substrate/blob/8d17a16a5653974ec17ed617306606e9dbb6e4b4/frame/support/src/storage/mod.rs#L1232-L1264). I used `translate`, because of [this example](https://github.com/paritytech/substrate/blob/7ba1ef526fbcc224d3bd446c83ee66a83cacec49/frame/assets/src/migration.rs#L70-L79).

I read, that `translate` has the downside, that it silently ignores wrong decodings. Means, when the `OldMarketOf<T>` does not match the storage structure currently present before the runtime upgrade, then we wouldn't notice and it would not change the values. This means a check with `try-runtime` `pre-upgrade` and `post-upgrade` is mandatory to check that the storage was correctly translated! In example one could check that all markets are still present ([count all markets](https://github.com/paritytech/substrate/blob/7ba1ef526fbcc224d3bd446c83ee66a83cacec49/frame/assets/src/migration.rs#L102-L105)) and if [all values are as expected](https://github.com/paritytech/substrate/blob/7ba1ef526fbcc224d3bd446c83ee66a83cacec49/frame/assets/src/migration.rs#L116-L118) after the runtime upgrade.

Chralt98:
```rust
        let old_markets = storage_key_iter::<MarketIdOf<T>, OldMarketOf<T>, Blake2_128Concat>(
            MARKET_COMMONS,
            MARKETS,
        )
        .collect::<BTreeMap<_, _>>();
        Self::set_temp_storage(old_markets, "old_markets");
```

In combination with the above (`pre_upgrade`), this is dangerous. If `OldMarketOf<T>` is the wrong structure for the old markets, then the `BTreeMap` would be empty and counter is zero.

Assume `translate` fails for each element, because of the wrong decoding struct. From [here](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.StorageMap.html#method.translate) `NOTE: If a value fail to decode because storage is corrupted then it is skipped.`, then we have the same storage structure as before the upgrade, but the new decoding fails then (no items in Markets). In this particular case this means, that Markets is empty with the new decoding structure.

To mitigate this is to count the items inside `Markets` and assert against `Markets` counter after the migration.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2390,11 +2461,56 @@ mod pallet {
...
+                    let negative_imbalance = Self::slash_oracle_bond(market_id, None)?;
+                    overall_imbalance.subsume(negative_imbalance);
+                }
             } else {
-                let imbalance = Self::slash_oracle_bond(market_id, None)?;
-                overall_imbalance.subsume(imbalance);
+                // outsider report
+                if is_correct {
+                    // reward outsider reporter with oracle bond
+                    let missing = Self::repatriate_oracle_bond(market_id, &report.by)?;
```
Chralt98, line 2505: I don't want a fail in `on_resolution` because of `repatriate_reserved_named` here. Should we ignore the result of [`repatriate`](https://github.com/paritytech/substrate/blob/911f65be7949342e12601cc3015d5c47e9339de4/frame/balances/src/lib.rs#L1024-L1041)?

zrml/prediction-markets/src/migrations.rs
Chralt98: The mistake would be noticeable now with the log error from this commit https://github.com/zeitgeistpm/zeitgeist/pull/903/commits/89d75f8c51ac06b737e1ac9b0ca18197475043eb

runtime/battery-station/src/parameters.rs
maltekliemann: Either same as oracle bond or significantly higher. We're looking to discourage users from making a report unless they are absolutely certain they're doing the right thing.

runtime/zeitgeist/src/parameters.rs
maltekliemann: Either same as oracle bond or significantly higher. We're looking to discourage users from making a report unless they are absolutely certain they're doing the right thing.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2390,11 +2461,56 @@ mod pallet {
...
+                if let Some(bond) = &market.bonds.outsider {
+                    if !bond.is_settled {
+                        return true;
+                    }
+                }
+                false
+            };
+
+            let unreserve_outsider = || -> DispatchResult {
+                if outsider_bond_pending() {
```
maltekliemann, line 2480: Should we warn if the outsider bond is not pending? Seems like a serious logic error to me.

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
 }
 
+#[test]
+fn report_reserves_outsider_report_bond_and_unreserved_by_on_resolution() {
+    ExtBuilder::default().build().execute_with(|| {
+        let end = 100;
+        assert_ok!(PredictionMarkets::create_market(
+            Origin::signed(ALICE),
+            BOB,
+            MarketPeriod::Block(0..100),
```
maltekliemann, line 3441:
```suggestion
            MarketPeriod::Block(0..end),
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -2347,6 +2412,12 @@ mod pallet {
                         err
                     );
                 }
+
```
maltekliemann, line 2415: Should we use `repatriate` here, as well?

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
+
+#[test]
+fn disputed_market_results_in_rewarding_outsider() {
+    ExtBuilder::default().build().execute_with(|| {
+        let end = 100;
+        let alice_balance_before = Balances::free_balance(&ALICE);
+        assert_ok!(PredictionMarkets::create_market(
+            Origin::signed(ALICE),
+            BOB,
+            MarketPeriod::Block(0..100),
```
maltekliemann, line 3492:
```suggestion
            MarketPeriod::Block(0..end),
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
+
+#[test]
+fn outsider_reports_wrong_outcome() {
+    ExtBuilder::default().build().execute_with(|| {
+        let end = 100;
+        let alice_balance_before = Balances::free_balance(&ALICE);
+        assert_ok!(PredictionMarkets::create_market(
+            Origin::signed(ALICE),
+            BOB,
+            MarketPeriod::Block(0..100),
```
maltekliemann, line 3558:
```suggestion
            MarketPeriod::Block(0..end),
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
+        // Check that the outsider gets the OracleBond together with the OutsiderBond
+        assert_eq!(
+            Balances::free_balance(&CHARLIE),
+            charlie_balance_before + OracleBond::get() + OutsiderBond::get()
+        );
+    });
+}
+
+#[test]
+fn disputed_market_results_in_rewarding_outsider() {
```
maltekliemann, line 3485: There are four situations, in case of a dispute:

- Oracle reports and is correct
- Oracle reports and is not correct
- Outsider reports and is correct
- Outsider reports and is incorrect

Do we have tests for all four?

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
+        assert_eq!(Balances::free_balance(&CHARLIE), charlie_balance_before - OutsiderBond::get());
+        let charlie_balance_before = Balances::free_balance(&CHARLIE);
+        // on_resolution called
+        run_blocks(market.deadlines.dispute_duration);
+        assert_eq!(Balances::reserved_balance(&CHARLIE), SENTINEL_AMOUNT);
+        // Check that the outsider gets the OracleBond together with the OutsiderBond
+        assert_eq!(
+            Balances::free_balance(&CHARLIE),
+            charlie_balance_before + OracleBond::get() + OutsiderBond::get()
+        );
```
maltekliemann, line 3480: Not important, but definitely makes sense to have this.

```suggestion
        );
        let market = MarketCommons::market(&0).unwrap();
        assert!(market.bonds.outsider.unwrap().is_settled);
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
...
+            ScoringRule::CPMM,
+        ));
+
+        let outsider = CHARLIE;
+        // Reserve a sentinel amount to check that we don't unreserve too much.
+        assert_ok!(Balances::reserve_named(
+            &PredictionMarkets::reserve_id(),
+            &outsider,
+            SENTINEL_AMOUNT
+        ));
```
maltekliemann, line 3507: Random thought: Can we make it part of the mock's genesis to reserve `SENTINEL_AMOUNT` from everyone's balances?

zrml/prediction-markets/src/tests.rs
maltekliemann: In case of an _undisputed_ report, do we have a test for "outsider reports and is incorrect"?

maltekliemann (changes requested):
Looking good, just a couple of small suggestions. :thumbsup: `translate` seems very useful.

zrml/prediction-markets/src/lib.rs
Chralt98: 76fadc661cc78dd3fc73eaa38f5ed00401d6d325

runtime/zeitgeist/src/parameters.rs
Chralt98: Wow, couldn't be more impressed by your mathematical thinking.

Chralt98: 2399ceb0b3fd594671ffe4eb320429cdb1c3fda6

runtime/battery-station/src/parameters.rs
Chralt98: 2399ceb0b3fd594671ffe4eb320429cdb1c3fda6

zrml/prediction-markets/src/migrations.rs
```diff
@@ -43,13 +43,22 @@ use zeitgeist_primitives::types::{
...
 use zrml_market_commons::Pallet as MarketCommonsPallet;
 
+#[cfg(any(feature = "try-runtime", test))]
 const MARKET_COMMONS: &[u8] = b"MarketCommons";
+#[cfg(any(feature = "try-runtime", test))]
 const MARKETS: &[u8] = b"Markets";
-const MARKET_COMMONS_REQUIRED_STORAGE_VERSION: u16 = 4;
-const MARKET_COMMONS_NEXT_STORAGE_VERSION: u16 = 5;
+
+const MARKET_COMMONS_REQUIRED_STORAGE_VERSION: u16 = 5;
```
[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/903/commits/4f90a62a2778b3c80c54936dd17fd7a6c5da1e77

zrml/prediction-markets/src/migrations.rs
```diff
@@ -63,97 +72,102 @@ pub struct OldMarket<AI, BN, M> {
...
             MARKETS,
         )
         .collect::<BTreeMap<_, _>>();
         Self::set_temp_storage(old_markets, "old_markets");
+
+        let markets = Markets::<T>::iter_keys().count() as u32;
+        let decodable_markets = Markets::<T>::iter_values().count() as u32;
+        if markets != decodable_markets {
+            // This is not necessarily an error, but can happen when there are Calls
+            // in an Agenda that are not valid anymore with the new runtime.
```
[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/903#discussion_r1063236208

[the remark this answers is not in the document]
Chralt98: Only this part was copied:

```rust
        let markets = Markets::<T>::iter_keys().count() as u32;
        let decodable_markets = Markets::<T>::iter_values().count() as u32;
        if markets != decodable_markets {
            // This is not necessarily an error, but can happen when there are Calls
            // in an Agenda that are not valid anymore with the new runtime.
            log::error!(
                "Can only decode {} of {} markets - others will be dropped",
                decodable_markets,
                markets
            );
        } else {
            log::info!("Markets: {}, Decodable Markets: {}", markets, decodable_markets);
        }
```

zrml/prediction-markets/src/lib.rs
Chralt98: 9e42f0d25f51ad1c975798dc3e44b7915d91d367

Chralt98: If we warn, then this warning will come for every market, that has an outsider reporter (with no bond). Are you okay with that? I think it depends on the amount of markets which are in the `Disputed` or `Reported` state, which have got outsiders. If the number is zero, then we are good to go. I just want to avoid the spam of these warnings for many resolution. But I expect less markets, which already have got outsider reports.

zrml/prediction-markets/src/lib.rs
Chralt98: Yeah indeed, we should!

bf6920c52d92d11884b406e0cc113368616e759e

zrml/prediction-markets/src/tests.rs
```diff
@@ -3431,6 +3431,187 @@ fn on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_mark
     });
 }
```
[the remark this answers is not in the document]
Chralt98: I could find the following additional tests, which include the outsider report:

`on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_market_on_outsider_report`

- only report (without dispute)
- shows that the Oracle Bond of the market creator was slashed
- only ValidityBond was returned

`on_resolution_correctly_reserves_and_unreserves_bonds_for_permissionless_market_with_disputed_outcome_with_outsider_report`

- permissionless
- report and simples disputes, while the last dispute is the outcome by the original outsider report
- shows that the Oracle Bond of the market creator was slashed
- only ValidityBond was returned

`on_resolution_correctly_reserves_and_unreserves_bonds_for_advised_approved_market_with_disputed_outcome_with_outsider_report`

- advised
- report and simples disputes, while the last dispute is the outcome by the original outsider report
- shows that the Oracle Bond of the market creator was slashed

`on_resolution_correctly_reserves_and_unreserves_bonds_for_approved_advised_market_on_outsider_report`

- advised
- only report (without dispute)
- shows that the Oracle Bond of the market creator was slashed

My added tests:

`report_reserves_outsider_report_bond_and_unreserved_by_on_resolution`

- only report (without dispute)
- shows that the Oracle Bond was moved to the outsider and the Outsider Bond was returned

`disputed_market_results_in_rewarding_outsider`

- report and simples disputes, while the last dispute is the outcome by the original outsider report
- shows that the Oracle Bond was moved to the outsider and the Outsider Bond was returned

`outsider_reports_wrong_outcome`

- report and simples disputes, while the last dispute is *not* the outcome by the original outsider report
- shows that the last disputor gets the oracle and outsider bonds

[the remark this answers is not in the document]
Chralt98: eaab9bd17668d3111d49753e21c5971745a66c03

zrml/prediction-markets/src/tests.rs
Chralt98: - [ ] Oracle reports and is correct
- [ ] Oracle reports and is not correct
- [ ] Outsider reports and is correct
- [ ] Outsider reports and is incorrect

Chralt98: Regarding your last question. If nobody disputed, then everybody accepted the outsider report (in case it exists). So the case that it's incorrect is not defined. I think you mean [this case](https://github.com/zeitgeistpm/zeitgeist/blob/eaab9bd17668d3111d49753e21c5971745a66c03/zrml/prediction-markets/src/lib.rs#L2464-L2469). It is covered [here](https://github.com/zeitgeistpm/zeitgeist/blob/eaab9bd17668d3111d49753e21c5971745a66c03/zrml/prediction-markets/src/tests.rs#L3391-L3456).

The case that is actually missing is a disputed market on which the outsider reported and the resolved disputed outcome is the same as the outsider report outcome.

zrml/prediction-markets/src/tests.rs
Chralt98:
```
	#[pallet::genesis_build]
	impl<T: Config<I>, I: 'static> GenesisBuild<T, I> for GenesisConfig<T, I> {
		fn build(&self) {
			let total = self.balances.iter().fold(Zero::zero(), |acc: T::Balance, &(_, n)| acc + n);
			<TotalIssuance<T, I>>::put(total);

			<StorageVersion<T, I>>::put(Releases::V2_0_0);

			for (_, balance) in &self.balances {
				assert!(
					*balance >= <T as Config<I>>::ExistentialDeposit::get(),
					"the balance of any account should always be at least the existential deposit.",
				)
			}

			// ensure no duplicates exist.
			let endowed_accounts = self
				.balances
				.iter()
				.map(|(x, _)| x)
				.cloned()
				.collect::<std::collections::BTreeSet<_>>();

			assert!(
				endowed_accounts.len() == self.balances.len(),
				"duplicate balances in genesis."
			);

			for &(ref who, free) in self.balances.iter() {
				assert!(T::AccountStore::insert(who, AccountData { free, ..Default::default() })
					.is_ok());
			}
		}
	}
```

> T::AccountStore::insert(who, AccountData { free, ..Default::default() })

This sets `reserved` to Default::default(), which is zero. So I hoped, that we could configure that, but we can't. So we need to set it in the executable environment. This means the smallest would be a method, which reserves the sentinel amount for each test case. We could do that, but I am also okay to do it only, if we want to check against reserves in the test.

Chralt98: b85409cce1a3c14c6d8398154ed7705c10c38a11

zrml/prediction-markets/src/lib.rs
maltekliemann: The number of marktes should be very low, so I'm perfectly okay with keeping the warning.

zrml/prediction-markets/src/tests.rs
maltekliemann: > Regarding your last question. If nobody disputed, then everybody accepted the outsider report (in case it exists). So the case that it's incorrect is not defined. I think you mean [this case](https://github.com/zeitgeistpm/zeitgeist/blob/eaab9bd17668d3111d49753e21c5971745a66c03/zrml/prediction-markets/src/lib.rs#L2464-L2469). It is covered [here](https://github.com/zeitgeistpm/zeitgeist/blob/eaab9bd17668d3111d49753e21c5971745a66c03/zrml/prediction-markets/src/tests.rs#L3391-L3456).

I have no idea what I meant. If the report is undisputed, then by definition the outsider report is correct. Sorry, no idea what that was supposed to mean.

zrml/prediction-markets/src/tests.rs
maltekliemann: Ah, that's too bad. I think abstracting them into a function is a good solution.

"Just one more thing..." If we correctly place `debug_assert` calls, do we even need the sentinel reserves anymore?

Chralt98: Good thought. The reason for sentinel reserves is to ensure, that we recognise if we unreserve or slash too much than expected. You might mean to put `debug_assert` calls into the runtime code, not tests, otherwise you could just use normal asserts. Yes, you could place `debug_assert` inside the runtime code to check if we even have this much of funds we want to unreserve or slash.

Chralt98: But you could also do it the way we already have it ;-)

zrml/prediction-markets/src/migrations.rs
```diff
@@ -187,13 +198,13 @@ impl<T: Config + zrml_market_commons::Config> OnRuntimeUpgrade
             assert_eq!(new_market.report, old_market.report);
             assert_eq!(new_market.resolved_outcome, old_market.resolved_outcome);
             assert_eq!(new_market.dispute_mechanism, old_market.dispute_mechanism);
-            assert_eq!(new_market.bonds.oracle.unwrap().value, T::OracleBond::get());
```
maltekliemann, line 190: The `assert_eq!` for the `base_asset` can be replaced by a normal `assert_eq!(new_market.base_asset, old_market.base_asset)` at this point.

zrml/prediction-markets/src/tests.rs
```diff
@@ -63,6 +63,30 @@ fn gen_metadata(byte: u8) -> MultiHash {
...
+    assert_ok!(Balances::reserve_named(&PredictionMarkets::reserve_id(), &FRED, SENTINEL_AMOUNT));
+    assert_eq!(Balances::reserved_balance(&ALICE), SENTINEL_AMOUNT);
+    assert_eq!(Balances::reserved_balance(&BOB), SENTINEL_AMOUNT);
+    assert_eq!(Balances::reserved_balance(&CHARLIE), SENTINEL_AMOUNT);
+    assert_eq!(Balances::reserved_balance(&DAVE), SENTINEL_AMOUNT);
+    assert_eq!(Balances::reserved_balance(&EVE), SENTINEL_AMOUNT);
+    assert_eq!(Balances::reserved_balance(&FRED), SENTINEL_AMOUNT);
+}
+
+fn check_reserve(account: &AccountIdTest, expected: Balance) {
```
maltekliemann, line 86: Yeah, this is a good solution, because it abstracts away the `SENTINEL_AMOUNT` from the tester (at least up to the point where a test fails...).

maltekliemann (changes requested):
Yeah, nice. Couple o' nitpicks, but nothing really problematic.

maltekliemann:
Just realized that you should also document the new functions.

maltekliemann:
About the outsider bond. Anything `>= 2 * OracleBond::get()` should be fine to prevent dumb guessing, but the question is if it should be even higher. Disputes currently cost ten times of the oracle bond.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2422,11 +2542,44 @@ mod pallet {
...
-                let imbalance = Self::slash_oracle_bond(market_id, None)?;
-                overall_imbalance.subsume(imbalance);
+                // outsider report
+                if is_correct {
+                    // reward outsider reporter with oracle bond
+                    Self::repatriate_oracle_bond(market_id, &report.by)?;
+                    unreserve_outsider()?;
+                } else {
+                    let oracle_imbalance = Self::slash_oracle_bond(market_id, None)?;
+                    let outsider_imbalance = slash_outsider()?;
```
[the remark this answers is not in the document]
Chralt98: I see your point. T

> What if the market was misconfigured and the market ends too soon so that there's no outcome to report. The oracle solves this problem by not reporting.

The implementation before this PR had this problem too. The oracle had `oracle_duration` time to wait until an outsider could ignorantly report an outcome.

> Seems unfair to the oracle, which always acted with good intent.

So the main problem is how to recognise that a market ended too soon. If we could find a solution for this, we could extend the oracle duration.

> What I'm getting at is that this setup is incentivizing the oracle in this situation to report a random outcome in the hope of guessing the correct outcome so that it doesn't get slashed.

I see. Yeah, we should better find a way of making the oracle duration more flexible as described above.

> What's annoying about this particular situation is that even if our dispute system has a means of delaying the resolution of a market, we still need to trigger a dispute in order to delay the market. But the dispute can only be triggered if someone has reported.

Exactly! We should find a way to delay the market beforehand. So that we have both worlds: Allow a delay by the MDM and by some force which knows when a market couldn't have a final outcome at this time.

Ok, let me brainstorm: your approach (initiate a dispute) suits it well I think. But how to determine whether an oracle forgot to submit a report and an oracle which by purpose did not report because the final outcome was not present?

So we need to either unreserve or slash the oracle bond based on that. When we say: always unreserve the OracleBond when the oracle did not report in time, then the oracle does not get slashed even in the case that the final outcome was present and the oracle just forgot to submit it or worse: by intention.

Another approach: Just let the oracle delay the oracle duration itself. If the advisory committee disagrees on the delay, then the oracle bond gets slashed.

Chralt98:
> Just realized that you should also document the new functions.

I only see the macro function "impl_is_bond_pending" which comes in my mind. Which functions do you exactly mean?

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: > But how to determine whether an oracle forgot to submit a report and an oracle which by purpose did not report because the final outcome was not present?

I think this boils down to whether or not the dispute was justified. In other words:

- If the report period ends, the oracle didn't report and the MDM decides that the report needs to be delayed, then the oracle acted correctly and is not punished.
- If the report period ends, the oracle didn't report and the MDM decides that the market should resolve to outcome A, then the oracle is punished.

But this results in problems if the outcome becomes known between the end of the report period and the end of the dispute. I'm okay with the oracle getting punished in this situations. There are always going to be loopholes and people thinking they've been treated unfairly.

Another option would be to let the MDM decide whether or not to punish the oracle.

> Just let the oracle delay the oracle duration itself. If the advisory committee disagrees on the delay, then the oracle bond gets slashed.

That's kind of like raising a dispute with `authorized` as MDM, right? For advised markets, this would be perfectly fine. In that case, it's the AC's fault that the market wasn't correctly configured. For permissionless markets, this puts too much power in the AC's hands. In this case, I would really prefer a dispute.

I think this warrants it's own ZIP. It definitely doesn't belong into this PR.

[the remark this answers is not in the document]
Chralt98: > and the MDM decides that the report needs to be delayed

How would you let the MDM decide whether a report needs to be delayed? In case of authorized it's clear, that the advisory committee could call a new extrinsic which determines that. We need to properly define new requirements here for authorized and court. How does the delay agreement look like in detail. Who decides how, that a report delay is justified. It feels pretty general especially in the case of court that the MDM just decides. Who, how, when, what?

Yeah, I agree, it seems like a whole own ZIP.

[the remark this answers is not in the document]
Chralt98: One thing to keep in mind with this is that the oracle should propose an outcome. Otherwise I could as market creator just set the oracle duration to the minimum and just wait with my oracle outcome submit until the oracle duration is over to eventually get my bond back. Market creators are then incentivised to set the periods incorrectly so that it's justified that the market ended too soon.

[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/issues/956

maltekliemann (dismissed):

zrml/prediction-markets/src/lib.rs
```diff
@@ -2423,11 +2548,44 @@ mod pallet {
...
+            let unreserve_outsider = || -> DispatchResult {
+                if Self::is_outsider_bond_pending(market_id, market, true) {
+                    Self::unreserve_outsider_bond(market_id)?;
+                }
+                Ok(())
+            };
+
+            let slash_outsider = || -> Result<NegativeImbalanceOf<T>, DispatchError> {
+                if Self::is_outsider_bond_pending(market_id, market, true) {
+                    let imbalance = Self::slash_outsider_bond(market_id, None)?;
```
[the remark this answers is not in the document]
Chralt98: I see your point. This is essentially also my concern about resolutions. There are many places inside `on_resolution` where the code potentially could throw an error. That's why we used the approach of `is_outsider_bond_pending` here and are logging the warning. Whenever we add some error case to `slash_outsider_bond` it is not covered by `is_outsider_bond_pending` anymore. But with the current solution there is no way `slash_outsider_bond` could fail, because the assumptions are ensured in the `if` statement above.

[This](https://github.com/zeitgeistpm/zeitgeist/blob/a5f184a12ae0c8850b0d8b907d7800651f0c077d/zrml/prediction-markets/src/lib.rs#L147) is covered by the mentioned `if` statement and the [market existence](https://github.com/zeitgeistpm/zeitgeist/blob/a5f184a12ae0c8850b0d8b907d7800651f0c077d/zrml/prediction-markets/src/lib.rs#L193) is covered by the [`resolution_manager` here](https://github.com/zeitgeistpm/zeitgeist/blob/a5f184a12ae0c8850b0d8b907d7800651f0c077d/zrml/prediction-markets/src/lib.rs#L2896).

maltekliemann (approved):

vivekvpandya (approved):

PR #908, opened 2022-12-14
Add maximum lifetime for markets

### LINKED ISSUE #624: [PM] Consider setting an upper bound for the lifetime of a market
See https://github.com/zeitgeistpm/zeitgeist/pull/614#issuecomment-1126763123

maltekliemann:
By _lifetime_ I mean the duration between creating a market and market close. Together with the limits for the members of the `deadlines` parameter, this implies a realistic limit for how long a market can exist on our chain.

I also remove `MaxMarketPeriod`, which was included to resolve the informational issue 5.4.2 of the Chaintroopers security audit: https://blog.zeitgeist.pm/zeitgeist-completes-all-code-audits/, also found here: zeitgeistpm/runtime-audit-2#14. We now use saturated arithmetic everywhere, which fixes the overflow issues, and the newly included `MaxMarketLifetime*` constants ensure that (unless misconfigured) it is impossible to create a market which cannot be properly resolved (at least for the next thousand years, at which point using a `u32` for the block number will be a problem). So I consider this issue to still be solved, even if I remove `MaxMarketPeriod`.

Fixes #624.


primitives/src/constants.rs
```diff
@@ -77,6 +77,9 @@ pub const GLOBAL_DISPUTES_LOCK_ID: [u8; 8] = *b"zge/gdlk";
 pub const LM_PALLET_ID: PalletId = PalletId(*b"zge/lymg");
 
 // Prediction Markets
+pub const MAX_MARKET_LIFETIME_IN_BLOCKS: BlockNumber = 365 * BLOCKS_PER_DAY;
+pub const MAX_MARKET_LIFETIME_IN_MOMENTS: Moment =
+    (MAX_MARKET_LIFETIME_IN_BLOCKS as u64) * (MILLISECS_PER_BLOCK as u64);
```
sea212, line 82: Missing Docstring.

runtime/battery-station/src/parameters.rs
```diff
@@ -155,31 +155,35 @@ parameter_types! {
...
-    // 60_000 = 1 minute. Should be raised to something more reasonable in the future.
-    /// Minimum number of milliseconds a Rikiddo market must be in subsidy gathering phase.
-    pub const MinSubsidyPeriod: Moment = 60_000;
+    /// Maximum string length for edit reason.
+    pub const MaxEditReasonLen: u32 = 1024;
+    /// Maximum block period for a grace_period.
+    /// The grace_period is a delay between the point where the market closes and the point where the oracle may report.
+    pub const MaxGracePeriod: BlockNumber = MAX_GRACE_PERIOD;
+    pub const MaxMarketLifetimeInBlocks: BlockNumber = MAX_MARKET_LIFETIME_IN_BLOCKS;
+    pub const MaxMarketLifetimeInMoments: Moment = MAX_MARKET_LIFETIME_IN_MOMENTS;
```
sea212, line 168: Docstring missing.

runtime/zeitgeist/src/parameters.rs
```diff
@@ -155,31 +155,35 @@ parameter_types! {
...
-    // 60_000 = 1 minute. Should be raised to something more reasonable in the future.
-    /// Minimum number of milliseconds a Rikiddo market must be in subsidy gathering phase.
-    pub const MinSubsidyPeriod: Moment = 60_000;
+    /// Maximum string length for edit reason.
+    pub const MaxEditReasonLen: u32 = 1024;
+    /// Maximum block period for a grace_period.
+    /// The grace_period is a delay between the point where the market closes and the point where the oracle may report.
+    pub const MaxGracePeriod: BlockNumber = MAX_GRACE_PERIOD;
+    pub const MaxMarketLifetimeInBlocks: BlockNumber = MAX_MARKET_LIFETIME_IN_BLOCKS;
+    pub const MaxMarketLifetimeInMoments: Moment = MAX_MARKET_LIFETIME_IN_MOMENTS;
```
sea212, line 168: Missing Docstrings.

sea212 (commented):
LGTM. Tests cover the cases and succeed. Just some docstrings missing.

runtime/battery-station/src/parameters.rs
maltekliemann: Why are we double-booking the docs of the pallet config parameters for every runtime that we maintain? I wouldn't do that for reasons of abstractions. Right now, when the doc changes, we need to change it in three different places.

maltekliemann: Side-by-side comparison with and without docstrings. Which one's easier to read, edit and alphabetize? And let's be real here for a second: Does `MaxMarketPeriod` really need the docstring "The maximum market period"?

Screenshot 2022-12-22 at 11 57 54

sea212: > Why are we double-booking the docs of the pallet config parameters for every runtime that we maintain? I wouldn't do that for reasons of abstractions. Right now, when the doc changes, we need to change it in three different places.

We've done that because those are three separate crates that deserve their own documentation and most of the parameters are present in the runtime crates, but not in the primitives crate. When you inspect the documentation of each of the crates, you can see an appropriate documentation. 

<details>
  <summary>
  Battery Station Documentation
  </summary>
  </img>
</details


If we remove them from the runtime crates, most of the types will be undocumented. For the others that are present in the primitives crate and documented: How should one know to look there? I think for us this is obvious, for new developers it probably won't be so straightforward to figure out what these values are for.

I see the problem of having to update the docs in three places if they change.
I am not really open to remove all parameter docs from the runtime crates, because most of the documentation is not available in the parameters crate, however I would be open to leave the docs away that are present in primitives and add a top-level docstring to the parameters module in the runtime crates that references to the constant module in the primitives crate for the missing documentation.

sea212: > Side-by-side comparison with and without docstrings. Which one's easier to read, edit and alphabetize? And let's be real here for a second: Does `MaxMarketPeriod` really need the docstring "The maximum market period"?
> 
> Screenshot 2022-12-22 at 11 57 54

I think this is not an issue of the docstrings (in fact that help devs to know what it's about), but of the syntax highlighting. In the right example we see the docstrings being highlighted in a way that puts them in the foreground. At the same time the the constants are also highlighted in a way that puts them in the foreground (just not as much as the docstrings, so the docstrings are visually more conspicuous). I bet that if the docstrings in the right example would be highlighted like the comments in the left example (gray, which renders them inconspicuous relative to the dark gray background), they would not bother us in the right example and the actual code would be perceived more continuous.

In regards to `MaxMarketPeriod`, disregarding actual implementation details, I agree that the names are self-explanatory. I am currently questioning if we should include documentation about implementation details, for example: When exactly is the market period counted? Is it from market creation to resolution? Market start to market close? Market creation to market close? It would help a bit when setting this value, however, it would introduce double booking again: If the implementation details in regards to `MaxMarketPeriod` in PM change, we also have to update the docs in runtime. This is error-prone. What is your opinion about this?
I remember when creating the [XCM configuration and structures](https://github.com/zeitgeistpm/zeitgeist/blob/75bcae6e21c15b6fbf91e3f8aee76adfe7f6bf9d/runtime/zeitgeist/src/xcm_config/config.rs) needed for that, I decided to go with docstrings that include implementation details when useful in this context because there are a lot of different components that are dispersed all over different crates and it takes quite some time to figure out what is going on. That way we have it all in one place.

maltekliemann: > I think this is not an issue of the docstrings (in fact that help devs to know what it's about), but of the syntax highlighting. In the right example we see the docstrings being highlighted in a way that puts them in the foreground. At the same time the the constants are also highlighted in a way that puts them in the foreground (just not as much as the docstrings, so the docstrings are visually more conspicuous). I bet that if the docstrings in the right example would be highlighted like the comments in the left example (gray, which renders them inconspicuous relative to the dark gray background), they would not bother us in the right example and the actual code would be perceived more continuous.

That's true, grey would work much better here.

> I see the problem of having to update the docs in three places if they change.
> I am not really open to remove all parameter docs from the runtime crates, because most of the documentation is not available in the parameters crate, however I would be open to leave the docs away that are present in primitives and add a top-level docstring to the parameters module in the runtime crates that references to the constant module in the primitives crate for the missing documentation.

I understand. I'm okay with that. As long as it's not just a copy-paste situation, I'm okay with doubling the docs. We're not changing them often.

> If the implementation details in regards to MaxMarketPeriod in PM change, we also have to update the docs in runtime. This is error-prone. What is your opinion about this?

I think some terms (like _market lifetime_. and everything assocaited with the lifecycle of a markete) should rather be defined in a "meta documentation". Otherwise you end up not only double-booking between the runtimes and the pallets, but also between different objects in `zrml_prediction_markets`. For example, the market life cycle needs to be explained in one place. Otherwise, the description of the market cycle is split between `ReportingPeriod`, `MinOracleDuration`, `MaxOracleDuration`, etc. Once you understand how the state transitions work, you don't require long explanations in the docstrings. The documentation page is a good place for that, I think. Not sure if that really answer the question, though.

primitives/src/constants.rs
maltekliemann: [3ff0b04](https://github.com/zeitgeistpm/zeitgeist/pull/908/commits/3ff0b04284481e8d9b2cff72f695e326adcd0434)

runtime/zeitgeist/src/parameters.rs
maltekliemann: [85cf8ce](https://github.com/zeitgeistpm/zeitgeist/pull/908/commits/85cf8ce82c98c06632adbdb0474bd1640ddd9b15)

runtime/battery-station/src/parameters.rs
maltekliemann: [85cf8ce](https://github.com/zeitgeistpm/zeitgeist/pull/908/commits/85cf8ce82c98c06632adbdb0474bd1640ddd9b15)

sea212 (dismissed):

sea212 (approved):

zrml/prediction-markets/src/lib.rs
```diff
@@ -2182,27 +2187,26 @@ mod pallet {
...
+                    let now = <zrml_market_commons::Pallet<T>>::now();
+                    let now_frame = Self::calculate_time_frame_of_moment(now);
                     let end_frame = Self::calculate_time_frame_of_moment(range.end);
                     ensure!(now_frame < end_frame, Error::<T>::InvalidMarketPeriod);
                     ensure!(range.start < range.end, Error::<T>::InvalidMarketPeriod);
+                    let lifetime = range.end.saturating_sub(now); // Never saturates!
                     ensure!(
-                        range.end <= T::MaxMarketPeriod::get().saturated_into::<MomentOf<T>>(),
-                        Error::<T>::InvalidMarketPeriod
+                        lifetime <= T::MaxMarketLifetimeInMoments::get(),
```
vivekvpandya, line 2208: Question: so convert moment to blocks (or Timeframe) to check that it lasts at least one block? So can we not just have MaxMarketLifeInBlocks and compute lifetime in terms of Blocks (converted from moment by using calculate_time_frame_of_moment() and there for we can have only single config constant required.

vivekvpandya:
Apart from https://github.com/zeitgeistpm/zeitgeist/pull/908#discussion_r1062198757 it looks good to me.

zrml/prediction-markets/src/lib.rs
maltekliemann: Hmm... good question. I don't think I have a good reason for that. Is there any advantage to having two config parameters?

vivekvpandya: if we need different values for those two then its required to have to parameters. Otherwise I don't see any benefit.

vivekvpandya (approved):
Thanks for accepting suggestion! Looks good!

sea212 (approved):

PR #931, opened 2023-01-06
Basic tag.yml script

vivekvpandya:


vivekvpandya:
fixes #926

vivekvpandya:
Proof that this action works https://github.com/zeitgeistpm/zeitgeist/actions/runs/3853531292

.github/workflows/tag.yml
```diff
@@ -0,0 +1,23 @@
+name: Try Runtime on New Tags
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
Chralt98, line 6: https://stackoverflow.com/a/71076957

> This means that the workflow will only be triggered when a tag is pushed with a name that match the pattern of "v" followed by one or more digits, a dot, one or more digits, another dot and one or more digits. This pattern corresponds to a semantic versioning format, like v1.2.3. This way, this configuration is useful if you want to trigger actions only on specific versions of your software, like publishing to a package manager, deploying or releasing an specific version to a certain environment.

*Generated by ChatGPT

OK

.github/workflows/tag.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      # TODO(#839): `rustup show` installs the rustc version specified in the `rust-toolchain` file
+      # according to https://rust-lang.github.io/rustup/overrides.html. We don't use actions-rs due
+      # to the lack of support of `rust-toolchain` files with TOML syntax:
+      # https://github.com/actions-rs/toolchain/issues/126.
+      - name: Install rust toolchain
+        run: rustup show
+
+      - run: make try-runtime-upgrade-battery-station 
+      - run: make try-runtime-upgrade-zeitgeist
```
Chralt98, line 23: We could run them in parallel with two jobs `try_runtime_battery_station`, `try_runtime_zeitgeist`.

Chralt98: And use the [swatinem rust cache](https://github.com/Swatinem/rust-cache) to load common dependencies not again.

https://github.com/zeitgeistpm/zeitgeist/blob/e991d9edbc76db6f2ecc2780a7caf4a4702cff6a/.github/workflows/rust.yml#L31-L32

> The rust-cache action should be placed before the cargo build and cargo test steps because it is used to restore the dependencies of the project from the cache. By restoring the dependencies before building and testing the project, you can ensure that the dependencies are up-to-date and that the build and test steps are not delayed by the time it takes to download and compile the dependencies.

*Generated by ChatGPT

.github/workflows/tag.yml
```diff
@@ -3,21 +3,25 @@ name: Try Runtime on New Tags
 on:
   push:
     tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - '*'
+      # - 'v[0-9]+.[0-9]+.[0-9]+'
```
Chralt98, line 7: I think `- 'v[0-9]+.[0-9]+.[0-9]+'` is great. Just for testing the CI you use `*`, right?

Chralt98 (approved):
Looks beautiful!

PR #937, opened 2023-01-09
Fix order of arguments in `get_spot_price`

maltekliemann:
The problem is that polkadot-js doesn't play well with optional parameters followed by non-optional parameters. In fact, the Rpc calls cannot be used with polkadot-js (they do still work when using websocat).

Anyways, reordering the parameters here to ensure that they can be used by the SDK.


vivekvpandya (approved):
Looks good! Is there similar issue with other pallets?

maltekliemann:
> [...] Is there similar issue with other pallets?

Don't think so.

sea212 (approved):

PR #943, opened 2023-01-12
[PM] Set market end when admin closes the market

### LINKED ISSUE #942: After using adminMoveMarketToClosed still need to wait for end
When using adminMoveMarketToClosed and then trying to report the outcome, extrisic fails with the message shown in the screenshot. The market has gracePeriod set to 0.

image

image

It would be perfect if the `range.end` in the storage is changed as well to the moment this extrinsic executed. We might need this for automated tests eventually, but for now it is fine if it stays the same so manual testing can be quickly done.

Chralt98:
Fixes #942

This is related to @TvrtkoM 's [comment on Discord](https://discord.com/channels/737780518313000960/817041223201587230/1062655318363996192).

> Really good point. The error is based on the actual market period end, which was put in by the market creator. Because you used adminMoveMarketToClosed this end was not edited to be at the time of the call to adminMoveMarketToClosed. To be able to report, you currently need to wait until the real market end time is reached.

> The question, when solving the problem is: Should we modify the MarketPeriod(range.end) to be at the time the market close was set by the admin or should we rather leave it unmodified and add another storage field.

I decided to modify the end, because I see no negative consequences at the moment.

This allows to [report the market directly](https://github.com/zeitgeistpm/zeitgeist/blob/e991d9edbc76db6f2ecc2780a7caf4a4702cff6a/zrml/prediction-markets/src/lib.rs#L1169-L1174) after closing it as an admin.

### DIFF
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -326,6 +326,7 @@ mod pallet {
             let open_ids_len = Self::clear_auto_open(&market_id)?;
             let close_ids_len = Self::clear_auto_close(&market_id)?;
             Self::close_market(&market_id)?;
+            Self::set_market_end(&market_id)?;
             // The CloseOrigin should not pay fees for providing this service
             Ok((
                 Some(T::WeightInfo::admin_move_market_to_closed(open_ids_len, close_ids_len)),
@@ -2301,6 +2302,25 @@ mod pallet {
             Self::deposit_event(Event::MarketClosed(*market_id));
             total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
             Ok(total_weight)
+        }
+
+        pub(crate) fn set_market_end(market_id: &MarketIdOf<T>) -> Result<Weight, DispatchError> {
+            <zrml_market_commons::Pallet<T>>::mutate_market(market_id, |market| {
+                market.period = match market.period {
+                    MarketPeriod::Block(ref range) => {
+                        let current_block = <frame_system::Pallet<T>>::block_number();
+                        MarketPeriod::Block(range.start..current_block)
+                    }
+                    MarketPeriod::Timestamp(ref range) => {
+                        let current_time_frame = Self::calculate_time_frame_of_moment(
+                            <zrml_market_commons::Pallet<T>>::now(),
+                        );
+                        MarketPeriod::Timestamp(range.start..current_time_frame)
+                    }
+                };
+                Ok(())
+            })?;
+            Ok(T::DbWeight::get().reads_writes(1, 1))
         }
 
         /// Handle market state transitions at the end of its active phase.
--- a/zrml/prediction-markets/src/lib.rs
+++ b/zrml/prediction-markets/src/lib.rs
@@ -2354,10 +2354,8 @@ mod pallet {
                         MarketPeriod::Block(range.start..current_block)
                     }
                     MarketPeriod::Timestamp(ref range) => {
-                        let current_time_frame = Self::calculate_time_frame_of_moment(
-                            <zrml_market_commons::Pallet<T>>::now(),
-                        );
-                        MarketPeriod::Timestamp(range.start..current_time_frame)
+                        let now = <zrml_market_commons::Pallet<T>>::now();
+                        MarketPeriod::Timestamp(range.start..now)
                     }
                 };
                 Ok(())
--- a/zrml/prediction-markets/src/tests.rs
+++ b/zrml/prediction-markets/src/tests.rs
@@ -99,14 +99,20 @@ fn simple_create_scalar_market(
 }
 
 #[test]
-fn admin_move_market_to_closed_successfully_closes_market() {
+fn admin_move_market_to_closed_successfully_closes_market_and_sets_end() {
     ExtBuilder::default().build().execute_with(|| {
-        frame_system::Pallet::<Runtime>::set_block_number(1);
-        simple_create_categorical_market(MarketCreation::Permissionless, 0..2, ScoringRule::CPMM);
+        run_blocks(7);
+        let now = frame_system::Pallet::<Runtime>::block_number();
+        let end = 42;
+        simple_create_categorical_market(MarketCreation::Permissionless, now..end, ScoringRule::CPMM);
+        run_blocks(3);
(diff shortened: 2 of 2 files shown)


zrml/prediction-markets/src/lib.rs
```diff
@@ -2345,6 +2346,25 @@ mod pallet {
...
+                market.period = match market.period {
+                    MarketPeriod::Block(ref range) => {
+                        let current_block = <frame_system::Pallet<T>>::block_number();
+                        MarketPeriod::Block(range.start..current_block)
+                    }
+                    MarketPeriod::Timestamp(ref range) => {
+                        let current_time_frame = Self::calculate_time_frame_of_moment(
+                            <zrml_market_commons::Pallet<T>>::now(),
+                        );
+                        MarketPeriod::Timestamp(range.start..current_time_frame)
```
maltekliemann, line 2360: The `Timestamp` variant of `MarketPeriod` actually holds the UNIX timestamp (in ms), not the time frame.

```suggestion
                        let now = <zrml_market_commons::Pallet<T>>::now();
                        MarketPeriod::Timestamp(range.start..now)
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -104,19 +104,25 @@ fn simple_create_scalar_market(
...
-            0..2,
+            now..end,
             ScoringRule::CPMM,
         );
+        run_blocks(3);
         let market_id = 0;
         assert_ok!(PredictionMarkets::admin_move_market_to_closed(Origin::signed(SUDO), market_id));
         let market = MarketCommons::market(&market_id).unwrap();
         assert_eq!(market.status, MarketStatus::Closed);
+        let new_end = now + 3;
```
maltekliemann, line 123: I think this should be (probably missing a cast here :wink:):

```suggestion
        let new_end = now + 3 * MILLISECS_PER_BLOCK;
```

maltekliemann (changes requested):
I think this solution is fine. It messes a bit with the history of the market, but these changes can be traced back through an indexer. Plus we're only using these functions on the testnet.

maltekliemann:
Actually, this solution is perfectly idiomatic. It just feels confusing because we don't use a cache to determine when we're allowed to report.

zrml/prediction-markets/src/tests.rs
Chralt98: We are using `MarketPeriod::Block` here. If we had used `MarketPeriod::Timestamp` here I would totally agree.

Chralt98:
>  we don't use a cache to determine when we're allowed to report.

Yeah, we currently only cache when the market closes and derive that the report does go after it. But because I don't see any negative consequences to this solution, I would go with it. Do you see any negative consequences?

zrml/prediction-markets/src/tests.rs
maltekliemann: Oh, my bad.

maltekliemann: So there's no test for this when using timestamps? If not, could you add one? Seems important.

maltekliemann:
> > we don't use a cache to determine when we're allowed to report.
> 
> Yeah, we currently only cache when the market closes and derive that the report does go after it. But because I don't see any negative consequences to this solution, I would go with it. Do you see any negative consequences?

No, I think this is absolutely the correct solution.

zrml/prediction-markets/src/tests.rs
Chralt98: ad3f275465481a1d7bd95990f883c5b550bac3d9

maltekliemann (approved):
There is one concern here, which is why these functions should absolutely not be used for anything but testing. Since the grace period is specified in terms of _duration_ rather than a timepoint, the grace period of a market that is closed with `admin_move_market_to_closed` will end prematurely. Not a problem, unless you're dealing with a real market.

https://github.com/zeitgeistpm/zeitgeist/blob/ad3f275465481a1d7bd95990f883c5b550bac3d9/zrml/prediction-markets/src/lib.rs#L1198-L1200

As for the question about possible side effects. The thing we need to look out for is that it doesn't break the `market_status_manager`. Closing the market forcefully clears the auto-close cached in `MarketIdsPerClose*`, and even if it didn't that wouldn't be a problem because `on_market_close` takes care not to mess with a market that is in an unexpected state.

sea212 (commented):
PR LGTM.

In regards to https://github.com/zeitgeistpm/zeitgeist/pull/943#pullrequestreview-1269136107
> There is one concern here, which is why these functions should absolutely not be used for anything but testing. Since the grace period is specified in terms of duration rather than a timepoint, the grace period of a market that is closed with admin_move_market_to_closed will end prematurely. Not a problem, unless you're dealing with a real market.

I think we should handle every function in a pallet as if it was used in a production environment, unless it is clearly only enabled in specific environments such as `test`.
How should we differentiate between the different environments they are deployed in, for example if it is in BS or Zeitgeist? I think that what you Malte meant with "real" markets, the ones on the mainnet are real and the ones on BS are not, right? Do you think there is a way to adjust the grace period properly?

Besides that, those admin functions are not used on the mainnet, are they? In that case, would somebody please open an issue to filter them there?

maltekliemann:
> I think we should handle every function in a pallet as if it was used in a production environment, unless it is clearly only enabled in specific environments such as `test`. How should we differentiate between the different environments they are deployed in, for example if it is in BS or Zeitgeist? I think that what you Malte meant with "real" markets, the ones on the mainnet are real and the ones on BS are not, right? Do you think there is a way to adjust the grace period properly?

Just to clarify, by "testing purposes" I mean using these function on Battery Station for playing with markets. Yes, we can easily adjust the grace period. The only question is if that's "correct". For example, if I want to test the grace period works correctly on Battery Station, it would be more natural if its length would remain unchanged (instead of its end)?

> Besides that, those admin functions are not used on the mainnet, are they? In that case, would somebody please open an issue to filter them there?

They've never been used on the mainnet and I would strongly suggest to filter them in the future.

sea212 (approved):

sea212:
As a side note, when developing pallets always assume they are used by other projects. If there are some side effects during a function invocation, we'll have to document it. We should also mark when a function is not adequate for a production environment.

Issue #924, opened 2023-01-04
Clients should be automatically upgraded on the servers


sea212:

sea212:
Solved by https://github.com/zeitgeistpm/zeitgeist/pull/944

PR #944, opened 2023-01-12
Zeitgeist Client Update Automation

samuelarogbonlo:
This should resolve (#924). 

This PR contains the automation script to auto-update the zeitgeist clients on the self-hosted servers and OnFinality. @sea212 Can you possibly take a look at this whenever you can? Thanks.


scripts/ansible/client-auto-update-playbook.yml
```diff
@@ -0,0 +1,65 @@
+- hosts: zeitgeist
+  gather_facts: no
+  become: true 
+  environment: 
+    RELEASE_VERSION: "{{ lookup('env', 'RELEASE_VERSION') }}"
+
+  tasks:
+    - name: Get Zeitgeist services
+      no_log: true
```
Chralt98, line 9: Why did you decided to use `no_log: true` in all cases? This would make it harder to debug.

scripts/ansible/client-auto-update-playbook.yml
```diff
@@ -0,0 +1,65 @@
+- hosts: zeitgeist
+  gather_facts: no
+  become: true 
+  environment: 
+    RELEASE_VERSION: "{{ lookup('env', 'RELEASE_VERSION') }}"
```
Chralt98, line 5: Where do we set `RELEASE_VERSION`? In the environment variables? What happens, if it is not set? Can we print the RELEASE_VERSION in the Github Actions Debug Console?

Chralt98 (commented):
Impressed by ChatGPTs response:

> The most important mistake to fix would likely be the lack of error handling and validation. This script does not check for errors that might occur during the download or execution of the software, which can make it difficult to troubleshoot issues if they arise. Additionally, it does not validate if the variable RELEASE_VERSION is set or not, which can cause issues if not set. This can result in the script trying to download the wrong version of the software or trying to download a version that does not exist.

> Another important mistake would be the lack of compatibility and security checks. The script does not check if the new version of Zeitgeist software is compatible with the current system, or if the new version of the software is signed and verified or not. This could potentially lead to issues if the software is not compatible or if the new version is not verified and could be a security risk.

> Lastly, the script does not check if the new version of Zeitgeist software is already installed or not, and does not check if the new version of Zeitgeist software is tested before applying the update. This could lead to unnecessary updates and could cause issues if the new version is not fully tested.

scripts/ansible/client-auto-update-playbook.yml
samuelarogbonlo: I added the `no_log` to prevent any of the server IP addresses from being outputted when the task runs. The Zeitgeist repository is public, I believe anyone can look through the actions and see the output from each workflow ran.

scripts/ansible/client-auto-update-playbook.yml
```diff
@@ -0,0 +1,73 @@
...
+        mv zeitgeist zeitgeist_$PREV_VERSION.bak
+      when: mnt.stat.exists and mnt.stat.isdir
+        
+    - name: Create Backup from Previous Zeitgeist Client
+      no_log: true
+      shell: |
+        cd /services/zeitgeist/bin
+        PREV_VERSION=$(./zeitgeist --version | grep -Eo '([0-9]{1,}\.)+[0-9]{1,}' | tr -d '.')
+        mv zeitgeist zeitgeist_$PREV_VERSION.bak
+      when: not mnt.stat.exists and mnt.stat.isdir
```
sea212, line 40: We might want to remove old backups here to avoid bloating.

scripts/ansible/client-auto-update-playbook.yml
```diff
@@ -0,0 +1,73 @@
...
+        path: /mnt/
+      register: mnt
+
+    - name: Create Backup from Previous Zeitgeist Client in Mount Dir
+      no_log: true
+      shell: |
+        cd /mnt/*/services/zeitgeist/bin
+        PREV_VERSION=$(./zeitgeist --version | grep -Eo '([0-9]{1,}\.)+[0-9]{1,}' | tr -d '.')
+        mv zeitgeist zeitgeist_$PREV_VERSION.bak
+      when: mnt.stat.exists and mnt.stat.isdir
```
sea212, line 32: We might want to remove old backups here to avoid bloating.

sea212 (changes requested):
Main needs to be merged into this branch.
How can we test this before merging? Maybe by adding manual dispatch and executing it.

scripts/ansible/client-auto-update-playbook.yml
```diff
@@ -0,0 +1,73 @@
...
+      shell: |
+        cd /services/zeitgeist/bin
+        PREV_VERSION=$(./zeitgeist --version | grep -Eo '([0-9]{1,}\.)+[0-9]{1,}' | tr -d '.')
+        mv zeitgeist zeitgeist_$PREV_VERSION.bak
+      when: not mnt.stat.exists and mnt.stat.isdir
+
+    - name: Download Zeitgeist Client to Mount Dir 
+      no_log: true
+      shell: |
+        cd /mnt/*/services/zeitgeist/bin
```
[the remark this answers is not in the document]
samuelarogbonlo: Yes for now there's no more than one sub-directory in the /mnt/ directory. These sub-directories contain the parachain and relay-chain data mounted in volumes for each zeitgeist node.

sea212:
Blocked: Waiting for secrets to be added to GH

sea212 (approved):

PR #951, opened 2023-01-20
Use tag when publishing a parachain docker image

### LINKED ISSUE #548: [CICD] Tag parachain docker images properly on DockerHub
The actual tag that a parachain node docker image is build upon is not to [DockerHub](https://hub.docker.com/r/zeitgeistpm/zeitgeist-node-parachain). This makes it difficult to select an image based on a specific zeitgeist node version.

sea212:
Closes #548

### DIFF
--- a/.github/workflows/docker-hub-parachain.yml
+++ b/.github/workflows/docker-hub-parachain.yml
@@ -23,7 +23,7 @@ jobs:
           images: |
             zeitgeistpm/zeitgeist-node-parachain
           tags: |
-            type=sha
+            type=semver,pattern={{version}}
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1


.github/workflows/docker-hub-parachain.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           images: |
             zeitgeistpm/zeitgeist-node-parachain
           tags: |
-            type=sha
+            type=semver,pattern={{version}}
```
Chralt98, line 26: > This line of code is saying that the tags for the Docker images being built are going to be in the format of a semantic version (i.e. 1.0.0, 2.3.1, etc.) and the specific version number will be determined by the {{version}} variable.

*ChatGPT

Instead of a `sha` hash, we use a semantic version for the docker image.

Chralt98 (approved):

.github/workflows/docker-hub-parachain.yml
vivekvpandya: where does this version variable set?

sea212: @vivekvpandya The GitHub Action [metadata-action](https://github.com/docker/metadata-action#typesemver) does set it.

PR #953, opened 2023-01-21
Update weights

sea212:
Updates the weights that resulted from Zeitgeist benchmark reference machine.
This time the delta is quite enormous, I am wondering why that is the case. The benchmark script was executed as always, the server is the same and the server did not carry out any other significant tasks during benchmarking.


Chralt98 (approved):
I would bet the underlying reason for the significant higher weights are something internal by the [Polkadot upgrade](https://github.com/zeitgeistpm/zeitgeist/commit/3e07be62131d7fcbca0bdb2fccc8749a67aa3cc9).

Chralt98:
No it's actually that Vivek has run the benchmark script on his machine. :-D

vivekvpandya (approved):

PR #957, opened 2023-01-24
Include integration tests in runtime only if `test` feature is set

sea212:
Integration tests are also included in Battery Station Runtime if only `parachain` feature is set. They should only be included if `test` feature is set.

### DIFF
--- a/runtime/battery-station/src/lib.rs
+++ b/runtime/battery-station/src/lib.rs
@@ -78,7 +78,6 @@ use sp_runtime::{
 use nimbus_primitives::CanAuthor;
 use sp_version::RuntimeVersion;
 
-#[cfg(feature = "parachain")]
 #[cfg(test)]
 pub mod integration_tests;
 #[cfg(feature = "parachain")]


runtime/battery-station/src/lib.rs
```diff
@@ -78,7 +78,6 @@ use sp_runtime::{
 use nimbus_primitives::CanAuthor;
 use sp_version::RuntimeVersion;
 
-#[cfg(feature = "parachain")]
```
Chralt98, line 81: https://github.com/zeitgeistpm/zeitgeist/blob/54099511b3819507fd0306db98fc7b96b1240b79/runtime/battery-station/src/integration_tests/xcm/mod.rs#L17

Ok, so the parachain feature is only required for the xcm crate. But we could have other modules inside `integration_tests` which should be able to run in standalone mode.

Chralt98 (approved):

runtime/battery-station/src/lib.rs
sea212: Yes, in addition the logic applied here was plainly wrong. It would use the module when either `parachain`, `test` or both are set. The correct way to express that both feature have to be set would be like in the code you referenced.

vivekvpandya (approved):

maltekliemann (approved):

PR #960, opened 2023-01-30
Gensis: User proper inflation config and eligiblity ratio

sea212:
Applies proper configuration for `parachain-staking` at genesis.
The inflation config initialization is adjusted in respect to [one pr from moonbeam](https://github.com/PureStake/moonbeam/pull/232) and [the latest state of the code](https://github.com/PureStake/moonbeam/blob/master/node/service/src/chain_spec/moonbeam.rs#L190-L217). The new code properly derives per-round inflation based on annual inflation **and expected staking amounts**. The total inflation based on expected staking amounts now looks like that:
1. If 5% or less of the total supply are staked, the inflation caused by collation and delegation rewards is 2%
2. Between 5% and 15%, the inflation [...] is 3.5%
3. For more than 15%, the inflation [...] is 5%

The changes allow allow to configure different `InflationConfig` for each network. This won't affect the running networks and can be readjusted once a new network is about to be spawned.

Besides that, the `EligibilityRatio` is set to `1`, meaning that only one collator is selected per block (unless he timeouts). Anything else can lead to network congestion.

### DIFF
--- a/node/src/chain_spec/dev.rs
+++ b/node/src/chain_spec/dev.rs
@@ -36,6 +36,7 @@ use zeitgeist_primitives::{
 #[cfg(feature = "parachain")]
 use {
     super::battery_station::inflation_config,
+    sp_runtime::Perbill,
     zeitgeist_primitives::constants::{ztg::TOTAL_INITIAL_ZTG, BASE},
 };
 
@@ -80,7 +81,9 @@ pub fn dev_config() -> Result<BatteryStationChainSpec, String> {
                     )],
                     collator_commission: DefaultCollatorCommission::get(),
                     inflation_info: inflation_config(
-                        sp_runtime::Perbill::from_percent(5),
+                        Perbill::from_parts(20),
+                        Perbill::from_parts(35),
+                        Perbill::from_parts(50),
                         TOTAL_INITIAL_ZTG * BASE,
                     ),
                     nominations: vec![],
--- a/node/src/chain_spec/dev.rs
+++ b/node/src/chain_spec/dev.rs
@@ -30,7 +30,7 @@ use battery_station_runtime::{
 use sc_service::ChainType;
 use sp_core::sr25519;
 use zeitgeist_primitives::{
-    constants::ztg::{LIQUIDITY_MINING, LIQUIDITY_MINING_PTD},
+    constants::ztg::{LIQUIDITY_MINING, LIQUIDITY_MINING_PTD, STAKING_PTD},
     types::Balance,
 };
 #[cfg(feature = "parachain")]
@@ -81,9 +81,9 @@ pub fn dev_config() -> Result<BatteryStationChainSpec, String> {
                     )],
                     collator_commission: DefaultCollatorCommission::get(),
                     inflation_info: inflation_config(
-                        Perbill::from_perthousand(20),
-                        Perbill::from_perthousand(35),
-                        Perbill::from_perthousand(50),
+                        STAKING_PTD * Perbill::from_percent(40),
+                        STAKING_PTD * Perbill::from_percent(70),
+                        STAKING_PTD,
                         TOTAL_INITIAL_ZTG * BASE,
                     ),
                     nominations: vec![],
--- a/node/src/chain_spec/dev.rs
+++ b/node/src/chain_spec/dev.rs
@@ -81,9 +81,9 @@ pub fn dev_config() -> Result<BatteryStationChainSpec, String> {
                     )],
                     collator_commission: DefaultCollatorCommission::get(),
                     inflation_info: inflation_config(
-                        Perbill::from_parts(20),
-                        Perbill::from_parts(35),
-                        Perbill::from_parts(50),
+                        Perbill::from_perthousand(20),
+                        Perbill::from_perthousand(35),
+                        Perbill::from_perthousand(50),
                         TOTAL_INITIAL_ZTG * BASE,
                     ),
                     nominations: vec![],
--- a/node/src/chain_spec/dev.rs
+++ b/node/src/chain_spec/dev.rs
@@ -22,7 +22,6 @@ use super::{
     get_account_id_from_seed, get_from_seed, token_properties, AdditionalChainSpec,
     EndowedAccountWithBalance,
 };
-
 #[cfg(feature = "parachain")]
 use battery_station_runtime::{
     DefaultBlocksPerRound, DefaultCollatorCommission, DefaultParachainBondReservePercent,
@@ -33,6 +32,11 @@ use sp_core::sr25519;
 use zeitgeist_primitives::{
     constants::ztg::{LIQUIDITY_MINING, LIQUIDITY_MINING_PTD},
     types::Balance,
+};
+#[cfg(feature = "parachain")]
+use {
+    super::battery_station::inflation_config,
+    zeitgeist_primitives::constants::{ztg::TOTAL_INITIAL_ZTG, BASE},
 };
 
 const INITIAL_BALANCE: Balance = Balance::MAX >> 4;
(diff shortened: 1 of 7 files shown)


primitives/src/constants/ztg.rs
```diff
@@ -46,10 +46,19 @@ pub const TEAM_AND_ADVISORS: u128 = 15_000_000;
...
+    + SEED_SALE
+    + STRATEGIC_SALE
+    + TEAM_AND_ADVISORS
+    + ZEITGEIST_FOUNDATION;
+
 // Inflation
 
 /// Perthousand liquidity mining inflation. 2%
-pub const LIQUIDITY_MINING_PTD: Perbill = Perbill::from_perthousand(20);
+pub const LIQUIDITY_MINING_PTD: Perbill = Perbill::from_perthousand(0);
```
vivekvpandya, line 61: Comment above seems to need an update.

primitives/src/constants/ztg.rs
```diff
@@ -46,10 +46,19 @@ pub const TEAM_AND_ADVISORS: u128 = 15_000_000;
...
 
 /// Perthousand collator staking inflation. 1.5%
-pub const STAKING_PTD: Perbill = Perbill::from_perthousand(15);
+pub const STAKING_PTD: Perbill = Perbill::from_perthousand(50);
```
vivekvpandya, line 64: Comment above seems to need an update.

primitives/src/constants.rs
```diff
@@ -29,6 +29,7 @@ use crate::types::{Balance, BlockNumber};
 use frame_support::{parameter_types, PalletId};
 
 // Definitions for time
+pub const BLOCKS_PER_YEAR: BlockNumber = (BLOCKS_PER_DAY * 36525) / 100;
```
vivekvpandya, line 32: Not able to follow the math here, shouldn't it be just BLOCKS_PER_DAY * 365 ? Or there is something I don't see.

node/src/chain_spec/dev.rs
```diff
@@ -75,7 +80,12 @@ pub fn dev_config() -> Result<BatteryStationChainSpec, String> {
                         super::battery_station::DEFAULT_STAKING_AMOUNT_BATTERY_STATION,
                     )],
                     collator_commission: DefaultCollatorCommission::get(),
-                    inflation_info: crate::chain_spec::DEFAULT_COLLATOR_INFLATION_INFO,
+                    inflation_info: inflation_config(
+                        Perbill::from_parts(20),
```
vivekvpandya, line 84: Don't we need from_percent() here?

node/src/chain_spec/mod.rs
```diff
@@ -42,46 +42,53 @@ use zeitgeist_primitives::{
...
-        pub(crate) const DEFAULT_COLLATOR_INFLATION_INFO: pallet_parachain_staking::InflationInfo<Balance> = {
-            let hours_per_year = 8766;
-            let millisecs_per_year = hours_per_year * 60 * 60 * 1000;
-            let round_millisecs = DefaultBlocksPerRound::get() as u64 * MILLISECS_PER_BLOCK as u64;
-            let rounds_per_year = millisecs_per_year / round_millisecs;
-
-            let annual_inflation = ztg::STAKING_PTD;
-            let expected_annual_amount = ztg::COLLATORS * zeitgeist_primitives::constants::BASE;
-            let round_inflation_parts = annual_inflation.deconstruct() as u64 / rounds_per_year;
-            let round_inflation = Perbill::from_parts(round_inflation_parts as _);
```
vivekvpandya, line 64: I think here use of from_parts was correct as there was deconstruct() method call above.

vivekvpandya:
Apart from few questions rest looks good.

node/src/chain_spec/dev.rs
sea212: We could use `from_percent` for the case of event percentage, such as the highlighted `2%`. However, the other parts of the configuration also use one decimal fraction place, like `3.5%`, that's why we need `from_parts`, it maps `x / 1000`, such as `35 / 1000 = 3.5%`. To keep it uniform, I just used `from_parts` for any value in the configuration.

node/src/chain_spec/mod.rs
sea212: Yes, as far as I can tell the code derived correct per round inflation. However, it was inflexible because it did not allow to use different annual and per round inflation based on different expected staking values. In addition to that, every network had to use the same `InflationConfig` as defined in `mod.rs`. Now we can derive a function to create `InflationConfig` that respects the configuration of the runtime in which it is used.

node/src/chain_spec/dev.rs
sea212: How the heck did I assume `from_parts` is `x / 1_000` for `Perbill`? :laughing: It's `x / 1_000_000_000`.

https://github.com/zeitgeistpm/zeitgeist/pull/960/commits/ee29b50d6e7a5098e6a6b05bbe3ce1587429e27d, thanks Vivek!

vivekvpandya (dismissed):

primitives/src/constants/ztg.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/960/commits/f499f803652a4152aa3578e3316b53c1a54fa610

primitives/src/constants/ztg.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/960/commits/f499f803652a4152aa3578e3316b53c1a54fa610

sea212:
@Chralt98 seems like our [CodeCov token is expired](https://github.com/zeitgeistpm/zeitgeist/actions/runs/4053278452/jobs/6973711172). Can you add a new one?

Chralt98:
I never added a CodeCov token. According to this https://docs.codecov.com/docs/frequently-asked-questions#where-is-the-repository-upload-token-found the token should not be required for public repositories.

node/src/chain_spec/mod.rs
```diff
@@ -42,46 +42,53 @@ use zeitgeist_primitives::{
...
+                        min: annual_inflation_min,
+                        ideal: annual_inflation_ideal,
+                        max: annual_inflation_max,
+                    };
+                    pallet_parachain_staking::inflation::InflationInfo {
+                        // staking expectations
+                        expect: pallet_parachain_staking::inflation::Range {
+                            min: Perbill::from_percent(5).mul_floor(total_supply),
+                            ideal: Perbill::from_percent(10).mul_floor(total_supply),
+                            max: Perbill::from_percent(15).mul_floor(total_supply),
```
Chralt98, line 81: For the migration to Polkadot we also migrate our balances storage to the new runtime. This means we also "copy" the current `totalIssuance`.

We have a difference here for this, which is calculated with `total_supply - totalIssuance` of:

`1021933231948330118 - 1000000000000000000 = 21933231948330118` ZTG

This currently means, that there is an inflation difference. I know this can be changed by governance with [`set_inflation`](https://github.com/PureStake/moonbeam/blob/358538033db53aa8b0d38737eb560f6bb58c4390/pallets/parachain-staking/src/lib.rs#L777-L799) or [`set_staking_expectations`](https://github.com/PureStake/moonbeam/blob/358538033db53aa8b0d38737eb560f6bb58c4390/pallets/parachain-staking/src/lib.rs#L757-L776). So it should be not a big of deal, right?

primitives/src/constants/ztg.rs
```diff
@@ -46,10 +46,19 @@ pub const TEAM_AND_ADVISORS: u128 = 15_000_000;
...
 
-/// Perthousand liquidity mining inflation. 2%
-pub const LIQUIDITY_MINING_PTD: Perbill = Perbill::from_perthousand(20);
+/// Perthousand liquidity mining inflation. 0%
+pub const LIQUIDITY_MINING_PTD: Perbill = Perbill::from_perthousand(0);
 
-/// Perthousand collator staking inflation. 1.5%
-pub const STAKING_PTD: Perbill = Perbill::from_perthousand(15);
+/// Perthousand collator staking inflation. 5%
+pub const STAKING_PTD: Perbill = Perbill::from_perthousand(50);
```
Chralt98, line 64:
```suggestion
```

`STAKING_PTD` is unused.

Chralt98 (commented):
Ok, this seems like a useful genesis config for the parachain-staking pallet (especially if we want to migrate over to Polkadot).

Currently this is the staking rate on Zeitgeist:

`staking_rate = staking_amount / total_issuance`

- `staking_amount`: query from `parachainStaking -> total`
- `total_issuance`: query from `balances -> totalIssuance`

`staking_rate = 172185525432272240 / 1021933231948330118 approx. = 16,85%`

So with the current inflation plan, we would get an annual inflation of 5% currently.

Although I am personally not a big fan of inflating a currency at all. But it is what it is.

node/src/chain_spec/mod.rs
sea212: We will use the same inflation config on Polkadot as on Kusama. Those values can be used for next networks or can be adjusted if necessary.

vivekvpandya (approved):

PR #963, opened 2023-02-02
Add automatic copyright notice verification

### LINKED ISSUE #925: License headers should be automatically updated

maltekliemann:
This didn't turn out as good as I wanted it to and currently has a couple of limitations (see below). How to use this currently:

- Install `scripts/check-license` by running `pip install scripts/check-license` and `pip install -r scripts/check-license/requirements.txt`.
- You have to specify all files that you want to check. The workflow does this automatically for all `*.rs` files using https://github.com/Ana06/get-changed-files. This is a fork of https://github.com/jitterbit/get-changed-files which solves this issue and allows us to filter files by their extension: https://github.com/jitterbit/get-changed-files/issues/11. This won't work if any of our files have a whitespace in their name. So... one more reason not to do that, I guess.
- Run `check-license [files]`. You can also run `check-license -w [files]` to write the missing copyright notices.

Current limitations:

- We don't check for the presence of the GPL license. This can easily be implemented by checking if the `File` object's blob starts with the proper string. It might be helpful to remove the leading `"// "`.
- Different file types can be handled by extending the `File` class a little bit. We may want to do that in the future.
- The checker doesn't dig deep into the git log to verify if old copyrights are correct. Can be done, but totally not within the scope of this PR.

You can see the workflow fail due to a missing copyright in a Rust file here: https://github.com/zeitgeistpm/zeitgeist/actions/runs/4085940425/jobs/7044586744.

Blocked by #964, which needs to be merged before merging this PR to ensure that we're starting off with a correct state. Closes #925.


.github/workflows/rust.yml
```diff
@@ -34,6 +34,26 @@ jobs:
...
+          pip install scripts/check-license
+          pip install -r scripts/check-license/requirements.txt
+      - name: Query files changed
+        id: files_changed
+        uses: Ana06/get-changed-files@v1.2
+        with:
+          filter: '*.rs$'
+      - name: Check copyright notices
+        run: check-license ${{ steps.files_changed.outputs.added_modified }}
+
```
sea212, line 56: The script does offer the option to correct those errors, what's your position to automatically adding a commit with the updated license headers and optionally passing the workflow log in a comment in the PR?
- Add a commit: https://github.com/zeitgeistpm/zeitgeist/actions/runs/985953773/workflow#L138-L147
- Comment success/failure + logs: https://github.com/zeitgeistpm/zeitgeist/actions/runs/985953773/workflow#L173-L189

scripts/check-license/.gitignore
```diff
@@ -0,0 +1,129 @@
...
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
```
sea212, line 11: I see the point of adding the .gitignore in that folder and I think its sufficient for almost any case. However, maybe someone decides to initialize a pyenv within the root to cover anything from the root which will create a `.python-version` file in the root which is tracked by git. I think it would be better if we include either those rules in the repositories root `.gitignore`.

.github/workflows/rust.yml
sea212: Writing would just trigger in the case of a PR that should be merged into main.

sea212 (commented):
This is a partial review, I'll do the rest soon.

.github/workflows/rust.yml
maltekliemann: And the commit is triggered by placing a label?

maltekliemann: I'm okay with automating the process, but I have to caution everyone to double-check the diff everytime they do this. And just to prove my point, I botched the test for the `File.write` function because I forgot how `mock` works...

scripts/check-license/.gitignore
maltekliemann: 28e8f705fe5aed271ed2c49274b01cfae896d8e6

.github/workflows/rust.yml
maltekliemann: Maybe just a quick argument against automating the write: It probably takes just as much time to execute the write locally and then push the changes manually as it takes to trigger the write remotely, but its much safer to make the changes locally and run `git diff` once just to check if there's been a mistake.

sea212 (approved):
Didn't expect such a complex solution, beware copyright notices :smile:

PR #965, opened 2023-02-03
Fix name of `pallet_multisig`

maltekliemann:
Unfortunately, it's currently not possible to use multisigs via polkadot-js (see the screenshot below).

Screenshot 2023-02-03 at 23 27 36

The problem is solved by renaming the identifier of `pallet_multisig` from `MultiSig` to `Multisig`.


sea212 (approved):
Good discovery. Interesting. The technical committee page works the same way, it requires a specific name in the runtime. Wondering if other pallets are still affected by this.

Chralt98 (approved):

PR #971, opened 2023-02-13
RPC for querying prices for all assets of given pool_id.

### LINKED ISSUE #804: Add rpc call to fetch historical asset prices for a pool over a range of blocks.
The current SDK only has support for listing price history for assets in a pool when connecting to the indexer, and it is my aim to be able to provide some capacity of this functionality using node rpc alone in the sdk v2.

The currently implemented rpc method `swaps.get_spot_prices` satisfies the requirement for getting the prices of one asset over a range of blocks, but needs to be called `N` amount of times for the number of assets in the pool.

I am suggesting somethings like the following rpc interface; that would list all asset prices pr _non ztg_ asset in the pool for a range of blocks. _My types might be wrong so think of it as pseudocode if they are_ ^^

```rust
fn get_asset_spot_prices_for_pool(
  &self,
  pool_id: PoolId,
  blocks: Vec<BlockNumber>,
) -> Result<BTreeMap<Asset, Vec<SerdeWrapper<Balance>>>>;
```

_This would make it less io intensive to fetch all relevant asset price histories when viewing a market through rpc alone and you want some price history graph or the like; as you would have to make only one call._

**Also relevant to ensure correct typing when interacting with the rpc calls in the sdk is this issue:** #803

vivekvpandya:
Fixes #804


zrml/swaps/rpc/src/lib.rs
```diff
@@ -75,6 +76,14 @@ where
         with_fees: bool,
         blocks: Vec<BlockNumber>,
     ) -> RpcResult<Vec<SerdeWrapper<Balance>>>;
+
+    #[method(name = "swaps_getAssetSpotPricesForPool")]
+    async fn get_asset_spot_prices_for_pool(
```
maltekliemann, line 81: I'd say this needs a different name. Otherwise the difference to the function `get_spot_prices` is unclear. After all, that function also gets spot prices of assets which are in a pool. I suggest `get_spot_prices_all` or `get_all_spot_prices`. Open to suggestions.

zrml/swaps/rpc/src/lib.rs
```diff
@@ -205,4 +215,56 @@ where
...
+            })?
+            .map_err(|e| {
+                CallError::Custom(ErrorObject::owned(
+                    Error::RuntimeError.into(),
+                    "Unable to get pool. DispatchError",
+                    Some(format!("{:?}", e)),
+                ))
+            })?;
+        let mut result_map = BTreeMap::<Asset<MarketId>, Vec<SerdeWrapper<Balance>>>::new();
+        for asset in &pool.assets {
```
maltekliemann, line 244: This seems like we're mixing business logic and RPC logic. The structure of the code should be the following:

- The runtime API exposes some of the storage/getter of `zrml_swap` to the RPC module
- The RPC module queries one of the entry points of the runtime API

In other words, you can expose a function `get_all_spot_prices` (or whatever its called) in `zrml_swaps`, which the runtime API exposes to the RPC module. This is the pattern that we've already used for all other RPC calls. The advantages (in addition to a clear separation of responsibility) are:

- The function below is much easier to write directly inside `zrml_swaps`.
- The behavior is testable without additional work. Just write the function in `zrml_swaps` and verify that it returns correct values there. Speaking of which, we definitely need tests for this function (can easily be derived from tests of `get_spot_price`).
- The RPC call is just a couple of lines long and looks pretty much exactly like the others.

primitives/src/pool.rs
```diff
@@ -55,8 +57,8 @@ where
 
 impl<Balance, MarketId> MaxEncodedLen for Pool<Balance, MarketId>
 where
-    Balance: MaxEncodedLen,
-    MarketId: MaxEncodedLen,
+    Balance: MaxEncodedLen + Decode,
+    MarketId: MaxEncodedLen + Decode,
```
maltekliemann, line 61: It's not necessary to require that `Balance` and `MarketId` implement `Decode` (here and further below). When deriving `Decode`, the compiler implements `Decode` if all generic parameters satisfy the constraint `Decode`. In particular, it's not necessary for the functions in `arbitrage.rs` to require that trait bound.

runtime/common/src/lib.rs
```diff
@@ -1491,6 +1491,12 @@ macro_rules! create_runtime_api {
                 fn pool_shares_id(pool_id: PoolId) -> Asset<SerdeWrapper<MarketId>> {
                     Asset::PoolShare(SerdeWrapper(pool_id))
                 }
+
+        fn pool_by_id(
+            pool_id: PoolId,
+        ) -> Result<Pool<Balance, MarketId>, DispatchError> {
+            Swaps::pool_by_id(pool_id)
+        }
```
maltekliemann, line 1499: The formatting here is broken.

zrml/swaps/runtime-api/src/lib.rs
```diff
@@ -19,16 +19,20 @@
...
+use zeitgeist_primitives::types::{Asset, Pool, SerdeWrapper};
 
 sp_api::decl_runtime_apis! {
     pub trait SwapsApi<PoolId, AccountId, Balance, MarketId> where
         PoolId: Codec,
         AccountId: Codec,
-        Balance: Codec + MaybeDisplay + MaybeFromStr + MaxEncodedLen,
-        MarketId: Codec + MaxEncodedLen,
+        Balance: Codec + MaybeDisplay + MaybeFromStr + MaxEncodedLen + Decode,
+        MarketId: Codec + MaxEncodedLen + Decode,
```
maltekliemann, line 34: No need to ask for these to implement `Decode` if `Pool<Balance, MarketId>` is required to implement `Decode`.

maltekliemann (changes requested):
In addition to the comments made in the code, could you please ensure that the copyright notices are updated.

zrml/swaps/rpc/src/lib.rs
```diff
@@ -205,4 +215,56 @@ where
...
+                    Some(e.to_string()),
+                ))
+            })?
+            .map_err(|e| {
+                CallError::Custom(ErrorObject::owned(
+                    Error::RuntimeError.into(),
+                    "Unable to get pool. DispatchError",
+                    Some(format!("{:?}", e)),
+                ))
+            })?;
```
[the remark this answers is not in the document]
vivekvpandya: pool_by_id() returns Result type too.

zrml/swaps/rpc/src/lib.rs
```diff
@@ -205,4 +215,56 @@ where
...
+        let mut result_map = BTreeMap::<Asset<MarketId>, Vec<SerdeWrapper<Balance>>>::new();
+        for asset in &pool.assets {
+            if asset != &pool.base_asset {
+                let prices: Vec<Result<SerdeWrapper<Balance>, CallError>> = blocks
+                    .clone()
+                    .into_iter()
+                    .map(|block| {
+                        let hash = BlockId::number(block);
+                        let res = api
+                            .get_spot_price(&hash, &pool_id, &asset, &pool.base_asset, with_fees)
```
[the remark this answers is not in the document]
vivekvpandya: so IIUC using asset as in_asset gives value of input asset given one unit of output asset right?
if we change as you propose it will be how much base asset needed for output asset.

zrml/swaps/rpc/src/lib.rs
vivekvpandya: I think code won't be much cleaner than this, because we want prices of all assets over blocks specified in Vec<>. 
So if we add method in Swaps which just gives prices for all assets then also we need to iterate over all blocks and create vector and hasmap from the results. 
Or we can create BTreeMap with (asset, block_number) as key then it may be easy to collect().

zrml/swaps/rpc/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: > if we change as you propose it will be how much base asset needed for one unit of output asset.

That's correct.

zrml/swaps/rpc/src/lib.rs
maltekliemann: The code will definitely be cleaner (as mentioned above, separation of responsibilities and testability). I'd say we add `fn get_all_spot_prices(...) -> BTreeMap<Asset, Vec<Balance>>` to `zrml_swaps` and add tests. The runtime API directly exposes this functionality, and then all the RPC has to do something very similar to `get_spot_prices`.

zrml/swaps/rpc/src/lib.rs
```diff
@@ -205,4 +217,37 @@ where
...
+
+    async fn get_all_spot_prices(
+        &self,
+        pool_id: PoolId,
+        with_fees: bool,
+        blocks: Vec<NumberFor<Block>>,
+    ) -> RpcResult<BTreeMap<SerdeWrapper<NumberFor<Block>>, Vec<(Asset<MarketId>, Balance)>>> {
+        let api = self.client.runtime_api();
+        let mut res: BTreeMap<SerdeWrapper<NumberFor<Block>>, Vec<(Asset<MarketId>, Balance)>> =
+            BTreeMap::new();
```
maltekliemann, line 229: You don't need this type annotation. Since you basically just copying the annotation from the function's signature, it's best to remove this:

```suggestion
        let mut res = BTreeMap::new();
```

zrml/swaps/rpc/src/lib.rs
```diff
@@ -75,6 +77,14 @@ where
...
         blocks: Vec<BlockNumber>,
     ) -> RpcResult<Vec<SerdeWrapper<Balance>>>;
+
+    #[method(name = "swaps_getAllSpotPrices")]
+    async fn get_all_spot_prices(
+        &self,
+        pool_id: PoolId,
+        with_fees: bool,
+        blocks: Vec<BlockNumber>,
+    ) -> RpcResult<BTreeMap<SerdeWrapper<BlockNumber>, Vec<(Asset<MarketId>, Balance)>>>;
```
maltekliemann, line 87: I don't see the point of having the `SerdeWrapper` around the `BlockNumber`.

maltekliemann (changes requested):

zrml/swaps/rpc/src/lib.rs
maltekliemann: This should work without `SerdeWrapper` unless `BlockNumber` is `u128`.

zrml/swaps/src/lib.rs
```diff
@@ -1470,6 +1470,22 @@ mod pallet {
...
+        ) -> Result<Vec<(Asset<MarketIdOf<T>>, BalanceOf<T>)>, DispatchError> {
+            let pool = Self::pool_by_id(*pool_id)?;
+            pool.assets
+                .into_iter()
+                .map(|asset| {
+                    let spot_price =
+                        Self::get_spot_price(pool_id, &pool.base_asset, &asset, with_fees)?;
+                    Ok((asset, spot_price))
+                })
+                .collect()
```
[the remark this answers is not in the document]
vivekvpandya: is it really needed? I mean JS can do it.

zrml/swaps/rpc/src/lib.rs
```diff
@@ -205,4 +216,38 @@ where
...
+                        Error::RuntimeError.into(),
+                        "Unable to get_all_spot_prices.",
+                        Some(e.to_string()),
+                    ))
+                })?
+                .map_err(|e| {
+                    CallError::Custom(ErrorObject::owned(
+                        Error::RuntimeError.into(),
+                        "Unable to get_all_spot_prices. DispatchError",
+                        Some(format!("{:?}", e)),
```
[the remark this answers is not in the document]
vivekvpandya: DispatchError does not have ToString or Display implemented. So I change other error to format!() to make things consistent.

zrml/swaps/src/lib.rs
```diff
@@ -1470,6 +1470,22 @@ mod pallet {
             }
         }
 
+        // Retunrs vector of units of base_asset required to buy one unit of given asset.
```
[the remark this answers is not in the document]
vivekvpandya: Vec< (Asset, Balance)> here given asset means corresponding asset.

zrml/swaps/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: Hmm... I may have been overzealous here. I mean... If you don't filter the base asset, you end up with the following tuple in each value of the map: `(base_asset, (1 + swap_fee) * BASE)`. I was going to say that we really don't need that info, but let's ask @yornaath: Is that info any use to you?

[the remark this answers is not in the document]
maltekliemann: Leaving this the way it is right now. Will change this later when we kick ZTG from the pools.

maltekliemann (approved):

sea212 (approved):
Tested using curl.

- [x] Works using single and multiple blocks
- [x] Provides correct results
- [ ] Handles errors properly

PR #983, opened 2023-03-03
The new randomness pallet

Chralt98:


Chralt98:
Halted in favour of #984 to avoid the maintenance of our own randomness pallet.

PR #990, opened 2023-03-13
Bump Battery Station `spec_version` in migration runtime

sea212:


runtime/battery-station/src/lib.rs
```diff
@@ -91,7 +91,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("zeitgeist"),
     impl_name: create_runtime_str!("zeitgeist"),
     authoring_version: 1,
-    spec_version: 42,
+    spec_version: 44,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 19,
```
Chralt98, line 97: I think you bumped `transaction_version` too to `20`.

sea212: It does not matter but for the sake of equality to the mainnet I'll bump it

sea212: bcaf4998abacab0a8602add764eaacc72c3cc556

Chralt98 (approved):

PR #991, opened 2023-03-16
Add court tests

Chralt98:


Chralt98:
"abandoned" because merged into #976

PR #997, opened 2023-04-19
Add orml-tokens calls to runtime's call enum

sea212:
It is currently not possible to interact with the tokens pallet via calls. The governance body is unable to utilize those calls. In addition to that, the sudo account cannot be used to generate tokens to test features (such as testing market with fakedot).


sea212:
My bad, it seems everything can be achieved using the `Currency` call enum, except transferring foreign assets from the treasury. But this can be achieved using `utility.dispatchAs` to invoke `assetManager.transfer` as the parachain account.

PR #999, opened 2023-04-25
Remove `MinLiquidity` parameter

### LINKED ISSUE #998: Review minLiquidity constant usage
When multiple base assets are allowed for a market the `minLiquidity` constant doesn't make much sense as each asset will have a different value. I think either this restriction should be removed or each asset should be configured with it's own `minLiquidity`

maltekliemann:
Fixes #998.

### DIFF
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -1471,9 +1471,30 @@ mod pallet {
             T::PalletId::get().into_sub_account_truncating((*pool_id).saturated_into::<u128>())
         }
 
-        // The minimum allowed balance in a liquidity pool.
+        // The minimum allowed balance of `asset` in a liquidity pool.
         pub(crate) fn min_balance(asset: Asset<MarketIdOf<T>>) -> BalanceOf<T> {
             T::AssetManager::minimum_balance(asset).max(MIN_BALANCE.saturated_into())
+        }
+
+        // Returns the minimum allowed balance allowed for a pool with id `pool_id` containing
+        // `assets`.
+        //
+        // The minimum allowed balance is the maximum of all minimum allowed balances of assets
+        // contained in the pool, _including_ the pool shares asset. This ensures that none of the
+        // accounts involved are slashed when a pool is created with the minimum amount.
+        //
+        // **Should** only be called if `assets` is non-empty. Note that there need not exist a pool
+        // with the specified `pool_id`.
+        pub(crate) fn min_balance_of_pool(
+            pool_id: PoolId,
+            assets: &[Asset<MarketIdOf<T>>],
+        ) -> BalanceOf<T> {
+            assets
+                .iter()
+                .map(|asset| Self::min_balance(*asset))
+                .max()
+                .unwrap_or_else(|| MIN_BALANCE.saturated_into())
+                .max(Self::min_balance(Self::pool_shares_id(pool_id)))
         }
 
         fn ensure_minimum_liquidity_shares(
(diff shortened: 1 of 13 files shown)


zrml/swaps/src/lib.rs
```diff
@@ -1475,11 +1471,32 @@ mod pallet {
             T::PalletId::get().into_sub_account_truncating((*pool_id).saturated_into::<u128>())
         }
 
-        // The minimum allowed balance in a liquidity pool.
+        // The minimum allowed balance of `asset` in a liquidity pool.
```
sea212, line 1474:
```suggestion
        /// The minimum allowed balance of `asset` in a liquidity pool.
```

sea212 (commented):
Looks fine. Probably makes sense to test that the pool creation fails when the liquidity constraint is not met.

maltekliemann:
> Looks fine. Probably makes sense to test that the pool creation fails when the liquidity constraint is not met.

This should do it, right? https://github.com/zeitgeistpm/zeitgeist/blob/649880168d7c1a40902a8a1993b9d8ec1215d1e9/zrml/swaps/src/tests.rs#L2376-L2397

sea212 (dismissed):
Let's stop the unintentional bullying of LPs :sweat_smile:

zrml/swaps/src/lib.rs
```diff
@@ -1475,11 +1471,32 @@ mod pallet {
...
+            pool_id: PoolId,
+            assets: &[Asset<MarketIdOf<T>>],
+        ) -> BalanceOf<T> {
+            assets
+                .iter()
+                .map(|asset| Self::min_balance(*asset))
+                .max()
+                .unwrap_or_else(|| MIN_BALANCE.saturated_into())
+                .max(Self::min_balance(Self::pool_shares_id(pool_id)))
+        }
```
Chralt98, line 1498: Do you know where the existential deposits of the pool assets are set (`min_balance`)? If yes, can you please share me the reference?

zrml/swaps/src/tests.rs
```diff
@@ -1983,8 +1994,8 @@ fn swap_exact_amount_out_exchanges_correct_values_with_cpmm() {
         assert_all_parameters(
             [239898989900, _26, _25, _25],
             0,
-            [_101 + 101010100, _99, _100, _100],
-            _100,
+            [LIQUIDITY + _1 + 101010100, LIQUIDITY - _1, _100, _100],
+            LIQUIDITY,
```
Chralt98, line 1998:
```suggestion
            [LIQUIDITY + _1 + 101010100, LIQUIDITY - _1, LIQUIDITY, LIQUIDITY],
            LIQUIDITY,
```

Chralt98 (approved):
LGTM, because the `MinLiquidity` parameter was one dimensional. Now we are multi-dimensional 🥇 .

[nitpick] Just an additional occurrence of `MinLiquidity` [here](https://github.com/zeitgeistpm/zeitgeist/blob/649880168d7c1a40902a8a1993b9d8ec1215d1e9/primitives/src/traits/swaps.rs#L33-L34).

zrml/swaps/src/lib.rs
maltekliemann: `minimum_balance` is just the `ExistentialDeposit`, which is defined in `runtime/*/parameters.rs`.

Chralt98:
```rust
    // Existential deposits used by orml-tokens.
    // Only native ZTG and foreign assets should have an existential deposit.
    // Winning outcome tokens are redeemed completely by the user, losing outcome tokens
    // are cleaned up automatically. In case of scalar outcomes, the market account can have dust.
    // Unless LPs use `pool_exit_with_exact_asset_amount`, there can be some dust pool shares remaining.
    // Explicit match arms are used to ensure new asset types are respected.
    pub ExistentialDeposits: |currency_id: CurrencyId| -> Balance {
        match currency_id {
            Asset::CategoricalOutcome(_,_) => ExistentialDeposit::get(),
            Asset::CombinatorialOutcome => ExistentialDeposit::get(),
            Asset::PoolShare(_)  => ExistentialDeposit::get(),
            Asset::ScalarOutcome(_,_)  => ExistentialDeposit::get(),
            #[cfg(feature = "parachain")]
            Asset::ForeignAsset(id) => {
                let maybe_metadata = <
                    orml_asset_registry::Pallet<super::Runtime> as orml_traits::asset_registry::Inspect
                >::metadata(&Asset::ForeignAsset(*id));

                if let Some(metadata) = maybe_metadata {
                    return metadata.existential_deposit;
                }

                1
            }
            #[cfg(not(feature = "parachain"))]
            Asset::ForeignAsset(_) => ExistentialDeposit::get(),
            Asset::Ztg => ExistentialDeposit::get(),
        }
    };
```

Ah fine, got it thanks!

sea212 (approved):

PR #1005, opened 2023-05-25
Transfer token dust to treasury

sea212:


runtime/battery-station/src/parameters.rs
```diff
@@ -144,7 +144,6 @@ parameter_types! {
 
     // ORML
     pub const GetNativeCurrencyId: CurrencyId = Asset::Ztg;
-    pub DustAccount: AccountId = PalletId(*b"orml/dst").into_account_truncating();
```
Chralt98, line 147: Should we look up the amount of funds on this account? Who controls this account?

Chralt98 (approved):

runtime/battery-station/src/parameters.rs
sea212: Only root can control this account. Lookup is not necessary, since tokens with value were not used yet.

PR #1007, opened 2023-05-26
Move ZBS/ZTG dust into the treasury

sea212:
Whenever an account is killed due to not meeting the existential deposit requirements, the remaining funds (dust) are moved into the treasury.


Chralt98 (dismissed):
It was burned before? [This](https://github.com/paritytech/substrate/blob/46127a9aacdc719c53f1129a92b9e7737a0f55ad/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs#L43-L55) is what I found for the default implementation.

Chralt98 (approved):

PR #1009, opened 2023-05-26
Fix pipeline

sea212:
A dependency is missing in the dockerfile: https://github.com/zeitgeistpm/zeitgeist/actions/runs/5089309589/jobs/9146777074

This PR unifies the build dependency process within `./scripts/init.sh` and updates all workflows and the Dockerfile to use `./scripts/init.sh`. The script always installs the rust toolchain based on the `rust-toolchain` file, which was previously not the case in every context. Lastly, the docker image was updated to use a minimal version of Ubuntu Jammy (update from 18.04 LTS to 22.04).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,30 @@
 # Based from https://github.com/paritytech/substrate/blob/master/.maintain/Dockerfile
 
-FROM phusion/baseimage:bionic-1.0.0 as builder
+FROM phusion/baseimage:jammy-1.0.1 as builder
 LABEL maintainer="hi@zeitgeit.pm"
 LABEL description="This is the build stage for the Zeitgeist node. Here is created the binary."
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-ARG PROFILE=release
+ARG PROFILE=production
 ARG FEATURES=default
 WORKDIR /zeitgeist
 
 COPY . /zeitgeist
 
 RUN apt-get update && \
-    apt-get dist-upgrade -y -o Dpkg::Options::="--force-confold" && \
-    apt-get install -y cmake pkg-config libssl-dev git clang libclang-dev
+    apt-get dist-upgrade -y -o Dpkg::Options::="--force-confold"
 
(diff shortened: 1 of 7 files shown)


scripts/init.sh
```diff
@@ -2,11 +2,17 @@
...
+if [ -f /.dockerenv ]; then
+   apt-get update && \
+   apt-get install -y build-essential clang curl libssl-dev protobuf-compiler
+else
+   sudo apt-get update && \
+   sudo apt-get install -y build-essential clang curl libssl-dev protobuf-compiler
 fi
 
-rustup target add wasm32-unknown-unknown --toolchain nightly-2022-04-13
+curl https://sh.rustup.rs -sSf | sh -s -- -y && \
```
Chralt98, line 16: The newest rust compiler version is taken, right? not our specified from the toolchain at this point.

Dockerfile
```diff
@@ -1,34 +1,30 @@
...
-ARG PROFILE=release
+ARG PROFILE=production
 ARG FEATURES=default
 WORKDIR /zeitgeist
 
 COPY . /zeitgeist
 
 RUN apt-get update && \
-    apt-get dist-upgrade -y -o Dpkg::Options::="--force-confold" && \
-    apt-get install -y cmake pkg-config libssl-dev git clang libclang-dev
```
Chralt98, line 17: `cmake` and `pkg-config` were removed.

Chralt98 (approved):
Was the Dockerfile build tested?

Dockerfile
sea212: We don't need it in the first stage dockerfile which builds the Zeitgeist binary. I have successfully built the docker image resulting from the dockerfile.

scripts/init.sh
sea212: It installs rustup and cargo. Then the script executes `rustup show` to install the toolchain and targets specified in the "rust-toolchain" file

sea212: It installs rustup and cargo. Then the script executes `rustup show` to install the toolchain and targets specified in the "rust-toolchain" file

PR #1011, opened 2023-05-28
Use reasonable storage fee

sea212:
During the integration of the contracts pallet it became evident to me that currently the chain overcharges ZTG for storage in certain pallets. A function `deposit` exists that is used to calculate the total fee in ZTG per storage key and bytes associated to it.

With the previous configuration, using up the total supply of ZTG would only allow to store 100 MB of bytes, not considering the costs per item. The cost per byte and items in ZTG in relation to the total ZTG supply was aligned with the cost of DOT in relation to the DOT supply used in [Polkadot](https://github.com/paritytech/polkadot/blob/97930a8b2cc211ecfb80f289c4c578a9157e6dc4/runtime/polkadot/constants/src/lib.rs#L36-L38). 

In addition to the `deposit` adjustment, `pallet-identity` now also uses the `deposit` function to determine the storage fee. It is aligned to the one used in [Polkadot](https://github.com/paritytech/polkadot/blob/97930a8b2cc211ecfb80f289c4c578a9157e6dc4/runtime/polkadot/src/lib.rs#L618C2-L619), effectively reducing the storage fee to register an identity from a minimum of 125 ZTG to a minimum of 1.5243 ZTG.

Affected pallets:
  - MultiSig, [no storage migration required](https://github.com/paritytech/substrate/blob/polkadot-v0.9.32/frame/multisig/src/lib.rs#L110-L111)
  - Identity, [no storage migration required](https://github.com/paritytech/substrate/blob/polkadot-v0.9.32/frame/identity/src/types.rs#L385-L386)
  - Preimage, [no storage migration required](https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/frame/preimage/src/lib.rs#L65-L70)
  - Proxy, [no storage migration required](https://github.com/paritytech/substrate/blob/polkadot-v0.9.32/frame/proxy/src/lib.rs#L565-L566)

### DIFF
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -42,8 +42,10 @@ pub const CENT: Balance = BASE / 100; // 100_000_000
 pub const MILLI: Balance = CENT / 10; //  10_000_000
 pub const MICRO: Balance = MILLI / 1000; // 10_000
 
+/// Storage cost per byte and item.
+// Approach: Achieve same cost per item and bytes in relation to total supply as on Polkadot.
 pub const fn deposit(items: u32, bytes: u32) -> Balance {
-    items as Balance * 20 * BASE + (bytes as Balance) * 100 * MILLI
+    items as Balance * 150 * CENT + (bytes as Balance) * 75 * MICRO
 }
 
 // Rikiddo and TokensConfig
--- a/docs/changelog_for_devs.md
+++ b/docs/changelog_for_devs.md
@@ -14,6 +14,7 @@ APIs/RPC interface.
 
 ## v0.3.9
 
+[#1011]: https://github.com/zeitgeistpm/zeitgeist/pull/1011
 [#937]: https://github.com/zeitgeistpm/zeitgeist/pull/937
 [#903]: https://github.com/zeitgeistpm/zeitgeist/pull/903
 
@@ -21,6 +22,9 @@ APIs/RPC interface.
 
 - ⚠️ Add `outsider` field to `MarketBonds` struct. In particular, the `Market`
   struct's layout has changed ([#903]).
+- Adjust `deposit` function used to calculate storage fees for the following
+  pallets: identity, multisig, preimage, proxy. The cost of adding a identity
+  reduced from a minimum of 125 ZTG to a minimum of 1.5243 ZTG ([#1011])
 
 ### Fixed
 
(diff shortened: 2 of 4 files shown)


primitives/src/constants.rs
```diff
@@ -42,8 +42,10 @@ pub const CENT: Balance = BASE / 100; // 100_000_000
 pub const MILLI: Balance = CENT / 10; //  10_000_000
 pub const MICRO: Balance = MILLI / 1000; // 10_000
 
+/// Storage cost per byte and item.
+// Approach: Achieve same cost per item and bytes in relation to total supply as on Polkadot.
 pub const fn deposit(items: u32, bytes: u32) -> Balance {
-    items as Balance * 20 * BASE + (bytes as Balance) * 100 * MILLI
+    items as Balance * 150 * CENT + (bytes as Balance) * 75 * MICRO
```
Chralt98, line 48: It might be a good idea to achieve the same cost per item as on Polkadot in relation to the USD instead of the total supply :-D. Because now I the default cost on [Polkadot is 20 DOTs](https://github.com/paritytech/polkadot/blob/97930a8b2cc211ecfb80f289c4c578a9157e6dc4/runtime/polkadot/constants/src/lib.rs#L36-L38), so 100 USD, right? Here it is (without bytes) 1.5 ZTG, so 0.15 USD, right?

sea212: There are two problems with your suggested approach:
1. It does not consider market cap. Polkadot is much more valuable, as it has a market cap of approx. 6 Billion USD right now. ZTG market cap is around approx. 5 Million USD. If Polkadot market cap would be the same as ZTG, you would also only pay roughly 0.07 USD. With current ZTG price, you also pay roughly 0.07 USD (1.5 * 5 cent). On the other hand, this means that as ZTG grows and becomes more valuable, the price for storage will also increase accordingly. If ZTG has a market cap of 6 Billion USD like Polkadot now, storage will also become more valuable (as the product is more valuable and utilized) and one storage item will cost also 100 USD. In addition to that, every ZTG already locked for storage deposit will also increase in value. So everyone who locks ZTG now for a storage item, locks only about 0.07 USD in value. But in the future, should ZTG reach a market cap like Polkadot, those ZTG will be worth 100 USD or even more.
2. If use the previous approach and charge ZTG worth 100 USD right now given the low market cap, we can only offer about 100 MB of storage space assuming that all the existing ZTG is only used to pay for storage space.

Chralt98 (approved):

PR #1012, opened 2023-05-28
Include pallet-contracts into the runtime

sea212:
Includes pallet-contracts into the runtime and uses as sensible configuration based on the reference contracts runtime ["kitchensink"](https://github.com/paritytech/substrate/blob/polkadot-v0.9.32/bin/node/runtime/src/lib.rs).
Alex@Parity:
> Just look at the kitchensink runtime. It uses our recommended Config.

Runtime benchmarks and apis have been added as well. Proper weights will be added during preparation of the release.

The contract upload is permissioned on the mainnet. The root origin has to dispatch the upload on behalf of an account using `utility.dispatch_as`.

Contracts was tested on a dev instance of the chain. Everything works as expected.


runtime/zeitgeist/src/lib.rs
```diff
@@ -142,6 +145,13 @@ impl Contains<RuntimeCall> for IsCallable {
                     _ => true,
                 }
             }
+            // Permissioned contracts: Only deployable via utility.dispatch_as(...)
+            RuntimeCall::Contracts(inner_call) => match inner_call {
+                instantiate_with_code { .. } => false,
+                instantiate_with_code_old_weight { .. } => false,
+                upload_code { .. } => false,
+                _ => true,
+            },
```
pgherveou, line 154: I would flip the logic here and specify the one you want to authorize and then use _ => false to deny everything else

runtime/common/src/lib.rs
```diff
@@ -655,6 +656,27 @@ macro_rules! impl_config_traits {
             type WeightInfo = weights::pallet_collective::WeightInfo<Runtime>;
         }
 
+        impl pallet_contracts::Config for Runtime {
+            type AddressGenerator = pallet_contracts::DefaultAddressGenerator;
+            type CallFilter = frame_support::traits::Nothing;
+            type CallStack = [pallet_contracts::Frame::<Runtime>; 5];
```
Chralt98, line 662: https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/bin/node/runtime/src/lib.rs#L1174

lower call stack more conservative?

runtime/battery-station/src/parameters.rs
```diff
@@ -80,6 +80,15 @@ parameter_types! {
     pub const TechnicalCommitteeMaxProposals: u32 = 64;
     pub const TechnicalCommitteeMotionDuration: BlockNumber = 7 * BLOCKS_PER_DAY;
 
+    // Contracts
+    pub const ContractsDeletionQueueDepth: u32 = 128;
+    pub ContractsDeletionWeightLimit: Weight = Perbill::from_percent(10) * RuntimeBlockWeights::get().max_block;
```
Chralt98, line 85: https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/bin/node/runtime/src/lib.rs#L1151-L1155

runtime/battery-station/src/parameters.rs
```diff
@@ -80,6 +80,15 @@ parameter_types! {
...
+    pub const ContractsDepositPerByte: Balance = deposit(0,1);
+    pub const ContractsDepositPerItem: Balance = deposit(1,0);
+    pub const ContractsMaxCodeLen: u32 = 246 * 1024;
```
Chralt98, line 88: https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/bin/node/runtime/src/lib.rs#L1182

runtime/common/src/lib.rs
```diff
@@ -303,6 +303,7 @@ macro_rules! create_runtime {
                 Identity: pallet_identity::{Call, Event<T>, Pallet, Storage} = 30,
                 Utility: pallet_utility::{Call, Event, Pallet, Storage} = 31,
                 Proxy: pallet_proxy::{Call, Event<T>, Pallet, Storage} = 32,
+                Contracts: pallet_contracts = 33,
```
Chralt98, line 306: When using this, it means that all possible pallet components are included, right?

runtime/zeitgeist/src/lib.rs
Chralt98, line 154: Will later look into this more specific to find potential calls that are not covered.

Chralt98 (commented):
Where I can find the configuration values of the kitchensink runtime? 

Seems like you provided a link to the normal runtime.

runtime/zeitgeist/src/parameters.rs
```diff
@@ -80,6 +80,15 @@ parameter_types! {
     pub const TechnicalCommitteeMaxProposals: u32 = 64;
     pub const TechnicalCommitteeMotionDuration: BlockNumber = 7 * BLOCKS_PER_DAY;
 
+    // Contracts
+    pub const ContractsDeletionQueueDepth: u32 = 128;
+    pub ContractsDeletionWeightLimit: Weight = Perbill::from_percent(10) * RuntimeBlockWeights::get().max_block;
+    pub const ContractsDepositPerByte: Balance = deposit(0,1);
+    pub const ContractsDepositPerItem: Balance = deposit(1,0);
+    pub const ContractsMaxCodeLen: u32 = 123 * 1024;
```
maltekliemann, line 88: Not sure if this deviation from kitchensink was planned.

```suggestion
    pub const ContractsMaxCodeLen: u32 = 128 * 1024;
```

runtime/zeitgeist/src/parameters.rs
```diff
@@ -80,6 +80,15 @@ parameter_types! {
     pub const TechnicalCommitteeMaxProposals: u32 = 64;
     pub const TechnicalCommitteeMotionDuration: BlockNumber = 7 * BLOCKS_PER_DAY;
 
+    // Contracts
+    pub const ContractsDeletionQueueDepth: u32 = 128;
+    pub ContractsDeletionWeightLimit: Weight = Perbill::from_percent(10) * RuntimeBlockWeights::get().max_block;
```
maltekliemann, line 85: Should we replace this with 10% of the weight reserved for normal dispatches?

runtime/common/src/lib.rs
maltekliemann, line 662: I must admit I don't have any experience writing ink!, but five callstack depth seems very close to the pain threshold. Are there security concerns at work here?

runtime/common/src/lib.rs
```diff
@@ -1330,6 +1358,69 @@ macro_rules! create_runtime_api {
                 }
             }
 
+            impl pallet_contracts::ContractsApi<Block, AccountId, Balance, BlockNumber, Hash>
```
maltekliemann, line 1361: The formatting seems off in a couple of places here (Allman brackets and missing trailing comma).

sea212: I put that code into a file inside the Zeitgeist repository and executed `cargo fmt`. In regards to the Allman brackets, that's what cargo fmt does. In case of the trailing comma, https://github.com/zeitgeistpm/zeitgeist/pull/1012/commits/17a6ef86cd4b822d3576a277c1038513d333a7c3.

runtime/zeitgeist/src/parameters.rs
sea212: They changed it in the lasted version, not sure why though, I just aligned the value: https://github.com/paritytech/substrate/blob/a626589bf9363b0b8659b75a1cf18a246cf1ed58/bin/node/runtime/src/lib.rs#L1242

runtime/battery-station/src/parameters.rs
sea212: I decided to deviate here since I want to avoid that extrinsics interacting with the protocol are postponed due to contract deletion taking up the whole block (or blocks). Since contract deletion is low priority, it can be spanned over multiple blocks.

runtime/zeitgeist/src/parameters.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1012/commits/bf2035760aa69cc40512fd4e532551f203f94ece

maltekliemann (approved):

runtime/zeitgeist/src/lib.rs
```diff
@@ -142,6 +146,16 @@ impl Contains<RuntimeCall> for IsCallable {
...
+            // Permissioned contracts: Only deployable via utility.dispatch_as(...)
+            RuntimeCall::Contracts(inner_call) => match inner_call {
+                call { .. } => true,
+                call_old_weight { .. } => true,
+                instantiate { .. } => true,
+                instantiate_old_weight { .. } => true,
+                remove_code { .. } => true,
+                set_code_contracts { .. } => true,
+                _ => false,
+            },
```
Chralt98, line 158: Ok, I scanned all possible calls from [here](https://github.com/paritytech/substrate/blob/5ea6d95309aaccfa399c5f72e5a14a4b7c6c4ca1/frame/contracts/src/lib.rs#L365-L676). Seems like your call filter does apply on every extrinsic which allows to upload a `code: Vec<u8>`, but the utility pallet can handle those dispatches. I see 👍

Chralt98 (approved):

Chralt98 (changes requested):
Please regard the "approve" comment above.

EDIT: everything is fine

runtime/zeitgeist/src/lib.rs
Chralt98: `dispatch_bypass_filter` is only allowed for the root origin ✅

PR #1014, opened 2023-05-31
Make `MaxMarketLifetime` a constant

maltekliemann:
This fixed an oversight from the PR introducing the `MaxMarketLifetime` config parameter.


Chralt98 (approved):
I am interested: What happened if you don't define the type configuration parameter as `#[pallet::constant]`? It does not appear in polkadot js under constants? What else?

I can only see [this](https://docs.rs/frame-support/latest/frame_support/attr.pallet.html#palletconstant) in the documentation. But what does this mean?

PR #1015, opened 2023-06-02
Move dust to treasury instead of dedicated dust account

maltekliemann:
Our dust collection is broken.

Dust is currently collected in `PalletId(*b"orml/dst").into_account_truncating()`, but that account is not on the dust removal whitelist and empty at genesis. This means that any transfer of an amount smaller than the existential deposit will fail. So every time that dust is collected, the [transfer fails quietly](https://github.com/open-web3-stack/open-runtime-module-library/blob/0482bf2b02ac8fea1279da0a34c0900fd369c444/tokens/src/lib.rs#L91-L108).


maltekliemann:
Nevermind, seems to already be fixed.

PR #1016, opened 2023-06-07
Fix docs of `MarketEdited`

maltekliemann:


sea212 (commented):
I see there are inconsistencies with the punctuation for the docstrings. While you are at it, could you align the punctuation please? ( I think we mostly end the docstring with a dot ).

PR #1021, opened 2023-06-12
Add style guide

maltekliemann:


sea212 (dismissed):

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,123 @@
...
+  `_ =>`. For example:
+
+  ```rust
+  match market_status {
+      MarketStatus::Active => {
+          // --- snip ---
+      }
+      _ => {
+          // --- snip ---
+      } // not allowed
```
Chralt98, line 99: I personally wouldn't be too strict here. There is a reason why this catch all pattern exists. Current example: I don't want to write all enum variants to [this match](https://github.com/zeitgeistpm/zeitgeist/blob/b6400f6016854e6b1f1d59c1473e1a352db5c200/runtime/common/src/lib.rs#L1026).

Chralt98 (dismissed):

sea212 (approved):

PR #1006, opened 2023-05-25
Test all features in test scripts

sea212:
This PR ensures that all crates within the workspace are tested.
In addition to that, it seperates the Coverage job into it's own workspace. This is a necessary action, since having both tests and coverage tests in one job results in GitHub hosted runners running out of disk space.

PR #1024, opened 2023-06-21
Separate tests into two jobs: standalone and parachain

sea212:
### What does it do?
It potentially fixes an "out of disk space" error in the pipeline

The fix separates the standalone and parachain tests into their respective workflow jobs, thus providing a whole instance for each case. 

No

None

https://github.com/zeitgeistpm/zeitgeist/pull/1006)](https://github.com/zeitgeistpm/zeitgeist/pull/1006

### References
None


Chralt98 (approved):
Let's try that!

PR #1028, opened 2023-06-22
Use argument in init.sh to avoid usage of sudo

sea212:
### What does it do?
Instead of relying on the ([seemingly unreliable](https://github.com/zeitgeistpm/zeitgeist/actions/runs/5347848379/jobs/9697515500#step:7:1295)) `/.dockerenv` file to determine a docker build requirement within `scripts/init.sh`, an optional argument is now used.

This PR is a result of a failed workflow.

No.

None.

[A previous PR](https://github.com/zeitgeistpm/zeitgeist/pull/1009) attempted to fix this already. While the solution worked until now, it seems it is not 100% reliable.

### References
None.

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 
+# Initializes a build environment.
+# Passing one argument results in avoiding the usage of sudo for privileged commands.
+
 set -e
 
 echo "*** Initializing build environment"
 
-# No sudo during docker build
-if [ -f /.dockerenv ]; then
+if [ -n "$1" ]; then
    apt-get update && \
    apt-get install -y build-essential clang curl libssl-dev protobuf-compiler
 else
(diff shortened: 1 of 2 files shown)


sea212:
Works: https://github.com/zeitgeistpm/zeitgeist/actions/runs/5348759942

scripts/init.sh
```diff
@@ -1,11 +1,13 @@
...
+# Initializes a build environment.
+# Passing one argument results in avoiding the usage of sudo for privileged commands.
+
 set -e
 
 echo "*** Initializing build environment"
 
-# No sudo during docker build
-if [ -f /.dockerenv ]; then
+if [ -n "$1" ]; then
```
Chralt98, line 10:
```suggestion
if [ "$1" == "nosudo" ]; then
```

Chralt98 (approved):

PR #1029, opened 2023-06-22
Fix name collision

sea212:
### What does it do?
Building the Zeitgeist runtime with srtool results in the following error:
>   error[E0423]: expected value, found struct variant `call`
     --> /build/runtime/zeitgeist/src/lib.rs:130:15
      |
  130 |         match call {
      |               ^^^^ you might want to surround a struct literal with parentheses: `(call { /* fields */ })`?
>
>  For more information about this error, try `rustc --explain E0423`.

The rust compiler used in the docker container is unable to distinguish `call`, which represents the `RuntimeCall` that is currently being process and `call`, which represents a dispatchable from `pallet-contracts`.

This PR assign unique names.

None.

No. 

- Renaming the contracts call variant.

No.

### References
None.


Chralt98 (approved):
reminds me of `set_code`

PR #1033, opened 2023-07-03
Release v0.3.9

sea212:
### What does it do?
It merges the changes applied during the release preparation of v0.3.9 to the main branch.

### What important points should reviewers know?
All the changes have already been reviewed, see referenced PRs below.

### Is there something left for follow-up PRs?
- Remove old migrations.
- Fix try-runtime checks on tag:
 > Thread 'main' panicked at 'spec version mismatch (local 46 != remote 45). This could cause some issues.',

### What alternative implementations were considered?
None.

### Are there relevant PRs or issues?
- https://github.com/zeitgeistpm/zeitgeist/pull/1025
- https://github.com/zeitgeistpm/zeitgeist/pull/1026
- https://github.com/zeitgeistpm/zeitgeist/pull/1028
- https://github.com/zeitgeistpm/zeitgeist/pull/1029
- https://github.com/zeitgeistpm/zeitgeist/pull/1030

### References
None.

PR #1035, opened 2023-07-03
Don't panic on spec_version mismatch during try-runtime

sea212:
### What does it do?
When a tag is pushed, a workflow is executed that runs `try-runtime` against the mainnet. When a tag is pushed, the `spec_version` usually is already incremented. `try-runtime` checks the condition that the spec version is equal on the remote and local runtime (that is used for testing). Consequently, this check will always fail. An example can be inspected at https://github.com/zeitgeistpm/zeitgeist/actions/runs/5359669650/jobs/9723563958?pr=1033:
> Thread 'main' panicked at 'spec version mismatch (local 46 != remote 45). This could cause some issues.'

This PR reduces the severity of this failing check from "error" to "warning".

/ 

/ 

/ 

- https://github.com/zeitgeistpm/zeitgeist/pull/1033

### References
/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ check-dummy:
 		try-runtime \
 		--execution=Native \
 		--chain=${TRYRUNTIME_CHAIN} \
+		--no-spec-check-panic \
 		on-runtime-upgrade \
 		live \
 		--uri=${TRYRUNTIME_URL}


Chralt98 (approved):
Checked here https://docs.substrate.io/reference/command-line-tools/try-runtime/ ✅

PR #1036, opened 2023-07-05
Add dispatch call indices

### LINKED ISSUE #981: Use call indices for dispatchable
By using call indices a certain internal order of dispatchable calls can be enforced. By doing so, `transaction_version` within `RuntimeVersion` does not have to be bumped if the order of the dispatchable calls changes. This allows to order dispatchable calls in another way, like lexicographical ordering and generally leads less potential for errors.

Here is an example: https://github.com/paritytech/cumulus/blob/d21bfd7ac5ab7b7faac7b315519b52c97f9d8319/pallets/parachain-system/src/lib.rs#L339

Chralt98:
### What does it do?

Fixes #981

Adds call indices to allow custom restructuring of dispatchables. 

None

None

None

None

### References

None

### DIFF
--- a/zrml/rikiddo/src/lib.rs
+++ b/zrml/rikiddo/src/lib.rs
@@ -1,3 +1,4 @@
+// Copyright 2022-2023 Forecasting Technologies LTD.
 // Copyright 2021-2022 Zeitgeist PM LLC.
 //
 // This file is part of Zeitgeist.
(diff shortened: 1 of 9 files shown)


zrml/rikiddo/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+// Copyright 2022-2023 Forecasting Technologies LTD.
```
sea212, line 1: Nothing was changed in this file

```suggestion
```

sea212 (dismissed):

sea212 (approved):

maltekliemann (approved):

PR #1039, opened 2023-07-07
Improve error message

maltekliemann:
### What does it do?

Improves the error message of the license checker.


Chralt98 (approved):
The true error should be more visible now.

PR #1042, opened 2023-07-17
Adjust mergify status checks and label management

sea212:
### What does it do?
- Completes the list of status checks required to pass before merging
- Fixes status check names required to pass before merging
- Removes the condition that merge checks are only executed when merging into `main`
- Mergify now removes invalid status labels when a conflict is detected and adds `s:revision-needed`

- Status labels and a non-conflict state machine should be always reflected in mergify
- Maybe the requirements for merging into branches should be update depending on the type of branch (i.e. main or release)

### References
- [Passing status checks not registered properly](https://github.com/zeitgeistpm/zeitgeist/runs/15037500885)
- [Propoer mergify status checks configuration](https://docs.mergify.com/conditions/#about-status-checks)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -27,4 +27,11 @@ pull_request_rules:
     actions:
         comment:
           message: This pull request is now in conflicts. Could you fix it @{{author}}? 🙏
+        label:
+          add:
+            - s:revision-needed
+          remove:
+            - s:accepted
+            - s:in-progress
+            - s:review-needed
 
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,13 +1,18 @@
 pull_request_rules:
   - name: Auto merge
     conditions:
-      - base=main
       - "#approved-reviews-by>=1"
       - "#changes-requested-reviews-by=0"
-      - check-success=format
-      - check-success=tests
-      - check-success=fuzz
-      - check-success=benchmark
+      - check-success=Coverage / Coverage
+      - check-success=Rust / Format
+      - check-success=Rust / Copyright Notices
+      - check-success=Rust / Checks (clippy)
+      - check-success=Rust / Checks (parachain)
+      - check-success=Rust / Checks (standalone)
+      - check-success=Rust / Quick check benchmarks
+      - check-success=Rust / Test standalone build
+      - check-success=Rust / Test parachain build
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -1,13 +1,18 @@
...
+      - check-success=Coverage / Coverage
+      - check-success=Rust / Format
+      - check-success=Rust / Copyright Notices
+      - check-success=Rust / Checks (clippy)
+      - check-success=Rust / Checks (parachain)
+      - check-success=Rust / Checks (standalone)
+      - check-success=Rust / Quick check benchmarks
+      - check-success=Rust / Test standalone build
+      - check-success=Rust / Test parachain build
+      - check-success=Rust / Fuzz
```
Chralt98, line 15: I checked the names and everything should be fine. From [here](https://github.com/zeitgeistpm/zeitgeist/blob/bac24e6f2f3bafd0f2b6baca0f57bb38b6999f2f/.github/workflows/rust.yml#L1) you use the workflow name and then you use the job names as you did.

.mergify.yml
```diff
@@ -22,4 +27,11 @@ pull_request_rules:
     actions:
         comment:
           message: This pull request is now in conflicts. Could you fix it @{{author}}? 🙏
+        label:
+          add:
+            - s:revision-needed
+          remove:
+            - s:accepted
+            - s:in-progress
+            - s:review-needed
```
Chralt98, line 36: Interesting! Does this ensure it's detects only the last commit of a PR that fails? Otherwise I could imagine a previous workflow fails although you fixed it already with a new commit and it always jumps to "revision needed".

Chralt98 (approved):

.mergify.yml
Chralt98: Except Coverage has a new workflow [here](https://github.com/zeitgeistpm/zeitgeist/blob/bac24e6f2f3bafd0f2b6baca0f57bb38b6999f2f/.github/workflows/coverage.yml#L1).

.mergify.yml
sea212: AFAIK mergify always considers the latest workflow runs, but I guess we'll have to see in practice if this holds true :)

PR #1043, opened 2023-07-17
Run workflows required for merges when label s:review-needed is set

sea212:
### What does it do?
It only executed the `Rust` and `Coverage` workflows when the label `s:review-needed` is set. This will save huge amounts of workflow execution time and cost.
This will require proper state label handling. Should a PR review demand changes, the state should be set to `s:revision-needed`. Before the changes are implemented, the state should be transitioned to `s:in-progress`. Only after the PR is ready for review again, the state `s:review-needed` should be set. This effectively executes the `Rust` and `Coverage` workflows again.

https://github.com/zeitgeistpm/zeitgeist/pull/1042 will change the mergify config in a way that also ensures that `s:revision-needed` is automatically set should the `Rust` or `Coverage` workflow not succeed.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,9 +1,9 @@
 name: Coverage
 
 on:
-  push:
-    branches: [ main ]
   pull_request:
+    types: [ labeled ]
+  push:
     branches: [ main ]
 
 env:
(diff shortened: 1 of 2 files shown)


sea212:
Checked: No review label set, all checks are skipped.

Chralt98 (approved):

PR #1044, opened 2023-07-17
Don't notify codeowners when only weight files or Cargo.toml are changed

sea212:
### What does it do?
Codeowners are not notified if only the weight files and/or Cargo.toml file are changed.

The codeowners file is parsed backwards and stops on the first result. If a rule excludes codeowners (by not specifying any code owners after a path), the parsing stops for that file.

### References
- https://stackoverflow.com/a/67439426/10273801

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,7 +5,7 @@
 
 /node/ @sea212
 /runtime/ @sea212
-# Ignore everything but mod.rs in /runtime/src/weights/
+# Ignore everything but mod.rs in /runtime/common/src/weights/
 /runtime/common/src/weights/*
 /runtime/common/src/weights/mod.rs @sea212
 /zrml/authorized/ @Chralt98
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,8 +6,8 @@
 /node/ @sea212
 /runtime/ @sea212
 # Ignore everything but mod.rs in /runtime/src/weights/
-/runtime/src/weights/*
-/runtime/src/weights/mod.rs @sea212
+/runtime/common/src/weights/*
+/runtime/common/src/weights/mod.rs @sea212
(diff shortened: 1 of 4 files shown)


sea212:
@Mergifyio refresh

CODEOWNERS
```diff
@@ -5,6 +5,9 @@
 
 /node/ @sea212
 /runtime/ @sea212
+# Ignore everything but mod.rs in /runtime/src/weights/
+/runtime/src/weights/*
+/runtime/src/weights/mod.rs @sea212
```
Chralt98, line 10:
```suggestion
/runtime/common/src/weights/*
/runtime/common/src/weights/mod.rs @sea212
```

CODEOWNERS
```diff
@@ -5,6 +5,9 @@
 
 /node/ @sea212
 /runtime/ @sea212
+# Ignore everything but mod.rs in /runtime/src/weights/
```
Chralt98, line 8:
```suggestion
# Ignore everything but mod.rs in /runtime/common/src/weights/
```

Chralt98 (changes requested):

CODEOWNERS
```diff
@@ -5,6 +5,9 @@
 
 /node/ @sea212
 /runtime/ @sea212
```
Chralt98, line 7: Doesn't this conflict with `/runtime/src/weights/*`?

sea212: The last matching rule in the CODEROWNERS file is always applied (see reference in the PR description).
In this case, it will:

- Accept runtime/common/src/mod.rs
- Decline everything else in runtime/common/src/mod.rs
- Accept everything remaining in runtime

Chralt98: I see now.

Chralt98 (approved):

PR #1046, opened 2023-07-20
Update crate and runtime version

sea212:
### What does it do?
- Update crate and runtime version
- Also removes unused code and modifies changelog layout for v0.3.10


sea212:
Selected wrong source branch.

PR #1047, opened 2023-07-20
Update versions

sea212:
### What does it do?
- Update crate and runtime version
- Also removes unused code and modifies changelog layout for v0.3.10

### DIFF
--- a/docs/changelog_for_devs.md
+++ b/docs/changelog_for_devs.md
@@ -12,9 +12,11 @@ https://keepachangelog.com/en/1.0.0/ and ⚠️ marks changes that might break
 components which query the chain's storage, the extrinsics or the runtime
 APIs/RPC interface.
 
-## v0.4.0
+## v0.3.10
 
 [#1022]: https://github.com/zeitgeistpm/zeitgeist/pull/1022
+
+### Added
 
 - Use pallet-asset-tx-payment for allowing to pay transaction fees in foreign
   currencies ([#1022]). This requires each transaction to specify the fee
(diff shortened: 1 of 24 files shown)


docs/changelog_for_devs.md
```diff
@@ -12,10 +12,12 @@ https://keepachangelog.com/en/1.0.0/ and ⚠️ marks changes that might break
 components which query the chain's storage, the extrinsics or the runtime
 APIs/RPC interface.
 
-## v0.4.0
+## v0.3.10
```
Chralt98, line 15: What is the threshold to exceed to come to `0.4.0`? Major changes that are defined how?

Chralt98 (approved):

docs/changelog_for_devs.md
sea212: We take https://semver.org/ as a guideline. 
> Given a version number MAJOR.MINOR.PATCH, increment the:
> [...]
> MINOR version when you add functionality in a backward compatible manner

Obviously we added some new functionality here, this happens every release. The functionality that we added usually assists in using the protocol or changes existing core functionality. I consider new functionality as additions to the core protocol. For example, Court, AMM 2.0, Liquidity Mining, etc. would lead to a bump in the MINOR version from my perspective.

sea212: Maybe AMM 2.0 turns out to be backwards incompatible, in that case a MAJOR version bump would be necessary. However, Malte informed me that before `1.0.0`, a MINOR bump can be done instead as defined in semver.

PR #1049, opened 2023-07-21
Support all fractional places

sea212:
### What does it do?
It adjusts all foreign token representations to the base `BASE`, currently `10^10`. More description TBD once ready for review.

### References
- [Proposal: Support for tokens with arbitrary fractional decimal places](https://hackmd.io/@lZVVinJVS6WI4IpRgmbsLA/Hyz0dJXFn)

### DIFF
--- a/runtime/battery-station/src/xcm_config/config.rs
+++ b/runtime/battery-station/src/xcm_config/config.rs
@@ -170,15 +170,29 @@ pub struct<AssetRegistry, TransactAssetDelegate> AlignedFractionalTransactAsset 
 
 impl<AssetRegistry: Inspect, TransactAssetDelegate: TransactAsset> TransactAsset for AlignedFractionalTransactAsset
 {
-	fn deposit_asset(asset: &MultiAsset, location: &MultiLocation, _context: &XcmContext) -> Result {
-        let decimals = AssetRegistry::metadata_by_location(location)
-            .ok_or_else(|| XcmError::FailedToTransactAsset(e.into()))?
-            .decimals
+	fn deposit_asset(&self, asset: &mut MultiAsset, location: &MultiLocation, context: &XcmContext) -> Result {
+        if let Fungible(amount) = asset.fun {
+            let decimals = AssetRegistry::metadata_by_location(location)
+                .ok_or_else(|| XcmError::FailedToTransactAsset(e.into()))?
+                .decimals
 
-        // TODO adjust places
+            if decimals > self.frac_dec_places {
+                let power = decimals.saturating_sub(self.frac_dec_places)
+                let adjust_factor = 10.saturating_pow(power)
+                // Floors the adjusted token amount, thus no tokens are generated
+                amount.saturating_div(adjust_factor)
+            } else {
+                let power = self.frac_dec_places.saturating_sub(decimals)
+                let adjust_factor = 10.saturating_pow(power)
+                amount.saturating_mul(adjust_factor)
+            };
+        }
+
+        TransactAssetDelegate::deposit_asset(asset, location, context)
 	}
 
 	fn withdraw_asset(
+        &self, 
 		asset: &MultiAsset,
 		location: &MultiLocation,
 		_maybe_context: Option<&XcmContext>,
@@ -187,6 +201,7 @@ impl<AssetRegistry: Inspect, TransactAssetDelegate: TransactAsset> TransactAsset
 	}
 
 	fn transfer_asset(
+        &self, 
 		asset: &MultiAsset,
 		from: &MultiLocation,
 		to: &MultiLocation,
--- a/runtime/battery-station/src/xcm_config/config.rs
+++ b/runtime/battery-station/src/xcm_config/config.rs
@@ -24,7 +24,7 @@ use crate::{
 
 use frame_support::{parameter_types, traits::Everything, WeakBoundedVec};
 use orml_asset_registry::{AssetRegistryTrader, FixedRateAssetRegistryTrader};
-use orml_traits::{location::AbsoluteReserveProvider, MultiCurrency};
+use orml_traits::{asset_registry::Inspect, location::AbsoluteReserveProvider, MultiCurrency};
 use orml_xcm_support::{
     DepositToAlternative, IsNativeConcrete, MultiCurrencyAdapter, MultiNativeAsset,
 };
@@ -156,6 +156,44 @@ parameter_types! {
         ).into(),
         native_per_second(),
     );
+}
+
+/// A generic warpper around implementations of the (xcm-executor) `TransactAsset` trait.
+///
+/// Aligns the fractional decimal places of every incoming token.
+/// Reconstructs the original number of fractional decimal places of every outgoing token.
+#[allow(clippy::type_complexity)]
+pub struct<AssetRegistry, TransactAssetDelegate> AlignedFractionalTransactAsset {
+    /// Number of fractional decimal places
+    frac_dec_places: u128,
+}
+
+impl<AssetRegistry: Inspect, TransactAssetDelegate: TransactAsset> TransactAsset for AlignedFractionalTransactAsset
+{
+	fn deposit_asset(asset: &MultiAsset, location: &MultiLocation, _context: &XcmContext) -> Result {
+        let decimals = AssetRegistry::metadata_by_location(location)
+            .ok_or_else(|| XcmError::FailedToTransactAsset(e.into()))?
+            .decimals
+
+        // TODO adjust places
+	}
+
+	fn withdraw_asset(
+		asset: &MultiAsset,
+		location: &MultiLocation,
+		_maybe_context: Option<&XcmContext>,
+	) -> result::Result<Assets, XcmError> {
+        // TODO adjust places
+	}
+
+	fn transfer_asset(
+		asset: &MultiAsset,
+		from: &MultiLocation,
+		to: &MultiLocation,
+		_context: &XcmContext,
+	) -> result::Result<Assets, XcmError> {
+        // TODO ???
+	}
 }
 
 /// Means for transacting assets on this chain.


runtime/zeitgeist/src/xcm_config/fees.rs
```diff
@@ -68,13 +63,22 @@ impl<
...
-        if let Some(fee_factor) = metadata.additional.xcm.fee_factor {
-            let base = 10u128.checked_pow(metadata.decimals)?;
-            bmul(default_per_second, fee_factor, base)
+        if native_decimals > foreign_decimals {
+            let power = native_decimals - foreign_decimals;
+            Some(fee_unadjusted.checked_div(10u128.checked_pow(power)?)?)
         } else {
-            Some(default_per_second)
+            let power = foreign_decimals - native_decimals;
+            Some(fee_unadjusted.checked_mul(10u128.checked_pow(power)?)?)
```
Chralt98, line 81: I try to understand. We have cut the decimal places of KSM for example by 10^2 previously. Then we use `default_per_second` multiplied by `fee_factor` (on a native base) to translate the value of ZTG into KSM terms (on a base of ten). And now we need to get back the fee of KSM on a base of twelve for properly charging the KSM XCM fee per second.

Can you give me an example for calculating a `fee_factor` (stored in the native base) for KSM? Step-by-step would be nice. I just try to understand if we not already adjusted the fee enough by using the conversion via the `fee_factor`.

runtime/battery-station/src/xcm_config/fees.rs
```diff
@@ -68,13 +63,22 @@ impl<
...
+            bmul(default_per_second, fee_factor).ok()?
+        } else {
+            default_per_second
+        };
 
-        if let Some(fee_factor) = metadata.additional.xcm.fee_factor {
-            let base = 10u128.checked_pow(metadata.decimals)?;
-            bmul(default_per_second, fee_factor, base)
+        if native_decimals > foreign_decimals {
+            let power = native_decimals - foreign_decimals;
```
Chralt98, line 77:
```suggestion
            let power = native_decimals.saturating_sub(foreign_decimals);
```

runtime/battery-station/src/xcm_config/fees.rs
```diff
@@ -68,13 +63,22 @@ impl<
...
+            Some(fee_unadjusted.checked_div(10u128.checked_pow(power)?)?)
         } else {
-            Some(default_per_second)
+            let power = foreign_decimals - native_decimals;
```
Chralt98, line 80:
```suggestion
            let power = foreign_decimals.saturating_sub(native_decimals);
```

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
```diff
@@ -172,6 +172,242 @@ fn transfer_ztg_sibling_to_zeitgeist() {
...
+            initial_sovereign_balance - transfer_amount
+        );
+    });
+}
+
+#[test]
+fn transfer_eth_sibling_to_zeitgeist() {
+    TestNet::reset();
+
+    let sibling_alice_initial_balance = ztg(10) + eth(1);
```
Chralt98, line 294: What is the point of adding those two different amounts?

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
```diff
@@ -172,6 +172,242 @@ fn transfer_ztg_sibling_to_zeitgeist() {
...
+
+    Sibling::execute_with(|| {
+        // Set the sovereign balance such that it is not subject to dust collection
+        assert_ok!(Balances::set_balance(
+            RuntimeOrigin::root(),
+            zeitgeist_parachain_account().into(),
+            initial_sovereign_balance,
+            0
+        ));
+        // Add 1 "fake" ETH to Alice's balance
```
Chralt98, line 313: 1 fake ETH and 10^10 amount (`ztg(10)`)?

Chralt98 (commented):
It is an interesting approach to handle foreign assets easier on our chain. Especially that you calculated the value of falling precision for having more foreign decimals places than our native base of ten.

runtime/zeitgeist/src/xcm_config/config.rs
```diff
@@ -158,6 +163,108 @@ parameter_types! {
...
+                let metadata = AssetRegistry::metadata(asset_id);
+
+                if let Some(metadata) = metadata {
+                    let decimals = metadata.decimals;
+
+                    asset_updated.fun = if decimals > native_decimals {
+                        let power = decimals.saturating_sub(native_decimals);
+                        let adjust_factor = 10u128.saturating_pow(power);
+                        // Floors the adjusted token amount, thus no tokens are generated
+                        Fungible(amount.saturating_div(adjust_factor))
```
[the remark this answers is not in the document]
sea212: Any XCM uses the global canonical representation, i.e. 18 fractional decimal places for ETH. 

Simplified:

deposit: When a reserve asset transfer with ETH is incoming, it puts the unadjusted ETH in the holding register. Then it deposits the adjusted ETH using the adjusted `deposit` function into the target pallet and account. Then it subtracts the unadjusted value from the holding register.

withdraw: When a reserve asset transfer with ETH is sent, it withdraws the adjusted ETH using the adjusted `withdraw` function from the target pallet and acount and afterwards puts the unadjusted ETH in the holding register (since the XCM contains the global canonical representation, in this case 18 fractional decimal places).

deposit: XCM in (unadjusted ETH) -> holding register (unadjusted) -> tokens (deposit, adjusted) -> update holding register (unadjusted)
withdraw: XCM out (unadjusted ETH) -> tokens (withdraw, adjusted) -> holding register (unadjusted)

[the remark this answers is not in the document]
sea212: The trick here is that we demand the global canonical representation for any token. Simplified, if 1 ETH enters, the XCM contains the amount `10^18`. If an XCM is created on Zeitgeist to send ETH, the XCM contains the amount `10^18` as value, although internally it is stored as `10^10`.

runtime/zeitgeist/src/xcm_config/fees.rs
sea212: The function is applied on the holding register, where the adjustment does not happen. For example:
1 KSM enters and is put into the holding register, holding register has value `10^12`. `fee_per_second` is caluclated by
1. fetching the default fee per second of execution time, which uses the native base representation `10^10`
2. multiplied by `fee_factor`, which uses native base representation  `10^10` (for any token)
3. adjusted to match KSM representation in the holding register, in that case `fee_per_second * 10^2`

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
sea212: Yes. Technically Alice already has 10 ztg from genesis.

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
sea212: The ztg is used to pay native fees (although Alice should already have 10 ZTG from Genesis).
`eth` is just used for transparency, in the end it is stored as native ZTG but treated like eth on the chain that receives the XCM.

runtime/zeitgeist/src/xcm_config/config.rs
```diff
@@ -158,6 +163,108 @@ parameter_types! {
...
+                    } else {
+                        let power = native_decimals.saturating_sub(decimals);
+                        let adjust_factor = 10u128.saturating_pow(power);
+                        Fungible(amount.saturating_mul(adjust_factor))
```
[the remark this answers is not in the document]
sea212: see https://github.com/zeitgeistpm/zeitgeist/pull/1049#discussion_r1279219324

runtime/zeitgeist/src/xcm_config/config.rs
```diff
@@ -158,6 +163,108 @@ parameter_types! {
...
+        let asset_adjusted = Self::adjust_fractional_places(asset);
+        TransactAssetDelegate::withdraw_asset(&asset_adjusted, location)
+    }
+
+    fn transfer_asset(
+        asset: &MultiAsset,
+        from: &MultiLocation,
+        to: &MultiLocation,
+    ) -> Result<Assets, XcmError> {
+        let asset_adjusted = Self::adjust_fractional_places(asset);
```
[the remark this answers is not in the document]
sea212: `TransferAsset` is used to transfer an asset on a remote destination from the sending account **to a receiving** account on the same chain, like in  [this](https://github.com/paritytech/polkadot/blob/35bd316e350230026cd5058b865431b0a85e6867/xcm/xcm-simulator/example/src/lib.rs#L346-L350) example (sent from sending account into parachain account). Messages that contain a location outside of Zeitgeist will fail in execution.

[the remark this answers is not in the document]
Chralt98: Ok, I think I see now. Means other parachains can transfer assets between Zeitgeist accounts.

runtime/zeitgeist/src/xcm_config/config.rs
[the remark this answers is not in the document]
Chralt98: I see. So the point that I missed is that any XCM message either `withdraw` or `deposit` is (hopefully) always constructed with the external decimal representation (not our internal ten base representation). 

I also missed the point in general that your `AlignedFractionalTransactAsset` is only meant for the internal representation to get data from the XCM into the tokens pallet. `AlignedFractionalTransactAsset` is not meant to get the asset amount out to the external representation, since the external representation is always the value inside the XCM.

It reminds me about a one-way hash function into the internal representation only.

Chralt98 (approved):

maltekliemann (approved):

docs/changelog_for_devs.md
```diff
@@ -12,6 +12,23 @@ https://keepachangelog.com/en/1.0.0/ and ⚠️ marks changes that might break
...
+## v0.3.11
+
+[#1049]: https://github.com/zeitgeistpm/zeitgeist/pull/1049
+
+### Changed
+
+- ⚠️ All tokens now use 10 fractional decimal places.
+  - cross-consensus messages (XCM) assume the global canonical representation for token balances.
+- The token metadata in the asset registry now assumes that the existential deposit and fee factor
+  are stored in base 10,000,000,000.
```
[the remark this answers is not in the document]
sea212: The alignment of line 22 is intentional, it is a sub point of the one in line 21.

PR #1050, opened 2023-07-24
Run try-runtime also on RC tags

sea212:
### What does it do?
This PR adjusts the "tag" workflow to make it trigger on release candidate tags as well. It also renames the workflow to "try-runtime".

Currently the "tag" workflow is only executed when a final release tag (i.e. `v0.3.10`) is pushed. Before the final release tag is created, release candidates are pushed (i.e. `v0.3.10-rc1`). The workflow should also execute on those release candidates, granting the maintainer the knowledge whether or not this precondition for a release is met.

### References
- [Regex tester](https://regexr.com/)


Chralt98 (approved):
> It also renames the workflow to "try-runtime".

I can't find this change in the diff.

LGTM

sea212:
> > It also renames the workflow to "try-runtime".
> 
> I can't find this change in the diff.

It's indicated at the file name: 
image

PR #1056, opened 2023-07-30
Foreign tx fee improvements

sea212:
### What does it do?
1. Retrieves the asset metadata using the `asset_id` (which is always a mandatory field) instead of the `location` (which is an optional field).
2. Moves parachain tests into a parachain module
3. Formats the macro code

### DIFF
--- a/runtime/common/src/fees.rs
+++ b/runtime/common/src/fees.rs
@@ -69,14 +69,16 @@ macro_rules! impl_foreign_fees {
             traits::{
                 fungibles::CreditOf,
                 tokens::{
-                    fungibles::Balanced,
-                    BalanceConversion, WithdrawConsequence, WithdrawReasons,
+                    fungibles::Balanced, BalanceConversion, WithdrawConsequence, WithdrawReasons,
                 },
                 ExistenceRequirement,
             },
             unsigned::TransactionValidityError,
         };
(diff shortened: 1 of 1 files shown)


runtime/common/src/fees.rs
```diff
@@ -120,16 +122,11 @@ macro_rules! impl_foreign_fees {
...
         ) -> Result<Balance, TransactionValidityError> {
-            let location = AssetConvert::convert(asset_id);
-            let metadata = location
-                .and_then(|loc| {
-                    <AssetRegistry as orml_traits::asset_registry::Inspect>::metadata_by_location(
-                        &loc,
-                    )
-                })
-                .ok_or(TransactionValidityError::Invalid(InvalidTransaction::Custom(
+            let metadata = <AssetRegistry as Inspect>::metadata(&asset_id).ok_or(
```
Chralt98, line 125: Nice! It really hurts that I see the simpler API call right next to the previously used :D https://github.com/open-web3-stack/open-runtime-module-library/blob/73b4ee04e402f240f270bfe83f59fc2837dbaf51/asset-registry/src/impls.rs#L185-L191 .. I really don't know why I shot myself in the foot.

Chralt98 (approved):

runtime/common/src/fees.rs
sea212: I know right, asked myself the same thing in respect to my review.

PR #1058, opened 2023-07-31
Use u32 asset id for foreign tx fee payment

Chralt98:
### What does it do?

This allows to integrate the front end extension and wallet properly. We use the advantage that the Polkadot asset hub relies on `u32` asset ids and we can therefore take this data structure too.

I couldn't just use the simple [FungiblesAdapter](https://github.com/paritytech/substrate/blob/eb80c4e3b81db9c0a2312dadcc20e007f9aa9c00/frame/transaction-payment/asset-tx-payment/src/payment.rs#L94-L174) anymore, because the asset id needed to be converted to `Asset::ForeignAsset(u32)` just before the `withdraw` call. Otherwise the `Tokens` wouldn't have known how to handle an asset id with just the `u32` type. 

Before that we used the custom Asset (CurrencyId), but since this does not properly integrate with the extension and wallet, we fall back on a mapping from `u32` to `Asset::ForeignAsset(u32)` internally. This also removes the requirement to define custom `signedExtensions` on the front end side, since we can use the `assetId` which is already in place for the wallets and Polkadot browser extension. 

Rewriting front end side code. We didn't really want to fork our own extensions and wallets..

### DIFF
--- a/runtime/common/src/fees.rs
+++ b/runtime/common/src/fees.rs
@@ -67,7 +67,7 @@ macro_rules! impl_foreign_fees {
         use frame_support::{
             pallet_prelude::InvalidTransaction,
             traits::{
-                fungibles::CreditOf,
+                fungibles::{CreditOf, Inspect},
                 tokens::{
                     fungibles::Balanced, BalanceConversion, WithdrawConsequence, WithdrawReasons,
                 },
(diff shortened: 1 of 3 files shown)


runtime/common/src/fees.rs
```diff
@@ -179,6 +164,76 @@ macro_rules! impl_foreign_fees {
...
+            type Balance = Balance;
+            type LiquidityInfo = CreditOf<AccountId, Tokens>;
+
+            fn withdraw_fee(
+                who: &AccountId,
+                call: &RuntimeCall,
+                _dispatch_info: &DispatchInfoOf<RuntimeCall>,
+                asset_id: Self::AssetId,
+                native_fee: Self::Balance,
+                _tip: Self::Balance,
```
sea212, line 180: Why is it fine to ignore the tip here?

sea212 (approved):

PR #1061, opened 2023-08-01
Fix workflow runs on main

sea212:
### What does it do?
It ensures that all conditioned jobs execute also on main when appropriate.


Chralt98:
just wanted to ensure that review-needed trigger works

PR #1062, opened 2023-08-03
Update versions

sea212:
### What does it do?
It updates the crate versions and the `RuntimeVersion`.


Chralt98 (approved):

sea212:
Forked of from bad base branch.

PR #1065, opened 2023-08-04
 Fix try-runtime makefile targets

sea212:
### What does it do?
Adjust the try-runtime targets in the Makefile to properly invoke try-runtime.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,18 +22,26 @@ check-dummy:
 		--bin=zeitgeist \
 		--features=parachain,try-runtime \
 		try-runtime \
-		--execution=Native \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,18 +22,26 @@ check-dummy:
 		--bin=zeitgeist \
 		--features=parachain,try-runtime \
 		try-runtime \
-		--execution=Native \
```
Chralt98, line 25: I think a good reason for Native execution is the ability to read errors instead of getting a wasm strapped message. Do we really need to use wasm execution to get try-runtime working?

Chralt98 (approved):

PR #1066, opened 2023-08-04
Fix parachain-staking benchmarks

sea212:
### What does it do?
Two errors hindered the successful execution of the benchmarks within `pallet-parachain-staking`.
1. `MinSelectedCandiates = 8` for battery station runtime, which is used in `dev` parachain, is less than `num_selected_candidates` in `dev` parachain's `GenesisConfig`. However, the condition `num_selected_candidates >= MinSelectedCandidates` must be met.
2. https://github.com/moonbeam-foundation/moonbeam/pull/2227

This PR fixes 1. and updated dependencies from the Moonbeam branch that incorporates 2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5033,7 +5033,7 @@ dependencies = [
 [[package]]
 name = "moonbeam-vrf"
 version = "0.1.0"
-source = "git+https://github.com/zeitgeistpm/moonbeam?branch=polkadot-v0.9.37-use-paritytech-dependencies#f869fd30031eb55f4a89ac261b36f639bd20473b"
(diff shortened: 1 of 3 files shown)


Chralt98 (approved):
The obvious question: I think you did initially use `num_selected_candidates = 1`, because if we use `cargo build` and then `./target/debug/zeitgeist --dev --tmp` we are usually having only one node instance running. So hopefully the dev configuration still produces blocks.

sea212:
> The obvious question: I think you did initially use `num_selected_candidates = 1`, because if we use `cargo build` and then `./target/debug/zeitgeist --dev --tmp` we are usually having only one node instance running. So hopefully the dev configuration still produces blocks.

We use the standalone client for one node environments, this parameter only affects the parachain client though.

PR #1067, opened 2023-08-04
Integrate the Sygma bridge

Chralt98:
### What does it do?

Integrate the Sygma pallets in order to transact assets registered at the Sygma bridge. By the end of Q3 Sygma will connect to stablecoins like `USDC` to eventually bring those to our prediction markets. 

Our current main branch needs `polkadot-v0.9.32`, but [`sygma-substrate-pallets`](https://github.com/sygmaprotocol/sygma-substrate-pallets/) does not have this version step. But the sygma pallets are available for `polkadot-v0.9.37` as our current polkadot maintenance PR is.

Register the Sygma assets separately.

The alternative was to transact over Phala via XCM to Sygma assets. This would need us to trust one centralised point e.g. Phala.


tolak:
Hey @Chralt98, just wanna mention that instead of integrating Sygma into your runtime directly, another option is to use SubBridge to forward the transfer to Zeitgeist through the XCM protocol. See [here](https://docs.phala.network/other-products/subbridge/technical-details#the-lifecycle-of-cross-chain-transaction).

Chralt98:
Hey @tolak, our required use-case is to bring stablecoins to Zeitgeist. Is this possible via XCMs over to Phala? If yes, how exactly could this be done (can you provide a XCM message example)?

sea212:
This PR as abandoned:
1. Integration is on hold
2. If integration happens, it will happen around end of this year
3. Alternative solutions might be applied until then

PR #1069, opened 2023-08-06
Fix fee tests

sea212:
### What does it do?
tbd


sea212:
Forked off wrong base

PR #1078, opened 2023-08-19
Adjust maximum market lifetime

maltekliemann:
### What does it do?

Changes the maximum market lifetime to 4 years.

### DIFF
--- a/primitives/src/constants.rs
+++ b/primitives/src/constants.rs
@@ -82,7 +82,7 @@ pub const LM_PALLET_ID: PalletId = PalletId(*b"zge/lymg");
 
 // Prediction Markets
 /// The maximum allowed market life time, measured in blocks.
-pub const MAX_MARKET_LIFETIME: BlockNumber = 365 * BLOCKS_PER_DAY;
+pub const MAX_MARKET_LIFETIME: BlockNumber = 4 * BLOCKS_PER_YEAR;
 /// Max. categories in a prediction market.
 pub const MAX_CATEGORIES: u16 = 64;
 /// The dispute_duration is time where users can dispute the outcome.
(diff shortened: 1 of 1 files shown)


primitives/src/constants.rs
```diff
@@ -82,7 +82,7 @@ pub const LM_PALLET_ID: PalletId = PalletId(*b"zge/lymg");
 
 // Prediction Markets
 /// The maximum allowed market life time, measured in blocks.
-pub const MAX_MARKET_LIFETIME: BlockNumber = 365 * BLOCKS_PER_DAY;
+pub const MAX_MARKET_LIFETIME: BlockNumber = 4 * BLOCKS_PER_YEAR;
```
sea212, line 85: Why do we need that?

maltekliemann: The AC has repeatedly been blocked from creating markets on the U.S. elections by the current one year limit. Virtually everybody else has markets on these topics already. This is unacceptable in my view.

A number of other topics come to mind that we've been meaning to create markets on:

- Climate (these often need to run for _at least_ a year)
- Geopolitical conflicts and macro trends
- AI adoption (happens on a scale larger than a year)
- (Less important) Archeological discoveries, contact with extraterrestrials, etc.

I think at least in some of these cases the argument that you can split these markets up into multiple small markets doesn't really matter. How do you split up a market on the U.S. presidential elections and how does that help the people who are just interested in who's going to be president?

We also need to keep combinatorial market design in mind here: If the welfare metric is measured at the end of a "cycle" (the election, for instance, with the welfare metric being the popular vote), but decisions are made much soon (like what candidate to run with), the market will require a longer duration.

I still fully believe that shorter markets generally make more sense a provide more precision per time, but that argument alone is not strong enough to put a straightjacket on the AC in what they can do.

Of course, technical debt is also a concern here. If, for some reason, we need to move to an entirely new market sturcture, we may not be able to migrate the old markets and be stuck with them for four years instead of one. I think this is a minor concern. If we ever fundamentally rewrite the logic so that it's actually grossly incompatible with the old version, maybe that's a problem onto itself.

sea212 (approved):

PR #1100, opened 2023-09-11
Implement trusted markets

maltekliemann:
### What does it do?

Makes the `dispute_mechanism` field of `Market` optional. If it's `None`, then the market is _trusted_, which means that the following rules apply:

- The market's `DisputePeriod` **must** be zero.
- The market is resolved immediately after reporting.
- The market may only be reported by the oracle, not by an outsider.

Adding `Trusted` to `MarketDisputeMechanism` is a bad idea and leads to even worse design that this PR.


zrml/prediction-markets/src/weights.rs
```diff
@@ -415,11 +416,14 @@ impl<T: frame_system::Config> WeightInfoZeitgeist for WeightInfo<T> {
...
     // Storage: Timestamp Now (r:1 w:0)
     // Storage: PredictionMarkets MarketIdsPerReportBlock (r:1 w:1)
-    fn report(_m: u32) -> Weight {
+    fn report_market_with_dispute_mechanism(_m: u32) -> Weight {
         Weight::from_ref_time(69_185_134)
             .saturating_add(T::DbWeight::get().reads(3))
             .saturating_add(T::DbWeight::get().writes(2))
     }
+    fn report_trusted_market() -> Weight {
+        Weight::from_ref_time(123)
```
Chralt98, line 425: I mean there is at least one read and write for the market. What is happening here? This can't be the real weight.

zrml/prediction-markets/src/migrations.rs
Chralt98, line 1: I think the last migration of the dispute bond addition will be already run beforehand. So you could start with a fresh migration and don't have to edit the old. AFAIK Harald will use a commit before your feature here for the release which contains court and zip-0 part II.

Chralt98 (changes requested):
Exactly the changes that I also would have expected for this feature. Well done!

I wouldn't call this feature trusted markets, but rather non-disputable markets (NDMs).

zrml/prediction-markets/src/lib.rs
```diff
@@ -2396,19 +2350,27 @@ mod pallet {
...
-            );
-            ensure!(
-                deadlines.dispute_duration <= T::MaxDisputeDuration::get(),
-                Error::<T>::DisputeDurationGreaterThanMaxDisputeDuration
-            );
+            if trusted {
+                ensure!(
+                    deadlines.dispute_duration == Zero::zero(),
+                    Error::<T>::NonZeroDisputePeriodOnTrustedMarket
+                );
```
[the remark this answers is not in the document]
maltekliemann: The problem here is that dispute durations are not part of the dispute mechanism itself. This allows us to create markets with nonsense states like `dispute_mechanism = None` and `dispute_duration > 0`. But this is not something that should be fixed as part of this PR.

zrml/prediction-markets/src/weights.rs
maltekliemann: Yeah, I didn't run any benchmarks.

zrml/prediction-markets/src/lib.rs
```diff
@@ -3120,6 +3086,108 @@ mod pallet {
...
+                let sender_is_oracle = sender == market.oracle;
+                let origin_has_permission = T::ResolveOrigin::ensure_origin(origin).is_ok();
+                ensure!(sender_is_oracle || origin_has_permission, Error::<T>::ReporterNotOracle);
+                market.report = Some(market_report.clone());
+                market.status = MarketStatus::Reported;
+                Ok(())
+            })?;
+            let market = <zrml_market_commons::Pallet<T>>::market(&market_id)?;
+            Self::resolve(&market_id, &market)?;
+            Ok(Some(T::WeightInfo::report_trusted_market()).into()) // TODO
```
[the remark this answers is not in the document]
maltekliemann: 3ebc3aa66dfd5cb80970b04ea178a40447c572d7

zrml/prediction-markets/src/lib.rs
```diff
@@ -3120,6 +3086,108 @@ mod pallet {
...
+                let sender = ensure_signed(origin.clone())?;
+                let sender_is_oracle = sender == market.oracle;
+                let origin_has_permission = T::ResolveOrigin::ensure_origin(origin).is_ok();
+                ensure!(sender_is_oracle || origin_has_permission, Error::<T>::ReporterNotOracle);
+                market.report = Some(market_report.clone());
+                market.status = MarketStatus::Reported;
+                Ok(())
+            })?;
+            let market = <zrml_market_commons::Pallet<T>>::market(&market_id)?;
+            Self::resolve(&market_id, &market)?;
```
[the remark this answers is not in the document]
maltekliemann: 89211038df37b077266175ae79b1fec99f784e25

zrml/prediction-markets/src/lib.rs
```diff
@@ -2221,6 +2167,12 @@ mod pallet {
         /// Clears this market from being stored for automatic resolution.
         fn clear_auto_resolve(market_id: &MarketIdOf<T>) -> Result<(u32, u32), DispatchError> {
             let market = <zrml_market_commons::Pallet<T>>::market(market_id)?;
+            // If there's no dispute mechanism, this function is noop. FIXME This is an
+            // anti-pattern, but it makes benchmarking easier.
+            let dispute_mechanism = match market.dispute_mechanism {
+                Some(ref result) => result,
+                None => return Ok((0, 0)),
+            };
```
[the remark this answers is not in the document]
maltekliemann: This won't work because the `return` statement only applies to the closure, not the entire function. I don't think Rust has a one-line solution for this.

Chralt98 (approved):

runtime/zeitgeist/src/lib.rs
```diff
@@ -174,10 +174,10 @@ impl Contains<RuntimeCall> for IsCallable {
                     create_market { scoring_rule: RikiddoSigmoidFeeMarketEma, .. } => false,
                     edit_market { scoring_rule: RikiddoSigmoidFeeMarketEma, .. } => false,
                     // Disable Court & SimpleDisputes dispute resolution mechanism
-                    create_market { dispute_mechanism: Court | SimpleDisputes, .. } => false,
-                    edit_market { dispute_mechanism: Court | SimpleDisputes, .. } => false,
+                    create_market { dispute_mechanism: Some(Court | SimpleDisputes), .. } => false,
+                    edit_market { dispute_mechanism: Some(Court | SimpleDisputes), .. } => false,
```
sea212, line 178: Are you 100% sure that the `|` represents a logical or instead of a binary or in this case? (as enum elements can be represented as numbers).

sea212 (approved):

runtime/zeitgeist/src/lib.rs
maltekliemann: > Are you 100% sure that the `|` represents a logical or instead of a binary or in this case?

Yeah. This is the content of the following RFC: https://rust-lang.github.io/rfcs/2535-or-patterns.html.

> (as enum elements can be represented as numbers)

Not sure what you mean. Sure, variants can be represented as numbers, but I don't think Rust ever does that. I mean, what would `Court | SimpleDisputes` even be? What if the resulting number doesn't represent a variant of the enum?

PR #1101, opened 2023-09-12
Update weights

sea212:
### What does it do?
Updates the weights based on the reference machine.

- Weightsv2 is now used, which also includes storage requirement
- The weight template has changed in the previous maintenance PR #1079


maltekliemann (approved):

Chralt98:
If we use the old weight-v1 system by `from_ref_time`, the nodes won't be able to process those transactions.

PR #1102, opened 2023-09-12
Fix taplo workflow

maltekliemann:
### What does it do?

Fixes the recent compiler issues encountered when trying to run the taplo workflow.


maltekliemann:
Closed because this isn't going anywhere.

PR #1105, opened 2023-09-14
Use self-hosted runner for docker parachain build

sea212:
### What does it do?
See title

- Run fails due to insufficient disk space https://github.com/zeitgeistpm/zeitgeist/actions/runs/6169452668
- In the near future, other self-hosted runners will be available


sea212:
Closed in favor of https://github.com/zeitgeistpm/zeitgeist/pull/1134

PR #1120, opened 2023-09-19
Adapt arbitrage weight calculations to proof sizes

maltekliemann:
### What does it do?

Previously, `execute_arbitrage_all` did the following calculation to estimate how many pools can be arbitrages. Let $w(a)$ be the weight required for arbitraging $a$ pools. We assume that $w$ is a linear function, $w(a) = ma + b$. We therefore have $w(1) - w(0) = m + b - b = m$. Suppose that $W$ is the available weight. Then the maximum number of pools we can arbitrage in the time available is $\frac{W - b}{m}$ (just solve $W = w(a)$ for $a$), ignoring the fact that we need to round down to make $a$ an integer.

But weights now have two components, ref time and proof size. The changes in this PR are the following:

- Introduce a minimum available proof size of 100 KB for `on_idle` to even do the calculation described above. This is definitely an overestimation, but we should be on the safe side.
- Perform the calculation described above both for the ref time and the proof size. This gives us a maximum amount of pools that can be arbitraged using the available ref time and a maximum amount of pools that can be arbitraged using the available proof size. Taking the minimum of both gives us the maximum number of pools we can arbitrage with the available resources.

### DIFF
--- a/zrml/swaps/src/lib.rs
+++ b/zrml/swaps/src/lib.rs
@@ -116,7 +116,7 @@ mod pallet {
     pub(crate) const ARBITRAGE_MAX_ITERATIONS: usize = 30;
     const ARBITRAGE_THRESHOLD: u128 = CENT;
     const MIN_BALANCE: u128 = CENT;
-    const ON_IDLE_MIN_WEIGHT: Weight = Weight::from_ref_time(1_000_000);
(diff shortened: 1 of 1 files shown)


sea212 (approved):
LGTM

PR #1121, opened 2023-09-20
Update weights v0.4.0

sea212:
### What does it do?
- Updates weights

- Should respect max PoV size now due to previous fixes


Chralt98 (approved):
No proof of validity weight conspicuous anymore 👏🏻

maltekliemann (approved):

PR #1127, opened 2023-09-28
Implement close markets early feature

Chralt98:
### What does it do?

It allows the market creator to schedule an early market close. It allows the AC to schedule a fast-track early market close. Additionally the market creator's early market close can be disputed by anyone, who is willing to pay the bond (CloseDisputeBond) required. The schedule of the early market close can also be rejected by the AC.

The decision to close a market early should be done as decentralised as possible. That's why I suggest to use the court system for this decision-making in the future.


maltekliemann (changes requested):
Some general comments:

- The terms _premature_ and _early_ are used interchangeably in this PR. This should be fixed; we should use one of these. Since _premature_ refers to closing a market too soon, it should be _early_. Some of the config variables are just called `CloseSomething`. I strongly suggest using `CloseEarlySomething` instead so it's clear what component they belong to.
- I haven't checked if `admin_destroy_market` cleans everything up properly. I will remove `admin_destroy_market` in a PR in the next couple of days.
- The changes made here don't play well with the grace period. I think that's fine. We should really think about sunsetting the grace period. If we end up not doing that, scheduling early close should extend to the grace period, i.e. a market in grace period should also be "early closeable".
- It might be a good idea to only allow the AC to schedule early closure for now so the UI team doesn't have to implement the dispute process. @Robiquet @sea212 Opinions?
- `slash_pending_bonds` must be updated.

Robiquet:
> * It might be a good idea to only allow the AC to schedule early closure for now so the UI team doesn't have to implement the dispute process. @Robiquet @sea212 Opinions?
Not sure I understand why this changes the need for a UI, can't the AC schedule an early closure anyway?

maltekliemann:
@Chralt98 will correct me if I'm wrong, but market creators are allowed to schedule and early close and any user is allowed to dispute that early close. My thinking was that this logic is too complex for an MVP.

Chralt98:
> but market creators are allowed to schedule and early close and any user is allowed to dispute that early close. My thinking was that this logic is too complex for an MVP.

This is possible as you described with the code. But it's also possible that the AC can schedule an early close themselves without the market creator doing anything.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1477,6 +1478,336 @@ mod pallet {
...
+        ///
+        /// Complexity: `O(n)`, where `n` is the maximum number of market ids
+        /// in `MarketIdsPerClose...` either at the old period end or new period end.
+        #[pallet::call_index(17)]
+        #[pallet::weight(
+            T::WeightInfo::schedule_early_close_as_authority(CacheSize::get(), CacheSize::get())
+                .max(T::WeightInfo::schedule_early_close_after_dispute(CacheSize::get(), CacheSize::get()))
+                .max(T::WeightInfo::schedule_early_close_as_market_creator(CacheSize::get(), CacheSize::get()))
+            )
+        ]
```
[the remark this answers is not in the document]
Chralt98: I suggest you to propose me the correct formatting or we just ignore the formatting in the weight macros. This task seems too pedantic to me. Do you have another approach? I mean the workflow is pretty bad for me to format it correctly..

zrml/prediction-markets/src/lib.rs
```diff
@@ -1477,6 +1478,336 @@ mod pallet {
...
+                        }
+                        MarketPeriod::Timestamp(ref range) => {
+                            let now_time = <zrml_market_commons::Pallet<T>>::now();
+                            now_time < range.end
+                        }
+                    }
+                },
+                "This will never happen, because schedule_early_close ensures that the new end is \
+                 always before the old end, 
+                    otherwise the switch to an old market period would lead to a never ending \
```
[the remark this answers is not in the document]
Chralt98: 705145e83420a072410e5344077abbd8ca6a129e

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
Chralt98: 705145e83420a072410e5344077abbd8ca6a129e

zrml/prediction-markets/src/lib.rs
```diff
@@ -2450,6 +2841,33 @@ mod pallet {
         pub(crate) fn close_market(market_id: &MarketIdOf<T>) -> Result<Weight, DispatchError> {
             <zrml_market_commons::Pallet<T>>::mutate_market(market_id, |market| {
                 ensure!(market.status == MarketStatus::Active, Error::<T>::InvalidMarketStatus);
+
+                match &market.premature_close {
```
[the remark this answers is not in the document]
Chralt98: 00b88cbefb94c05953101c8a97c6d3ae19dde7b2

zrml/prediction-markets/src/lib.rs
```diff
@@ -1477,6 +1478,336 @@ mod pallet {
 
             Ok(Some(T::WeightInfo::start_global_dispute(ids_len_1, ids_len_2)).into())
         }
+
+        /// Allows the advisory committee or the market creator to schedule an early close.
```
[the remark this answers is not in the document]
Chralt98: https://github.com/zeitgeistpm/zeitgeist/blob/175070378e0a99d5e02e475727568f384b308feb/zrml/prediction-markets/src/lib.rs#L499-L503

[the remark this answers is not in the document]
Chralt98: a2b3fe160fd66366d167b50a69346980d9a88acf

zrml/prediction-markets/src/lib.rs
```diff
@@ -1477,6 +1478,336 @@ mod pallet {
...
+            origin: OriginFor<T>,
+            #[pallet::compact] market_id: MarketIdOf<T>,
+        ) -> DispatchResultWithPostInfo {
+            let is_authorized = T::CloseMarketEarlyOrigin::try_origin(origin.clone()).is_ok();
+            if is_authorized {
+                T::CloseMarketEarlyOrigin::ensure_origin(origin.clone())?;
+            } else {
+                // check if market creator below
+                ensure_signed(origin.clone())?;
+            };
```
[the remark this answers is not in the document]
Chralt98: df342e9286cab117f358dc2526485ec3b1280b26

zrml/prediction-markets/src/tests.rs
```diff
@@ -2847,6 +2850,829 @@ fn dispute_reserves_dispute_bond() {
...
+            ScoringRule::CPMM,
+        );
+
+        // just to ensure events are emitted
+        run_blocks(2);
+
+        let market_id = 0;
+
+        assert_ok!(
+            PredictionMarkets::schedule_early_close(RuntimeOrigin::signed(SUDO), market_id,)
```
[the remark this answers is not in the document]
Chralt98: Normally this should be done by cargo format.. But here it is c83872d831daa1a6d0b398163a587981a6c6e47c

docs/changelog_for_devs.md
```diff
@@ -12,6 +12,19 @@ As of 0.3.9, the changelog's format is based on
...
+### Added
+
+- Add extrinsics to zrml-prediction-markets:
+  - `schedule_early_close`: Schedule an early close of a market.
+  - `dispute_early_close`: Dispute a scheduled early close of a market.
+  - `reject_early_close`: Reject a scheduled early close of a market.
+- Add events to zrml-prediction-markets:
+  - `MarketEarlyCloseScheduled`
+  - `MarketEarlyCloseDisputed`
+  - `MarketEarlyCloseRejected`
```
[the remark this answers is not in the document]
Chralt98: f43dbbaf8100779fb977430d2674d0addc871ef1

zrml/prediction-markets/src/lib.rs
```diff
@@ -1477,6 +1478,336 @@ mod pallet {
...
+                match &market.period {
+                    MarketPeriod::Block(range) => {
+                        let close_time = now_block.saturating_add(block_period);
+                        ensure!(close_time < range.end, Error::<T>::EarlyCloseRequestTooLate);
+                        Ok(MarketPeriod::Block(range.start..close_time))
+                    }
+                    MarketPeriod::Timestamp(range) => {
+                        let close_time = now_time.saturating_add(time_frame_period);
+                        ensure!(close_time < range.end, Error::<T>::EarlyCloseRequestTooLate);
+                        Ok(MarketPeriod::Timestamp(range.start..close_time))
```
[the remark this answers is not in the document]
Chralt98: f7fb8a863db62ad57782a2918f68c62016058064

Chralt98:
> I haven't checked if admin_destroy_market cleans everything up properly. I will remove admin_destroy_market in a PR in the next couple of days.

Ok!

> The changes made here don't play well with the grace period. I think that's fine. We should really think about sunsetting the grace period. If we end up not doing that, scheduling early close should extend to the grace period, i.e. a market in grace period should also be "early closeable".

Did you already create an issue for this? If not, can you do this please? I think as soon as the grace period starts, the market is already closed.

> It might be a good idea to only allow the AC to schedule early closure for now so the UI team doesn't have to implement the dispute process. @Robiquet @sea212 Opinions?

We decided already to allow the market creator and the superior AC to close the market early. Yeah, the front end needs to deal with this from then on. I don't think it's a good idea to throw all the features away that we decided on now.

> slash_pending_bonds must be updated.

Since admin_destroy_market is removed, slash_pending_bonds will be removed with it.

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
maltekliemann: > https://github.com/zeitgeistpm/zeitgeist/blob/175070378e0a99d5e02e475727568f384b308feb/zrml/prediction-markets/src/lib.rs#L499-L503

Yeah, that's wrong too.

zrml/prediction-markets/src/tests.rs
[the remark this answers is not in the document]
maltekliemann: Yeah, rustfmt won't touch those macros.

maltekliemann (approved):

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+        #[pallet::weight(
+            T::WeightInfo::schedule_early_close_as_authority(CacheSize::get(), CacheSize::get())
+                .max(T::WeightInfo::schedule_early_close_after_dispute(
+                    CacheSize::get(),
+                    CacheSize::get(),
+                ))
+                .max(T::WeightInfo::schedule_early_close_as_market_creator(
+                    CacheSize::get(),
+                    CacheSize::get(),
+                ))
```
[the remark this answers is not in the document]
maltekliemann: Nevermind, this is nonsense.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                                  time_frame_period|
+             -> Result<
+                MarketPeriod<T::BlockNumber, MomentOf<T>>,
+                DispatchError,
+            > {
+                match &market.period {
+                    MarketPeriod::Block(range) => {
+                        let close_time = now_block.saturating_add(block_period);
+                        ensure!(close_time < range.end, Error::<T>::EarlyCloseRequestTooLate);
+                        Ok(MarketPeriod::Block(range.start..close_time))
```
sea212, line 1413: Shouldn't it be inclusive?

```suggestion
                        Ok(MarketPeriod::Block(range.start..=close_time))
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                    }
+                    MarketPeriod::Timestamp(range) => {
+                        let close_time = now_time.saturating_add(time_frame_period);
+                        ensure!(close_time < range.end, Error::<T>::EarlyCloseRequestTooLate);
+                        Ok(MarketPeriod::Timestamp(range.start..close_time))
```
sea212, line 1418: Shouldn't it be inclusive?

```suggestion
                        Ok(MarketPeriod::Timestamp(range.start..=close_time))
```

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                | EarlyCloseState::Rejected => {
+                    return Err(Error::<T>::InvalidEarlyCloseState.into());
+                }
+            };
+
+            // ensure we don't dispute if the old market period is already over
+            // so that we don't switch back to an invalid market period
+            // this should never trigger, because at the time of scheduling a new market period,
+            // we ensure the new end is always before the old end
+            debug_assert!(
```
sea212, line 1556: debug `log`

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                    get_new_period(block_period, time_frame_period)?
+                }
+                Some(p) => {
+                    ensure!(is_authorized, Error::<T>::OnlyAuthorizedCanScheduleEarlyClose);
+
+                    match p.state {
+                        // in these case the market period got already reset to the old period
+                        EarlyCloseState::Disputed => {
+                            if Self::is_close_dispute_bond_pending(&market_id, &market, false) {
+                                Self::repatriate_close_dispute_bond(&market_id, &market.creator)?;
```
sea212, line 1460: Why does the market creator get it? I think it makes more sense to let the `Slash` handler deal with it.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                close_dispute_bond,
+            )?;
+
+            let ids_len_0 = Self::clear_auto_close(&market_id)?;
+
+            <zrml_market_commons::Pallet<T>>::mutate_market(&market_id, |market| {
+                market.period = early_close.old.clone();
+                market.bonds.close_dispute = Some(Bond::new(who.clone(), close_dispute_bond));
+                early_close.state = EarlyCloseState::Disputed;
+                market.early_close = Some(early_close);
```
sea212, line 1589: Can it happen that the market resolves at the original (first) close time, but the storage still contains `market.early_close = Some(early_close)`? In that case the final market storage element would represent an untruthful state.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+
+            if let Some(disputor_bond) = market.bonds.close_dispute.as_ref() {
+                let close_disputor = &disputor_bond.who;
+                Self::repatriate_close_request_bond(&market_id, close_disputor)?;
+                Self::unreserve_close_dispute_bond(&market_id)?;
+            }
+
+            <zrml_market_commons::Pallet<T>>::mutate_market(&market_id, |market| {
+                early_close.state = EarlyCloseState::Rejected;
+                market.early_close = Some(early_close);
```
sea212, line 1677: Can it happen that the market resolves at the original (first) close time, but the storage still contains `market.early_close = Some(early_close)`? In that case the final market storage element would represent an untruthful state.

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+            let close_dispute_bond = T::CloseEarlyDisputeBond::get();
+
+            T::AssetManager::reserve_named(
+                &Self::reserve_id(),
+                Asset::Ztg,
+                &who,
+                close_dispute_bond,
+            )?;
+
+            let ids_len_0 = Self::clear_auto_close(&market_id)?;
```
sea212, line 1583: It seems to me that this could be used to mock the market creator. Example:
Market creator schedules a truthful close in let's say 10 days.
`CloseEarlyBlockPeriod=5days` passes.
shortly before the close would happen, someome disputes and effectively denies the truthful close.

I think we have to mitigate this by enforcing a dispute duration after a close request that is bound to `CloseEarlyBlockPeriod`. Of course, the privileged origin

zrml/prediction-markets/src/lib.rs
```diff
@@ -1646,6 +2012,17 @@ mod pallet {
...
+        RequesterNotCreator,
+        /// The early close would be scheduled after the original market period end.
+        EarlyCloseRequestTooLate,
+        /// This early close state is not valid.
+        InvalidEarlyCloseState,
+        /// There is no early close scheduled.
+        NoEarlyCloseScheduled,
+        /// After there was an early close already scheduled,
+        /// only the `CloseMarketsEarlyOrigin` can schedule another one.
+        OnlyAuthorizedCanScheduleEarlyClose,
```
sea212, line 2025: Errors should always be put at the ending of the error enum, otherwise we mess up error indices and give frontend devs a hard time.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+        for i in 0..n {
+            MarketIdsPerCloseTimeFrame::<T>::try_mutate(
+                Pallet::<T>::calculate_time_frame_of_moment(new_range_end),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
+
+        let close_origin = T::CloseMarketEarlyOrigin::try_successful_origin().unwrap();
+        let call = Call::<T>::schedule_early_close { market_id };
+    }: { call.dispatch_bypass_filter(close_origin)? }
```
sea212, line 1188: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+        )?;
+
+        Pallet::<T>::dispute_early_close(
+            RawOrigin::Signed(caller.clone()).into(),
+            market_id,
+        )?;
+
+        let close_origin = T::CloseMarketEarlyOrigin::try_successful_origin().unwrap();
+        let call = Call::<T>::schedule_early_close { market_id };
+    }: { call.dispatch_bypass_filter(close_origin)? }
```
sea212, line 1232: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+        for i in 0..n {
+            MarketIdsPerCloseTimeFrame::<T>::try_mutate(
+                Pallet::<T>::calculate_time_frame_of_moment(new_range_end),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
+
+        let origin = RawOrigin::Signed(market_creator).into();
+        let call = Call::<T>::schedule_early_close { market_id };
+    }: { call.dispatch_bypass_filter(origin)? }
```
sea212, line 1269: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+        for i in 0..n {
+            MarketIdsPerCloseTimeFrame::<T>::try_mutate(
+                Pallet::<T>::calculate_time_frame_of_moment(new_range_end),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
+
+        let origin = RawOrigin::Signed(market_creator).into();
+        let call = Call::<T>::dispute_early_close { market_id };
+    }: { call.dispatch_bypass_filter(origin)? }
```
sea212, line 1319: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+
+        for i in 0..n {
+            MarketIdsPerCloseTimeFrame::<T>::try_mutate(
+                Pallet::<T>::calculate_time_frame_of_moment(new_range_end),
+                |ids| ids.try_push(i.into()),
+            ).unwrap();
+        }
+
+        let call = Call::<T>::reject_early_close { market_id };
+    }: { call.dispatch_bypass_filter(close_origin)? }
```
sea212, line 1369: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -1150,6 +1153,248 @@ benchmarks! {
...
+
+        Pallet::<T>::dispute_early_close(
+            RawOrigin::Signed(caller.clone()).into(),
+            market_id,
+        )?;
+
+        let close_origin = T::CloseMarketEarlyOrigin::try_successful_origin().unwrap();
+
+        let call = Call::<T>::reject_early_close { market_id };
+    }: { call.dispatch_bypass_filter(close_origin)? }
```
sea212, line 1396: Missing `verify` function.

"No path dependency" is no argument. The extrinsic will silently fail if something goes wrong and you will never know and benchmark executing the error branch. Look at some standard libraries, they have no path dependence in many cases and always `verify` the success of the call:

- [pallet-balances](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/balances/src/benchmarking.rs)
- [pallet-assets](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/assets/src/benchmarking.rs)
- [pallet-utility](https://github.com/paritytech/polkadot-sdk/blob/39c04fdd9622792ba8478b1c1c300417943a034b/substrate/frame/utility/src/benchmarking.rs)
- etc.

zrml/prediction-markets/src/lib.rs
Chralt98: e8a85e4e742583d69e31521bec22c2e6d21887b6

zrml/prediction-markets/src/lib.rs
Chralt98: I don't understand this. A market creator directly schedules a close at `now + CloseEarlyBlockPeriod`. If someone disputes shortly before the close, indeed it would extend the closing time, because the AC has to decide from then on. But the disputant reserved a bond, which is slashed, if the market creator was right.

zrml/prediction-markets/src/lib.rs
Chralt98: I see your point here. `early_close` of market always contains the `state` and the `old` and `new` market period. This means, if an early close request was rejected by the AC, the `market` still contains the information that an early close was rejected. If it was rejected, it does not mean that the market closed at the `new` end. The market always closes according to the `market.period` field. In addition, at the point of rejection, the market creator can't close the market early anymore. This ability belongs only to the advisory committee at that point. If we write `None` for the early close field, the market creator could schedule an early close again. Generally we had discussions about if the market creator should be able to schedule an early close again after one rejection. I don't think this is necessary and would actually put more strain on the AC, even if the market creator pays a bond each time. Just let the AC close the market early, if even required.

zrml/prediction-markets/src/lib.rs
Chralt98: Please look here https://github.com/zeitgeistpm/zeitgeist/pull/1127#discussion_r1370084229

If the `early_close` field ever contains `state` `Disputed` or `Rejected`, it means that the early close was not applied (as the name of the state outlines). But generally I see your point here!

zrml/prediction-markets/src/lib.rs
Chralt98: Good question. Was just an idea. Indeed, we could fund the treasury with it. On the other hand, since the AC consists of members of Zeitgeist, we would have an incentive conflict here. Because Zeitgeist would receive the bond if the AC accepts the close request by the market creator. I know it's not that huge of a effect, but the next best option is to just give it to the market creator, because the market creator had to wait (opportunity cost) with a reserved bond.

zrml/prediction-markets/src/lib.rs
Chralt98: 34770207a96cf5c92cb5f38907acad074e428d5b

zrml/prediction-markets/src/lib.rs
Chralt98: Good thought! I actually copied from [here](https://github.com/zeitgeistpm/zeitgeist/blob/4e2569526fa1a8d4b9c04f7aaecc380ef8b7f09d/zrml/prediction-markets/src/lib.rs#L2441-L2445), but as you focus me on it right now, I think about it.... 

Ah.. market period doesn't allow inclusive ranges:

```
mismatched types
expected struct `std::ops::Range<<T as frame_system::Config>::BlockNumber>`
   found struct `RangeInclusive<<T as frame_system::Config>::BlockNumber>`
```

I think the effect of having inclusive ranges here is marginal. (1 block)

zrml/prediction-markets/src/lib.rs
Chralt98: Good thought! I actually copied from [here](https://github.com/zeitgeistpm/zeitgeist/blob/4e2569526fa1a8d4b9c04f7aaecc380ef8b7f09d/zrml/prediction-markets/src/lib.rs#L2441-L2445), but as you focus me on it right now, I think about it.... 

Ah.. market period doesn't allow inclusive ranges:

```
mismatched types
expected struct `std::ops::Range<<T as frame_system::Config>::BlockNumber>`
   found struct `RangeInclusive<<T as frame_system::Config>::BlockNumber>`
```

I think the effect of having inclusive ranges here is marginal. (1 block)

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                        Ok(MarketPeriod::Block(range.start..close_time))
+                    }
+                    MarketPeriod::Timestamp(range) => {
+                        let close_time = now_time.saturating_add(time_frame_period);
+                        ensure!(close_time < range.end, Error::<T>::EarlyCloseRequestTooLate);
+                        Ok(MarketPeriod::Timestamp(range.start..close_time))
+                    }
+                }
+            };
+            let new_period = match &market.early_close {
```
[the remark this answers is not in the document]
Chralt98: Can you elaborate how you would do it and return the `new_period` without using a `let mut new_period`?

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
             T::DeployPool::deploy_pool(who, market_id, amount, spot_prices, swap_fee)?;
             Ok(Some(T::WeightInfo::create_market_and_deploy_pool(ids_len)).into())
         }
+
+        /// Allows the `CloseMarketsEarlyOrigin`` or the market creator to schedule an early close.
+        ///
+        /// The market creator schedules it `now + EarlyClose...Period` in the future.
+        /// This is to allow enough time for a potential dispute.
+        /// The market creator reserves a `CloseEarlyDisputeBond`, which is returned,
+        /// if the `CloseMarketsEarlyOrigin`` decides to accept the early close request
```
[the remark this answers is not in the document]
Chralt98: 9f7781f959dc305b1afe6c052e8b7d08e31c6eca

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+        /// or if it is not disputed.
+        /// It is slashed, if the early close request is disputed
+        /// and the `CloseMarketsEarlyOrigin`` decides to reject the early close.
```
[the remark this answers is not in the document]
Chralt98: 9f7781f959dc305b1afe6c052e8b7d08e31c6eca

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
             T::DeployPool::deploy_pool(who, market_id, amount, spot_prices, swap_fee)?;
             Ok(Some(T::WeightInfo::create_market_and_deploy_pool(ids_len)).into())
         }
+
+        /// Allows the `CloseMarketsEarlyOrigin`` or the market creator to schedule an early close.
```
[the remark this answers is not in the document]
Chralt98: 9f7781f959dc305b1afe6c052e8b7d08e31c6eca

runtime/zeitgeist/src/parameters.rs
```diff
@@ -203,6 +203,16 @@ parameter_types! {
...
+    /// (Slashable) Bond that is provided for disputing an early market close by the market creator.
+    pub const CloseEarlyDisputeBond: Balance = 2_000 * BASE;
+    // 43_200_000 = 12 hours. Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionTimeFramePeriod: Moment = 43_200_000;
+    // Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionBlockPeriod: BlockNumber = 12 * BLOCKS_PER_HOUR;
+    /// (Slashable) Bond that is provided for scheduling an early market close.
+    pub const CloseEarlyRequestBond: Balance = 200 * BASE;
```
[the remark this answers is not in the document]
Chralt98: I don't know why I decided that. I am open for a discussion here. @maltekliemann Do you agree to make them equal?

runtime/battery-station/src/parameters.rs
```diff
@@ -203,6 +203,16 @@ parameter_types! {
...
     /// The percentage of the advisory bond that gets slashed when a market is rejected.
     pub const AdvisoryBondSlashPercentage: Percent = Percent::from_percent(0);
+    /// (Slashable) Bond that is provided for disputing an early market close by the market creator.
+    pub const CloseEarlyDisputeBond: Balance = 50 * BASE;
+    // 43_200_000 = 12 hours. Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionTimeFramePeriod: Moment = 43_200_000;
+    // Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionBlockPeriod: BlockNumber = 12 * BLOCKS_PER_HOUR;
```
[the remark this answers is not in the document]
Chralt98: 56a980e9406ece9cd54ab2ba7808441f67e3e89c

runtime/battery-station/src/parameters.rs
```diff
@@ -203,6 +203,16 @@ parameter_types! {
...
+    /// (Slashable) Bond that is provided for scheduling an early market close.
+    pub const CloseEarlyRequestBond: Balance = 25 * BASE;
```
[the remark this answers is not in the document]
Chralt98: 56a980e9406ece9cd54ab2ba7808441f67e3e89c

runtime/battery-station/src/parameters.rs
```diff
@@ -248,6 +258,12 @@ parameter_types! {
     pub const OutsiderBond: Balance = 2 * OracleBond::get();
     /// Pallet identifier, mainly used for named balance reserves.
     pub const PmPalletId: PalletId = PM_PALLET_ID;
+    // Waiting time for market creator to close
+    // the market after an early close schedule.
+    pub const CloseEarlyBlockPeriod: BlockNumber = 5 * BLOCKS_PER_DAY;
```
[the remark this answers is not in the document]
Chralt98: 56a980e9406ece9cd54ab2ba7808441f67e3e89c

primitives/src/constants/mock.rs
```diff
@@ -80,6 +80,10 @@ parameter_types! {
 // Prediction Market parameters
 parameter_types! {
     pub const AdvisoryBond: Balance = 25 * CENT;
+    pub const CloseEarlyProtectionTimeFramePeriod: Moment = 3 * MILLISECS_PER_BLOCK as u64;
```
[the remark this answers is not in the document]
Chralt98: f814f54256f230ca938a52516559d1d14f734256

runtime/battery-station/src/parameters.rs
```diff
@@ -203,6 +203,16 @@ parameter_types! {
     pub const AdvisoryBond: Balance = 25 * CENT;
     /// The percentage of the advisory bond that gets slashed when a market is rejected.
     pub const AdvisoryBondSlashPercentage: Percent = Percent::from_percent(0);
+    /// (Slashable) Bond that is provided for disputing an early market close by the market creator.
+    pub const CloseEarlyDisputeBond: Balance = 50 * BASE;
+    // 43_200_000 = 12 hours. Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionTimeFramePeriod: Moment = 43_200_000;
```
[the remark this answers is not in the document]
Chralt98: f814f54256f230ca938a52516559d1d14f734256

runtime/battery-station/src/parameters.rs
```diff
@@ -248,6 +258,12 @@ parameter_types! {
     pub const OutsiderBond: Balance = 2 * OracleBond::get();
     /// Pallet identifier, mainly used for named balance reserves.
     pub const PmPalletId: PalletId = PM_PALLET_ID;
+    // Waiting time for market creator to close
+    // the market after an early close schedule.
+    pub const CloseEarlyBlockPeriod: BlockNumber = 5 * BLOCKS_PER_DAY;
+    // 432_000_000 = 5 days. Waiting time for market creator to close
+    // the market after an early close schedule.
+    pub const CloseEarlyTimeFramePeriod: Moment = 432_000_000;
```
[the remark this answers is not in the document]
Chralt98: f814f54256f230ca938a52516559d1d14f734256

runtime/zeitgeist/src/parameters.rs
```diff
@@ -203,6 +203,16 @@ parameter_types! {
     pub const AdvisoryBond: Balance = 200 * BASE;
     /// The percentage of the advisory bond that gets slashed when a market is rejected.
     pub const AdvisoryBondSlashPercentage: Percent = Percent::from_percent(0);
+    /// (Slashable) Bond that is provided for disputing an early market close by the market creator.
+    pub const CloseEarlyDisputeBond: Balance = 2_000 * BASE;
+    // 43_200_000 = 12 hours. Fat-finger protection for the advisory committe to reject
+    // the early market schedule.
+    pub const CloseEarlyProtectionTimeFramePeriod: Moment = 43_200_000;
```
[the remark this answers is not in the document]
Chralt98: f814f54256f230ca938a52516559d1d14f734256

runtime/zeitgeist/src/parameters.rs
```diff
@@ -248,6 +258,12 @@ parameter_types! {
     pub const OutsiderBond: Balance = 2 * OracleBond::get();
     /// Pallet identifier, mainly used for named balance reserves. DO NOT CHANGE.
     pub const PmPalletId: PalletId = PM_PALLET_ID;
+    // Waiting time for market creator to close
+    // the market after an early close schedule.
+    pub const CloseEarlyBlockPeriod: BlockNumber = 5 * BLOCKS_PER_DAY;
+    // 432_000_000 = 5 days. Waiting time for market creator to close
+    // the market after an early close schedule.
+    pub const CloseEarlyTimeFramePeriod: Moment = 432_000_000;
```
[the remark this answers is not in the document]
Chralt98: f814f54256f230ca938a52516559d1d14f734256

primitives/src/market.rs
```diff
@@ -239,6 +252,21 @@ impl<BN: MaxEncodedLen, M: MaxEncodedLen> MaxEncodedLen for MarketPeriod<BN, M>
     }
 }
```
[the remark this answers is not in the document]
Chralt98: 2d0b0c84b6ec130591da92cc8f6178a1fddf9858

zrml/prediction-markets/src/lib.rs
```diff
@@ -1349,6 +1350,338 @@ mod pallet {
...
+                            if Self::is_close_request_bond_pending(&market_id, &market, false) {
+                                Self::unreserve_close_request_bond(&market_id)?;
+                            }
+                        }
+                        EarlyCloseState::Rejected => {}
+                        EarlyCloseState::ScheduledAsMarketCreator
+                        | EarlyCloseState::ScheduledAsOther => {
+                            return Err(Error::<T>::InvalidEarlyCloseState.into());
+                        }
+                    }
```
sea212, line 1471:
```suggestion
            let new_period = if let Some(p) = &market.early_close {
                ensure!(is_authorized, Error::<T>::OnlyAuthorizedCanScheduleEarlyClose);

                match p.state {
                    // in these case the market period got already reset to the old period
                    EarlyCloseState::Disputed => {
                        if Self::is_close_dispute_bond_pending(&market_id, &market, false) {
                            Self::repatriate_close_dispute_bond(&market_id, &market.creator)?;
                        }
                        if Self::is_close_request_bond_pending(&market_id, &market, false) {
                            Self::unreserve_close_request_bond(&market_id)?;
                        }
                    }
                    EarlyCloseState::Rejected => {}
                    EarlyCloseState::ScheduledAsMarketCreator
                    | EarlyCloseState::ScheduledAsOther => {
                        return Err(Error::<T>::InvalidEarlyCloseState.into());
                    }
                }

                get_new_period(
                    T::CloseEarlyProtectionBlockPeriod::get(),
                    T::CloseEarlyProtectionTimeFramePeriod::get(),
                )?
            } else {
                let (block_period, time_frame_period) = if is_authorized {
                    // fat finger protection
                    (
                        T::CloseEarlyProtectionBlockPeriod::get(),
                        T::CloseEarlyProtectionTimeFramePeriod::get(),
                    )
                } else {
                    let market_creator = market_creator.ok_or(Error::<T>::RequesterNotCreator)?;
                    let close_request_bond = T::CloseEarlyRequestBond::get();

                    T::AssetManager::reserve_named(
                        &Self::reserve_id(),
                        Asset::Ztg,
                        &market_creator,
                        close_request_bond,
                    )?;

                    <zrml_market_commons::Pallet<T>>::mutate_market(&market_id, |market| {
                        market.bonds.close_request =
                            Some(Bond::new(market_creator, close_request_bond));
                        Ok(())
                    })?;

                    (T::CloseEarlyBlockPeriod::get(), T::CloseEarlyTimeFramePeriod::get())
                };

                get_new_period(block_period, time_frame_period)?
            };
```

zrml/prediction-markets/src/lib.rs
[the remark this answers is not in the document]
sea212: See https://github.com/zeitgeistpm/zeitgeist/pull/1127/files/c9cbc8e793c53345da94fa6908f36c2275366f87#r1370850354 for fully functional and properly formatted example.

zrml/prediction-markets/src/lib.rs
sea212: How about this:

```suggestion
                        Ok(MarketPeriod::Block(range.start..close_time + 1))
```

zrml/prediction-markets/src/lib.rs
sea212: How about this:

```suggestion
                        Ok(MarketPeriod::Block(range.start..close_time + 1))
```

zrml/prediction-markets/src/lib.rs
sea212: Right as of now the AC is kind of centralized, however that should not be an assumption. I don't know, it feels like the market creator is incentivized to mislead others into making false disputes.

zrml/prediction-markets/src/lib.rs
sea212: Please verify if the following scenario could occur given the current implementation:
1. `CloseEarlyBlockPeriod = 100 blocks`
2. `creator` requests `close` in 500 blocks
3. `close` in 500 blocks is approved
4. 490 blocks later, malicious disputor disputes the close request, effectively postponing it
5. malicious disputor has insight of the outcome at 500 blocks (lookup in internet or something), but since the malicious disputor postponed the closing, he is able to buy all correct outcome tokens in the market
6. malicious disputor is in a profit, even when losing the close early dispute bond.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

zrml/prediction-markets/src/benchmarks.rs
sea212: This is nonsense.

runtime/zeitgeist/src/parameters.rs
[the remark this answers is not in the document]
maltekliemann: I'm fine with either. I'd also add a zero to all bonds. 200 ZTG is like five bucks right now.

zrml/prediction-markets/src/lib.rs
Chralt98: Ok this works, but it doesn't really has an effect. We could also extend the `CloseEarlyProtectionBlockPeriod` by one block. If I use your approach, I need to adapt the tests and benchmarks, which doesn't make any sense to me, if we could also extend the periods by one.

zrml/prediction-markets/src/lib.rs
Chralt98: Ok this works, but it doesn't really has an effect. We could also extend the CloseEarlyProtectionBlockPeriod by one block. If I use your approach, I need to adapt the tests and benchmarks, which doesn't make any sense to me, if we could also extend the periods by one.

zrml/prediction-markets/src/lib.rs
Chralt98: Another option would be to `withdraw` or burn the bond.

> market creator is incentivized to mislead others into making false disputes

If the market creator would ever mislead, he would get slashed. Can you give an example how this would look in practice? It seems to me that this is a scenario that would never happen.

zrml/prediction-markets/src/lib.rs
Chralt98: > 1. `CloseEarlyBlockPeriod = 100 blocks`
> 2. `creator` requests `close` in 500 blocks

The creator can't request a close in 500 blocks, but only at `now + CloseEarlyBlockPeriod`.

> 3. `close` in 500 blocks is approved

This is conflicting with the point 4 you made. The market creators request is not approved if it is disputed.

> 4. 490 blocks later, malicious disputor disputes the close request, effectively postponing it

Yeah that's correct, the disputant postpones the close early request, effectively waiting for the AC to schedule a shorter early close or a rejection.

> 5. malicious disputor has insight of the outcome at 500 blocks (lookup in internet or something), but since the malicious disputor postponed the closing, he is able to buy all correct outcome tokens in the market

Good thought and now I see your point! The implemented feature is for the purpose of giving the market participants the opportunity to `redeem_shares` earlier than they would have done normally. It is a better user experience to not sit in your positions any longer than required. The market creator will only close at the point when the true outcome is already known anyways. So your described scenario could already play out before the market creator even thinks about closing the market early.

> 6. malicious disputor is in a profit, even when losing the close early dispute bond.

The profit at this point is much lower than the expected profit at the point of the true outcome becoming known. Assuming the true outcome is below 1.

runtime/zeitgeist/src/parameters.rs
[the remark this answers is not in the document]
Chralt98: e174c6407a3ba1e627feef9556ca4e16784b309f

zrml/prediction-markets/src/lib.rs
sea212: Ok, let's leave it as it is.

zrml/prediction-markets/src/lib.rs
sea212: I see, I thought the purpose was to correct the closing date based on the availability of information of the outcome.

zrml/prediction-markets/src/lib.rs
Chralt98: @maltekliemann

maltekliemann: I think ideally we should use the same rules as we do for disputes on oracle reports, but I don't have any strong opinions one way or the other.

As for the game theory, I see the point, except that I'm not really sure how you'd mislead people into making incorrect disputes. I think you could level the same criticism against the decentralized court.

sea212 (approved):

PR #1129, opened 2023-09-28
Use disk space optimization for benchmark checks

sea212:
### What does it do?
- Adds a config parameters to the benchmarking scripts that when set does `cargo clean` between the standalone and parachain benchmark runs

- Potential solution to fix broken CI pipekine due to insufficient amount of disk space.


scripts/benchmarks/run_benchmarks.sh
```diff
@@ -71,6 +71,12 @@ for pallet in ${ZEITGEIST_PALLETS[@]}; do
         $ADDITIONAL_PARAMS
 done
 
+if [ -n "$DISK_SPACE_OPTIMIZED" ]; then
+   echo "Disk space optimization actived."
```
sea212, line 75:
```suggestion
   echo "Disk space optimization activated."
```

Chralt98 (commented):
So this is an optimisation to get the `LLVM ERROR: IO failure on output stream: No space left on device` away? But it's still there in your CI. 

But in general if we have the first cargo build:

```
cargo build \
    --profile=$PROFILE \
    --features=runtime-benchmarks,$ADDITIONAL_FEATURES \
    --bin=zeitgeist
```

And the second cargo build is:

```
cargo build \
    --profile=$PROFILE \
    --features=runtime-benchmarks,parachain,$ADDITIONAL_FEATURES \
    --bin=zeitgeist
```

then we have essentially the same crates to compile, except for the second run there needs to be the feature parachain compiled. So the second cargo build adds additional compilation overhead (artifacts). So, can you explain what `cargo clean` gets done in this case? I think it even slows down compilation time, while having the same disk space required.

Chralt98:
Do benchmark runs produce artefacts in the `target` folder?

sea212:
Yes, the build time significantly increases since many build files that are shared are removed. The idea was to exchange disk space for build time to make GitHub hosted runners capable of finishing the job.

It seems like that approach does not lead to GitHub hosted runners being capable to produce results in time. Looks like now the issue is that it just takes too long.

Closing in favor of https://github.com/zeitgeistpm/zeitgeist/pull/1134

Issue #1093, opened 2023-09-06
[PredictionMarkets] Can't sell complete sets on inactive market


maltekliemann:
That's definitely wrong. Why would this not be allowed? Sucks for LPs who are needlessly sitting on their complete sets until the market resolves.

sea212:
Selling complete sets would change the probability, although the prediction / trade period is over, right? Seems not intuitive to me. Would be fine to allow selling complete sets in this case if we figure out a way to freeze prediction outcomes.

maltekliemann:
> Selling complete sets would change the probability, although the prediction / trade period is over, right?

No, that's completely independent.

PR #1144, opened 2023-10-13
Allow selling complete sets regardless of market status

maltekliemann:
### What does it do?

Allows selling complete sets regardless of market status. This is fine because even if the market is resolved, selling full sets can't be used to "arbitrage" redeeming. Also adds a test which ensures that `buy_complete_set` does observe the market state.

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1093.


Chralt98 (approved):
Why adding label `transactions-changed`?

maltekliemann:
Yeah, good question, that was wrong.

Issue #1131, opened 2023-09-28
Remove reliance on unicode standard from fixed conversion functions


maltekliemann:
https://github.com/zeitgeistpm/zeitgeist/pull/1092/#discussion_r1338740959

PR #1151, opened 2023-10-17
Improve durability of fixed point conversion function

maltekliemann:
### What does it do?

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1131 and generally restructures the function to be more rigid and clear.

### DIFF
--- a/primitives/src/math/fixed.rs
+++ b/primitives/src/math/fixed.rs
@@ -254,12 +254,18 @@ impl<F: Fixed + ToString, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
         let mut increment_int_part = false;
         let new_frac_part = if frac_part.len() < decimals_usize {
             format!("{}{}", frac_part, "0".repeat(decimals_usize.saturating_sub(frac_part.len())))
-        } else if frac_part.len() > decimal_usize {
-            let round_digit = frac_part.chars().nth(decimals_usize).and_then(|c| c.to_digit(10)).ok_or("Either the to_digit could not convert to a number or the char at decimals_usize was not found, although the frac_part length is higher than decimals_usize.")?;
+        } else if frac_part.len() > decimals_usize {
+            let round_digit =
+                frac_part.chars().nth(decimals_usize).and_then(|c| c.to_digit(10)).ok_or(
+                    "The char at decimals_usize was not found, although the frac_part length is \
+                     higher than decimals_usize.",
+                )?;
             if round_digit >= 5 {
                 increment_int_part = true;
             }
-            frac_part.chars().take(decimals_usize).collect()
+            frac_part.chars().take(decimals_usize).collect::<String>()
+        } else {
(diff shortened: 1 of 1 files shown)


primitives/src/math/fixed.rs
```diff
@@ -243,36 +243,35 @@ impl<F: Fixed + ToString, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
...
+            return Err("The string contains multiple decimal points");
         }
 
-        let (int_part, frac_part) = (parts[0], parts[1]);
-        let mut increment = false;
-
+        let mut increment_int_part = false;
         let new_frac_part = if frac_part.len() < decimals_usize {
             format!("{}{}", frac_part, "0".repeat(decimals_usize.saturating_sub(frac_part.len())))
         } else {
```
Chralt98, line 257:
```suggestion
        } else if frac_part.len() > decimal_usize {
```

> Note that `round_digit` might be `None`, for example if `frac_part = 245` and `decimals = 3`. There's no need to round in that case.

So I assume it's always the case that if `frac_part.len() == decimals` then we don't need to round.

primitives/src/math/fixed.rs
```diff
@@ -243,36 +243,35 @@ impl<F: Fixed + ToString, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
...
-                Some(d) if d >= '5' => increment = true,
-                _ => {}
+            let round_digit = frac_part.chars().nth(decimals_usize).and_then(|c| c.to_digit(10));
+            // Note that `round_digit` might be `None`, for example if `frac_part = 245` and
+            // `decimals = 3`. There's no need to round in that case.
+            if let Some(d) = round_digit {
+                if d >= 5 {
+                    increment_int_part = true;
+                }
             }
```
Chralt98, line 265:
```suggestion
            let round_digit = frac_part.chars().nth(decimals_usize).and_then(|c| c.to_digit(10)).ok_or("The char at decimals_usize was not found, although the frac_part length is higher than decimals_usize.")?;
            if round_digit >= 5 {
                increment_int_part = true;
            }
```

maltekliemann: Yeah, I had exactly this. But it turns out it's exactly _not_ what want ;) As the comment explains, `round_digit` might be `None`, in which case we can pretend that it's zero; or, equivalently, if it's `None`, then we just don't do any rounding. You'll find that a couple of tests will fail if we make the proposed change.

maltekliemann: Nevermind, just saw the comment above.

maltekliemann: Yeah, nice suggestion. The only problem here is that we also need an `else` case, so I think it's fine to leave it the way it is.

Chralt98 (approved):

sea212 (approved):

PR #1157, opened 2023-10-20
Fix build

sea212:
### What does it do?
Adds a missing import.

### DIFF
--- a/primitives/src/math/fixed.rs
+++ b/primitives/src/math/fixed.rs
@@ -31,7 +31,11 @@ use crate::{
         },
     },
 };
-use alloc::{borrow::ToOwned, format, string::ToString};
+use alloc::{
+    borrow::ToOwned,
+    format,
+    string::{String, ToString},
+};
 use core::convert::TryFrom;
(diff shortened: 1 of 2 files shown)


primitives/src/math/fixed.rs
```diff
@@ -252,21 +256,27 @@ impl<F: Fixed + ToString, N: TryFrom<u128>> FromFixedToDecimal<F> for N {
...
+                    frac_part.chars().nth(decimals_usize).and_then(|c| c.to_digit(10)).ok_or(
+                        "The char at decimals_usize was not found, although the frac_part length \
+                         is higher than decimals_usize.",
+                    )?;
+                if round_digit >= 5 {
+                    increment_int_part = true;
+                }
+                frac_part.chars().take(decimals_usize).collect::<String>()
+            }
+            Ordering::Equal => frac_part.to_string(),
```
maltekliemann, line 279: Looking at this now, this arm is just a special case of the first arm. You could just do the `format!` call even if `frac_part` has length `decimals_usize`.

maltekliemann (approved):

PR #1163, opened 2023-10-25
Update neo-swaps events

maltekliemann:
### What does it do?

Adds fields to neo-swaps events that are needed by subsquid.


saboonikhil (approved):
Thank you @maltekliemann !

sea212 (approved):

Issue #1176, opened 2023-11-01
Properly configure pallet-assets `RemoveItemsLimit`


sea212:
Currently the pallet-assets instances in the runtime use a conservative `RemoveItemsLimit` configuration. Optimally configuring `RemoveItemsLimit` requires that weights for pallet-assets from the reference machine are available. Once those are available, all pallet-assets instances should use an optimized value for `RemoveItemsLimit`.

maltekliemann:
Closing as no longer relevant.

PR #1177, opened 2023-11-01
New asset system - Separate asset types and add pallet-assets instances to runtime

sea212:
### What does it do?
- It creates new asset types
  - `u128` for user created assets and campaign assets
  - `MarketAssets` that represent any kind of volatile asset that is used during a market lifecycle 
- It adds [pallet-assets](https://github.com/paritytech/substrate/tree/polkadot-v0.9.38/frame/assets) to the runtime
  - One instance as `Assets`, this is the general instance that anyone can utilize to create new asset classes.
  - One instance as `CampaignAssets`, this is a permissioned instance that prohibits creation of asset using dispatchable calls. Only code, root, 2/3 council or all tech committee can trigger the creation of a new campaign assets. Campaign assets are special in that they are sufficient to create accounts and can be used to pay fees (to be implemented)
  - One instance as `MarketAssets`, this is a permissioned instance that prohibits creation of asset using dispatchable calls. Only code, root or 2/3 tech committee can trigger the creation of a new asset, such as `MarketAsset::CategoricalOutcome(0,0)`
- It adds custom pallet-assets weights and add pallet-assets to the benchmarking scripts

- `RemoveItemsLimit` could not be properly configured yet, see https://github.com/zeitgeistpm/zeitgeist/issues/1176

- https://github.com/zeitgeistpm/zeitgeist/issues/1176
- This is part 1 of multipe PRs that complete the new asset system. Logic that is still missing:
  - Make existing pallets use the new asset system.
  - Lazy migrate existing market assets: Pallets will continue using `orml-tokens` for existing market until all those existing market assets expired. Reason: Too many accounts hold existing market assets, migrating will be expensive and cannot be executed atomatically. During migration, the storage trie that hold old market asset information will most likely be manipulated by user interaction.
  - Abstract ordinary token functions like `transferring`, `reserving/freezing`, etc. in `AssetManager`.
  - Paying fees using `CampaignAssets`.
  - Using `Assets` and `CampaignAssets` as base assets in prediction markets.
  - Properly creating and destroying `MarketAssets` in pallets such as prediction markets.
  - Adjust Asset types in many pallets, such as xcm related pallets (should only allow `ZTG` and `ForeignAsset(id)` for now), orml-tokens, etc.
  - Future: Remove outcome asset classes from the old asset structure once all associated outcome tokens expired.

- Using one asset class enum and one pallet instance for pallet-instance with a `CreateOrigin` that knows which origins can create which asset class. Using instances instead seemed like a cleaner approach that logically separates asset classes with different purposes and restrictions.


runtime/common/src/lib.rs
```diff
@@ -625,6 +633,110 @@ macro_rules! impl_config_traits {
...
+            #[cfg(feature = "runtime-benchmarks")]
+            type BenchmarkHelper = CustomAssetsBenchmarkHelper;
+            type CallbackHandle = ();
+            type CreateOrigin = AsEnsureOriginWithArg<EnsureNever<AccountId>>;
+            type Currency = Balances;
+            type Extra = ();
+            type ForceOrigin = EitherOfDiverse<
+                EnsureRootOrTwoThirdsCouncil,
+                EnsureRootOrAllTechnicalCommittee,
+            >;
```
maltekliemann, line 691: I think we should use the same `ForceOrigin` for all pallets. The distinction between Council and Tech Comm isn't there right now anyways.

sea212: I disagree for the following reasons:

1. The `CampaignAsset` class is much more powerful, as it allows to spawn assets that can be used to pay transaction fees. Thus it should require at least higher privileges than other asset classes to be created.
2. There are distinctions:
  a. While other asset classes use the `ForceOrigin` solely to technically correct faults, the `CampaignAsset` class uses the `ForceOrigin` to realize business decisions (pushing into directions by utilizing campaigns)
  b. The Council is different from the Technical Committee (TC), even on Zeitgeist. In contrast to the TC, the Council contains all members of the executive. Currently the Council consists of 5 members, with an required agreement from the Council of at least 2/3 to spawn campaign assets, at least two members of the executive have to agree, thus also mapping agreement within the executive of Zeitgeist to move forward with campaigns.

I decided to add the TC to allow assistance in regards to technical aspects, such as cleaning up campaign assets after a campaign is over. However, I think granting the (complete) TC the `ForceOrigin` power invalidates my second argument. I am leaning towards removing the TC completely from the `ForceOrigin` of the `CampaignAsset` class. What's your opinion?

This makes me think about point 1. though. The `MarketAssets` have the `ForceOrigin` `TwoThirdsOrTechnicalCommittee`. Obviously `MarketAssets` have an even more tangible value. `MarketAssets` should never be managed directly as the pallets control the whole lifecycle of market assets such as outcome tokens or pool shares. I added this because I wanted to give the TC the opportunity to hotfix some inconsistent states introduced by pallets. Thinking more about it though, I think the TC should not have the power to manage market assets. What's your opinion?

Makes me also question right now if it makes sense at all to add the TC to any `ForceOrigin`, including `CustomAssets`. After all, with the help of the Council the TC is capable of spawning a referendum with a short period until dispatch.

runtime/zeitgeist/src/parameters.rs
```diff
@@ -63,6 +63,36 @@ parameter_types! {
 }
 
 parameter_types! {
+    // Assets (Campaign)
+    pub const CampaignAssetsAccountDeposit: Balance = deposit(1, 16);
+    pub const CampaignAssetsApprovalDeposit: Balance = ExistentialDeposit::get();
+    /// The amount of native currency that is frozen during the whole lifetime
+    /// if an asset class. Freezing happens at asset class creation.
+    /// Irrelevant - No origin can successfully call the associated dispatchable call..
```
sea212, line 71:
```suggestion
    /// Irrelevant - No origin can successfully call the associated dispatchable call.
```

runtime/battery-station/src/parameters.rs
```diff
@@ -63,6 +63,38 @@ parameter_types! {
 }
 
 parameter_types! {
+    // Assets (Campaign)
+    pub const CampaignAssetsAccountDeposit: Balance = deposit(1, 16);
+    pub const CampaignAssetsApprovalDeposit: Balance = ExistentialDeposit::get();
+    /// The amount of native currency that is frozen during the whole lifetime
+    /// if an asset class. Freezing happens at asset class creation.
+    /// Irrelevant - No origin can successfully call the associated dispatchable call..
```
sea212, line 71:
```suggestion
    /// Irrelevant - No origin can successfully call the associated dispatchable call.
```

runtime/common/src/lib.rs
maltekliemann: I'm okay with all of these options. As long as there are no elections, I don't see any difference between the Council and TC. In the long run (assuming elections take place), I'd prefer if the TC did not have control over market assets. This is based on the assumption that until then, all bugs will have been fixed or can be fixed by governance rather than TC intervention.

sea212: Does it make more sense like this: https://github.com/zeitgeistpm/zeitgeist/pull/1177/commits/a86c36af0456b24f37293cb2028cc56126fe65c8

Idea: Still convinced about at least 2/3 Council for CampaignAssets, full tech committee for MarketAssets (due to the potential impact) and 2/3 technical committee for CustomAssets.

Chralt98 (approved):

primitives/src/asset.rs
```diff
@@ -53,6 +56,24 @@ pub enum Asset<MI: MaxEncodedLen> {
...
+/// The `MarketAsset` enum represents all types of assets available in the
+/// prediction market protocol.
+///
+/// # Types
+///
+/// * `MI`: Market Id
+#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
+#[derive(Clone, Copy, Debug, Decode, Default, Eq, Encode, MaxEncodedLen, PartialEq, TypeInfo)]
+pub enum PredictionMarketAsset<MI: MaxEncodedLen> {
```
maltekliemann, line 68:
```suggestion
pub enum PredictionMarketAsset<MI> {
```

Not sure if that's a good idea, but I don't think we need this requirement here.

primitives/src/asset.rs
```diff
@@ -53,6 +56,24 @@ pub enum Asset<MI: MaxEncodedLen> {
...
+    CategoricalOutcome(MI, CategoryIndex),
+    #[default]
+    CombinatorialOutcome,
```
maltekliemann, line 71: This seems odd. I think `CombinatorialOutcome` shouldn't be in here at all, nor should it be the default.

primitives/src/asset.rs
maltekliemann: I guess it's for convenience?

sea212: We will need that for `MaxEncodedLen` implementation for the `Pool` struct.

maltekliemann: Ok, I think we can get rid of this after merging https://github.com/zeitgeistpm/zeitgeist/pull/1197

primitives/src/asset.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1177/commits/e7c912cbeb61dee905da5a7009ff9c294007e0fd

maltekliemann (approved):

Issue #1172, opened 2023-10-30
Required changes from squid for court events


saboonikhil:
Based on extensive discussion with @Chralt98, these would be my recommendations for the events connected to `court` pallet.

1. Putting `slashable` and  `weight` on `JurorRevealedVote`

![Screenshot 2023-10-30 at 3 19 28 PM](https://github.com/zeitgeistpm/zeitgeist/assets/36529278/406da7f7-835e-4cf6-9535-eeb57dffea07)

2. Adding `appeal_info` and updated `round_ends` to `CourtAppealed`. `appeal_number` would not be required then.
https://github.com/zeitgeistpm/zeitgeist/blob/76d6a2398e927bce9e739702c1ef8e98437ee26f/zrml/court/src/types.rs#L158-L165

PR #1182, opened 2023-11-07
Modify court events for indexer

Chralt98:
### What does it do?

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1172

### DIFF
--- a/zrml/court/src/lib.rs
+++ b/zrml/court/src/lib.rs
@@ -235,6 +235,7 @@ mod pallet {
         BoundedVec<CourtPoolItemOf<T>, <T as Config>::MaxCourtParticipants>;
     pub(crate) type DrawOf<T> = Draw<AccountIdOf<T>, BalanceOf<T>, HashOf<T>, DelegatedStakesOf<T>>;
     pub(crate) type SelectedDrawsOf<T> = BoundedVec<DrawOf<T>, <T as Config>::MaxSelectedDraws>;
+    pub(crate) type RoundTimingOf<T> = RoundTiming<BlockNumberFor<T>>;
     pub(crate) type AppealOf<T> = AppealInfo<AccountIdOf<T>, BalanceOf<T>>;
     pub(crate) type AppealsOf<T> = BoundedVec<AppealOf<T>, <T as Config>::MaxAppeals>;
     pub(crate) type RawCommitmentOf<T> = RawCommitment<AccountIdOf<T>, HashOf<T>>;
@@ -323,6 +324,8 @@ mod pallet {
             court_id: CourtId,
             vote_item: VoteItem,
             salt: T::Hash,
+            slashable_amount: BalanceOf<T>,
+            draw_weight: u32,
         },
         /// A juror vote has been denounced.
         DenouncedJurorVote {
@@ -339,7 +342,11 @@ mod pallet {
             delegated_jurors: Vec<T::AccountId>,
         },
         /// A market has been appealed.
-        CourtAppealed { court_id: CourtId, appeal_number: u32 },
+        CourtAppealed {
+            court_id: CourtId,
+            appeal_info: AppealOf<T>,
+            new_round_ends: Option<RoundTimingOf<T>>,
+        },
         /// A new token amount was minted for a court participant.
         MintedInCourt { court_participant: T::AccountId, amount: BalanceOf<T> },
         /// The juror and delegator stakes have been reassigned. The losing jurors have been slashed.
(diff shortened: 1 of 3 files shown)


saboonikhil (dismissed):
Thank you Gentleman!

zrml/court/src/lib.rs
```diff
@@ -878,16 +887,31 @@ mod pallet {
 
                     let raw_vote =
                         Vote::Revealed { commitment, vote_item: vote_item.clone(), salt };
-                    draws[index] = Draw { court_participant: who.clone(), vote: raw_vote, ..draw };
+                    let slashable_amount = draw.slashable;
+                    let draw_weight = draw.weight;
+                    draws[index] = Draw {
```
maltekliemann, line 892: We seem to have missed panicker here. Accessing a bounded vector via `Index` can panic if `index` is out of range.

zrml/court/src/lib.rs
```diff
@@ -957,6 +982,7 @@ mod pallet {
                 };
                 // sets round ends one after the other from now
                 court.update_round(round_timing);
+                new_round_ends = Some(court.round_ends.clone());
```
maltekliemann, line 985: Above you're using `let match`, but here you're using a mutable. Seems like inconsistent style.

maltekliemann: Also, why does `ne_round_ends` have to be defined inside the `if` expression?

Chralt98: > Also, why does new_round_ends have to be defined inside the if expression?

It's defined outside by `let mut new_round_ends = None;`. In the case the condition `appeal_number < T::MaxAppeals::get() as usize` is true, then the round ends were actually updated by `court.update_round(round_timing);`. I don't want to emit an event where `new_round_ends` is written to be the old court round in the case of the last appeal (global dispute backer). So that's why I decided to use `None` for `new_round_ends` to indicate that this appeal belongs to the activation for global disputes.

Chralt98: > Above you're using let match, but here you're using a mutable. Seems like inconsistent style.

I mean, true, it's another style than using `match`, and I know that you rather want to avoid unnecessary `mut` variables, but can you give me an example to achieve the same behaviour without using `mut` ?

zrml/court/src/lib.rs
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1182/commits/6e0574a91f15e85bcbeadeeb6c4482ee808bafd7

zrml/court/src/lib.rs
```diff
@@ -708,16 +715,18 @@ mod pallet {
...
             match draws.binary_search_by_key(&who, |draw| draw.court_participant.clone()) {
                 Ok(index) => {
-                    let draw = draws[index].clone();
-
-                    // allow to override last vote
-                    ensure!(
-                        matches!(draws[index].vote, Vote::Drawn | Vote::Secret { commitment: _ }),
-                        Error::<T>::InvalidVoteState
-                    );
+                    if let Some(draw) = draws.get_mut(index) {
```
maltekliemann, line 718: I know that `debug_assert!` statements are an integral part of how you write code. I'd just like to point out a key difference to what I would recommend, which is:

```
let draw = draws.get_mut(index).unwrap_or_else(|| Error::<T>::Unexpected);
```

The way this code is now written, you are absolutely relying on `draws.get_mut` returning `Some`. If it fails to do so in production, this function will produce unexpected results. That's definitely better than panicking, but be aware that you're actually taking a risk here.

zrml/court/src/lib.rs
maltekliemann: This would do it.

```rust
let (new_round_ends, ids_len_1) = if let ... {
    // ...
} else {
    (None, 0u32)
};
```

Not saying its better or worse, but I'd go with the same style everywhere.

zrml/court/src/lib.rs
```diff
@@ -867,27 +882,47 @@ mod pallet {
...
+                            let raw_vote =
+                                Vote::Revealed { commitment, vote_item: vote_item.clone(), salt };
+                            draw.vote = raw_vote;
+
+                            slashable_amount = draw.slashable;
+                            draw_weight = draw.weight;
+                        } else {
+                            debug_assert!(false, "Binary search should have found a valid index!");
+                        }
+                        (slashable_amount, draw_weight)
```
maltekliemann, line 909: This is a good example of why using `debug_assert!` may not be the best, because you're now forced to return `(slashable_amount, draw_weight)` even though you've already got them stored in the mutable `draws` - just because you don't error if something unexpected happens.

maltekliemann (approved):

Issue #1117, opened 2023-09-17
[orderbook-v1] Fix use of `total_unfilled`


maltekliemann:
I find the use of `total_unfilled` confusing.

Due to the rounding, the maker's balance will always be the first to hit zero, but the taker balance might still be non-zero. You're already ensuring in the `match` expression above that the taker balance is either zero or larger than some magic constant. So checking if `total_unfilled` is non-zero is just checking if the maker balance is non-zero. Why not just do that directly?

Issue #1115, opened 2023-09-17
[orderbook-v1] Miscellaneous issues


maltekliemann:
One PR is enough for these:

- [x] Add pallet and extrinsic documentation
- [x] Fix formatting (`benchmarks.rs`, weight expressions in `lib.rs`, line breaks in comments in `lib.rs`)
- [x] Fix inconsistent spelling of the pallet's name. Most often _orderbook_ and `Orderbook` are used, in benchmarks it's `OrderBook`. FYI, correct spelling of the term is in two words: _order book_.
- [x] Fix test names. Particularly, a test for `remove_order` is still called `cancel_order`.

PR #1183, opened 2023-11-08
Maintain order book

Chralt98:
### What does it do?

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1116

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1115

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1117


zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+            RuntimeOrigin::signed(ALICE),
+            market_id,
+            maker_asset,
+            10 * BASE,
+            taker_asset,
+            taker_amount,
+        ));
+
+        let bob_free_taker_asset = AssetManager::free_balance(taker_asset, &BOB);
+        assert_eq!(bob_free_taker_asset, 0);
```
maltekliemann, line 295: Ideally, Bob would own one unit less than required instead of nothing.

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+            maker_asset,
+            10 * BASE,
+            taker_asset,
+            taker_amount,
+        ));
+
+        AssetManager::deposit(taker_asset, &BOB, taker_amount).unwrap();
+
+        assert_noop!(
+            Orderbook::fill_order(RuntimeOrigin::signed(BOB), order_id, Some(taker_amount - 1)),
```
maltekliemann, line 329:
```suggestion
            Orderbook::fill_order(RuntimeOrigin::signed(BOB), order_id, Some(taker_amount - AssetManager::minimum_balance(taker_asset) + 1)),

File: zrml/orderbook/src/tests.rs [hunk 3]
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+        let market_id = 0u128;
+        let market = market_mock::<Runtime>();
+        Markets::<Runtime>::insert(market_id, market.clone());
+
+        let maker_asset = market.base_asset;
+        let taker_asset = Asset::CategoricalOutcome(0, 2);
+
+        let order_id = 0u128;
+        let taker_amount = 25 * BASE;
+
```
maltekliemann, line 377:
```suggestion
        let maker_amount = 10 * BASE;

```

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+        assert_ok!(Orderbook::place_order(
+            RuntimeOrigin::signed(ALICE),
+            market_id,
+            maker_asset,
+            10 * BASE,
```
maltekliemann, line 382:
```suggestion
            maker_amount,
```

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+        AssetManager::deposit(taker_asset, &BOB, taker_amount).unwrap();
+
+        let order = <Orders<Runtime>>::get(order_id).unwrap();
+        assert_eq!(
+            order,
+            Order {
+                market_id,
+                maker: ALICE,
+                maker_asset,
+                maker_amount: 10 * BASE,
```
maltekliemann, line 396:
```suggestion
                maker_amount,
```

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+                taker_asset,
+                taker_amount: 25 * BASE,
```
maltekliemann, line 398:
```suggestion
                taker_amount,
```

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+            }
+        );
+
+        assert_ok!(Orderbook::fill_order(
+            RuntimeOrigin::signed(BOB),
+            order_id,
+            Some(taker_amount / 2)
+        ));
+
+        let order = <Orders<Runtime>>::get(order_id).unwrap();
```
maltekliemann, line 408: `Order::<Runtime>::foo`

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+
+        assert_eq!(
```
maltekliemann, line 410: I'd say this test should also verify that the amounts are correctly changed.

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+            taker_amount,
+        ));
+
+        AssetManager::deposit(taker_asset, &BOB, taker_amount).unwrap();
+
+        assert_ok!(Orderbook::fill_order(RuntimeOrigin::signed(BOB), order_id, None));
+
+        assert!(<Orders<Runtime>>::get(order_id).is_none());
+    });
+}
```
maltekliemann, line 363: Why isn't this part of `it_fills_order_fully_maker_outcome_asset`?

zrml/orderbook/src/tests.rs
```diff
@@ -107,367 +541,506 @@ fn it_places_orders() {
...
+        assert_eq!(market_creator_balance_after - market_creator_balance_before, taker_fees);
+
+        let reserved_bob = AssetManager::reserved_balance(maker_asset, &BOB);
         assert_eq!(reserved_bob, 0);
 
         System::assert_last_event(
             Event::<Runtime>::OrderFilled {
                 order_id,
                 maker: BOB,
                 taker: ALICE,
```
maltekliemann, line 629: This event doesn't include which part of the maker amount was traded. I guess we could read that off the the transaction data. Is that standard practice, or should we add the other amounts too, @saboonikhil?

zrml/orderbook/src/tests.rs
```diff
@@ -107,367 +541,506 @@ fn it_places_orders() {
...
         );
 
-        let alice_bal = <Balances as Currency<AccountIdTest>>::free_balance(&ALICE);
-        let alice_shares = Tokens::free_balance(outcome_asset, &ALICE);
-        assert_eq!(alice_bal, BASE - 50);
-        assert_eq!(alice_shares, 10);
+        let alice_maker_asset_free = AssetManager::free_balance(taker_asset, &ALICE);
+        let alice_taker_asset_free = AssetManager::free_balance(maker_asset, &ALICE);
+        assert_eq!(alice_maker_asset_free, INITIAL_BALANCE);
+
```
maltekliemann, line 640:
```suggestion
```

zrml/orderbook/src/tests.rs
```diff
@@ -107,367 +541,506 @@ fn it_places_orders() {
...
-        let reserved_bob = Tokens::reserved_balance(outcome_asset, &BOB);
-        assert_eq!(reserved_bob, 860);
+        let reserved_bob = AssetManager::reserved_balance(maker_asset, &BOB);
+        assert_eq!(reserved_bob, filled_maker_amount);
     });
 }
 
 #[test]
-fn it_removes_orders() {
+fn it_removes_order() {
```
maltekliemann, line 899: This test seems to be mixing multiple tests (success and failure). I think we should split this.

zrml/orderbook/src/mock.rs
```diff
@@ -38,6 +43,41 @@ use zeitgeist_primitives::{
...
+
+    fn distribute(
+        _market_id: Self::MarketId,
+        asset: Self::Asset,
+        account: &Self::AccountId,
+        amount: Self::Balance,
+    ) -> Self::Balance {
+        let fees = Self::get_fee(_market_id, amount);
+        let _ = T::AssetManager::transfer(asset, account, &F::get(), fees);
+        fees
```
maltekliemann, line 77: Looking at this now, this implementation is wrong, isn't it? Because if the transfer fails, then the return value should be zero.

```suggestion
        match T::AssetManager::transfer(asset, account, &F::get(), fees) {
            Ok(_) => fees,
            Err(_) => Zero::zero(),
        }
```

maltekliemann (changes requested):
Don't forget the migration!

runtime/common/src/fees.rs
```diff
@@ -263,14 +267,20 @@ macro_rules! impl_market_creator_fees {
         }
 
         impl MarketCreatorFee {
+            fn do_get_fee(market_id: MarketId, amount: Balance) -> Result<Balance, DispatchError> {
+                let market = MarketCommons::market(&market_id)?; // Should never fail
+                let fee_amount = market.creator_fee.mul_floor(amount);
+                Ok(fee_amount)
+            }
+
```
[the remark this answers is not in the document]
maltekliemann: > Are you satisfied if we use unwrap_or_default() in both instances? Balance implements the default trait using Zero.

Seems fine to me. After all, the function always return the amount of fees deducted. If the transfer fails, no fees change hands, so the function should return zero.

> I don't think your suggestion solves the problem what needs to happen if do_distribute errors.

What's the problem?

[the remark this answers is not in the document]
maltekliemann: Ah, so the problem is that `get_fee` returns the wrong value if `distribute` errors. That's a good argument for always using the return value of `distribute` instead of `get_fee`. If you want to implement `get_fee` correctly, I guess you need to ensure that the transfer will actually go through or return zero.

zrml/orderbook/src/benchmarks.rs
```diff
@@ -50,71 +48,69 @@ fn generate_funded_account<T: Config>(seed: Option<u32>) -> Result<T::AccountId,
...
-
-    fill_order_bid {
-        let caller = generate_funded_account::<T>(None)?;
-        let (_, _, order_id) = place_order::<T>(OrderSide::Bid, Some(0))?;
+    fill_order {
+        let taker_asset = Asset::CategoricalOutcome::<MarketIdOf<T>>(0u32.into(), 0);
+        let caller = generate_funded_account::<T>(None, taker_asset)?;
+        let (_, market_id, order_id) = place_default_order::<T>(Some(0), taker_asset)?;
+        let maker_asset = T::MarketCommons::market(&market_id).unwrap().base_asset;
+        let caller = generate_funded_account::<T>(None, maker_asset)?;
```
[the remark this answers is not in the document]
maltekliemann: It's not a major problem. But even now it's really weird design, because you're now handling two markets IDs and solving the problem by just throwing one of them away.

zrml/orderbook/src/tests.rs
```diff
@@ -80,12 +116,410 @@ fn fill_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
 
+#[test_case(MarketStatus::Proposed; "proposed")]
+#[test_case(MarketStatus::Suspended; "suspended")]
+#[test_case(MarketStatus::Closed; "closed")]
+#[test_case(MarketStatus::CollectingSubsidy; "collecting subsidy")]
+#[test_case(MarketStatus::InsufficientSubsidy; "insufficient subsidy")]
+#[test_case(MarketStatus::Reported; "reported")]
+#[test_case(MarketStatus::Disputed; "disputed")]
+#[test_case(MarketStatus::Resolved; "resolved")]
+fn fill_order_fails_if_market_status_not_active(status: MarketStatus) {
```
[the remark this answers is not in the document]
maltekliemann: I think I put this comment in the wrong place. Sorry. Let me try again.

What I mean is that `remove_order` doesn't require that the market have scoring rule `OrderBook`, while `fill_order` does. In both cases you could argue that it is not necessary to have this requirement because `place_order` won't allow placing orders for markets with the wrong scoring rule to begin with. That's why you have to use `mutate_market` to create a scenario which can't happen in the first place just to test this error.

I think both approaches are valid, but we should stick with one approach for both functions.

zrml/orderbook/src/mock.rs
```diff
@@ -38,6 +43,44 @@ use zeitgeist_primitives::{
...
+        account: &Self::AccountId,
+        amount: Self::Balance,
+    ) -> Self::Balance {
+        let fees = Self::get_fee(_market_id, amount);
+        match T::AssetManager::transfer(asset, account, &F::get(), fees) {
+            Ok(_) => fees,
+            Err(_) => Zero::zero(),
+        };
+        fees
+    }
```
maltekliemann, line 81:
```suggestion
        match T::AssetManager::transfer(asset, account, &F::get(), fees) {
            Ok(_) => fees,
            Err(_) => Zero::zero(),
        }
```

zrml/orderbook/src/tests.rs
```diff
@@ -45,33 +48,69 @@ fn place_order_fails_with_wrong_scoring_rule(scoring_rule: ScoringRule) {
...
+                market_id,
+                market.base_asset,
+                10 * BASE,
+                Asset::CategoricalOutcome(0, 2),
+                25 * BASE,
+            ),
+            Error::<Runtime>::MarketIsNotActive
+        );
+    });
+}
```
maltekliemann, line 91:
```suggestion
}

```

runtime/common/src/fees.rs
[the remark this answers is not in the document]
maltekliemann: You can cherry-pick from this branch: https://github.com/zeitgeistpm/zeitgeist/tree/mkl-orderbook-abstraction

I'd say `get_fee` must either be rewritten as described above _or_ market as only to be used for testing purposes.

zrml/orderbook/src/lib.rs
```diff
@@ -160,199 +174,307 @@ mod pallet {
...
+                    order_data.market_id,
+                    // maker_fill is the amount what the maker wants to have (base asset from taker)
+                    maker_fill,
+                );
+                // base asset amount from taker minus fees to the maker (maker_fill)
+                maker_fill.checked_sub(&external_fees).ok_or(ArithmeticError::Underflow)?
+            };
+
+            Ok(maybe_adjusted_maker_fill)
+        }
```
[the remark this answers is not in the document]
maltekliemann: You can cherry-pick from this branch: https://github.com/zeitgeistpm/zeitgeist/tree/mkl-orderbook-abstraction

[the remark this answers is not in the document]
maltekliemann: Note: Given the fact that `get_fee` currently doesn't always do the right thing (and it may be difficult to ensure that it does), it probably shouldn't be used and `distribute` should be used instead.

zrml/orderbook/src/tests.rs
```diff
@@ -107,367 +534,536 @@ fn it_places_orders() {
...
+        ));
+
+        let order_id = 0u128;
+
         assert_noop!(
             Orderbook::remove_order(RuntimeOrigin::signed(BOB), order_id),
             Error::<Runtime>::NotOrderCreator,
         );
+    });
+}
```
maltekliemann, line 968: This is already covered by `remove_order_fails_if_not_order_creator`.

```suggestion
```

maltekliemann (changes requested):

zrml/orderbook/src/lib.rs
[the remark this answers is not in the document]
Chralt98: I only see the usage in the tests. I think about how we can reach the same thing in a better way.

runtime/common/src/fees.rs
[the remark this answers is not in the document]
Chralt98: 102a742142560efc91718c967f01c95b95cfbdc7

zrml/orderbook/src/benchmarks.rs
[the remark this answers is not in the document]
Chralt98: I admit, I don't really give a high attention to the design of benchmarks as soon as they are quick to write. What do you propose us to do now. Can you suggest something or should we just ignore this issue?

zrml/orderbook/src/tests.rs
Chralt98: 7f88460e725344dfb180576fb785fa9a737a47c8

zrml/orderbook/src/tests.rs
Chralt98: 7f88460e725344dfb180576fb785fa9a737a47c8

zrml/orderbook/src/mock.rs
Chralt98: 7f88460e725344dfb180576fb785fa9a737a47c8

zrml/orderbook/src/tests.rs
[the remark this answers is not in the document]
Chralt98: I still don't get it fully yet. 

So, should we remove the code testing a scenario that wouldn't occur? Or should we add a scoring rule check for remove_order?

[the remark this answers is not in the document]
Chralt98: I would also just remove the check from `fill_order`, because `place_order` prevents this. However I would add a debug_assert in fill order.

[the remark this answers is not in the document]
Chralt98: 53ef53e1197067a5642e8dd3b7d550b17289555a

Chralt98:
I will do the migration in this PR shortly.

zrml/orderbook/src/migrations.rs
Chralt98, line 1: I did test the try-runtime run on main and test net. Unfortunately we have no examples other than my written migration tests here.. Because there is no order with the old structure open at the moment.

zrml/orderbook/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+            // assert old fields
+            assert_eq!(old_order.market_id, new_order.market_id);
+            assert_eq!(old_order.maker, new_order.maker);
+            // assert new fields
+            let reserved = T::AssetManager::reserved_balance_named(
+                &OrderbookPallet::<T>::reserve_id(),
+                new_order.maker_asset,
+                &new_order.maker,
+            );
+            assert!(reserved > BalanceOf::<T>::zero());
```
maltekliemann, line 176:
```suggestion
```

maltekliemann: This is covered by the next `assert_eq`

zrml/orderbook/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+        });
+    }
+
+    fn set_up_version() {
+        StorageVersion::new(ORDER_BOOK_REQUIRED_STORAGE_VERSION).put::<OrderbookPallet<Runtime>>();
+    }
+
+    fn construct_old_new_tuple() -> (Vec<OldOrderOf<Runtime>>, Vec<OrderOf<Runtime>>) {
+        let market_id_0 = 0;
+        let outcome_asset_amoun_0 = 42000;
```
maltekliemann, line 278:
```suggestion
        let outcome_asset_amount_0 = 42000;
```

Etc.

zrml/orderbook/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+            maker: 1,
+            // the maker reserved the base asset for order side bid
+            maker_asset: Asset::Ztg,
+            maker_amount: base_asset_amount_0,
+            taker_asset: Asset::CategoricalOutcome(market_id_0, 0u16),
+            taker_amount: outcome_asset_amoun_0,
+        };
+
+        let market_id_1 = 1;
+        let outcome_asset_amoun_1 = 42000;
```
maltekliemann, line 300: same here

zrml/orderbook/src/benchmarks.rs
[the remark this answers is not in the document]
maltekliemann: No just leave it

zrml/orderbook/src/migrations.rs
```diff
@@ -0,0 +1,336 @@
...
+            );
+            TranslateOrderStructure::<Runtime>::on_runtime_upgrade();
+
+            let actual = crate::Orders::<Runtime>::get(0).unwrap();
+            assert_eq!(actual, new_orders[0]);
+
+            let actual = crate::Orders::<Runtime>::get(1).unwrap();
+            assert_eq!(actual, new_orders[1]);
+        });
+    }
```
maltekliemann, line 270: This is a duplicate of the previous test.

maltekliemann (changes requested):

zrml/orderbook/src/migrations.rs
maltekliemann: Nevermind, I can't read.

zrml/orderbook/README.md
```diff
@@ -1,3 +1,46 @@
...
+asset for outcome assets and vice versa.
+
+## Overview
+
+The order book can be set as a market's scoring rule. It allows to place,
+partially or fully fill and remove orders.
+
+## Terminology
+
+- `maker_partial_fill`: The partial amount of what the maker wants to fill.
```
Robiquet, line 13: How can a maker fill an order? Wouldn't that make them a taker?

Chralt98: Haha, good point @Robiquet .. Malte already pointed this out extensively. Good that you found another instance for this.

zrml/neo-swaps/src/mock.rs
```diff
@@ -140,12 +140,9 @@ where
...
-            amount.saturated_into(),
-            EXTERNAL_FEES.saturated_into(),
-        )
-        .unwrap()
-        .saturated_into();
+        let fees = zeitgeist_primitives::math::fixed::bmul(amount.saturated_into(), EXTERNAL_FEES)
+            .unwrap()
+            .saturated_into();
         let _ = T::MultiCurrency::transfer(asset, account, &F::get(), fees);
         fees
```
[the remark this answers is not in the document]
Chralt98: 1b3fc077ce0661258af610ee8691a14adb712c06

zrml/orderbook/src/benchmarks.rs
```diff
@@ -24,97 +24,87 @@
...
 }
 
-// Creates an account and gives it asset and currency. `seed` specifies the account seed,
-// None will return a whitelisted account
-// Returns `account`, `asset`, `outcome_asset_amount` and `base_asset_amount`
 fn order_common_parameters<T: Config>(
     seed: Option<u32>,
 ) -> Result<
-    (T::AccountId, Asset<MarketIdOf<T>>, BalanceOf<T>, BalanceOf<T>, MarketIdOf<T>),
+    (MarketIdOf<T>, T::AccountId, Asset<MarketIdOf<T>>, BalanceOf<T>, BalanceOf<T>),
```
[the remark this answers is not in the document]
Chralt98: I would say I changed the order, because the market_id seemed to be the most fundamental value.

zrml/orderbook/src/migrations.rs
Chralt98: 186b1d2ac14187f85bb95f25c37019c501fb6122

zrml/orderbook/src/migrations.rs
Chralt98: 41897a068921646edc68920f363fd7f048407350

zrml/orderbook/src/migrations.rs
Chralt98: 623e08f62b1864b2b7a60cfb30e3bb657ebdbb57

zrml/orderbook/README.md
Chralt98: a0e0f21bb21a7e1044cfaa4ae95d23329a30a394

zrml/orderbook/src/tests.rs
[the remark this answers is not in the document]
Chralt98: 011547635e993322cf4be2e5bde34c6a7a7b0790

zrml/orderbook/src/migrations.rs
Chralt98: I created some orders and used try-runtime again. I had to adjust the try-runtime post-upgrade function. After this everything worked as expected! 

209adc629f122f6b33a016a5557ff8d248bd951f

zrml/orderbook/src/migrations.rs
```diff
@@ -0,0 +1,335 @@
...
+                new_orders.clone(),
+            );
+            TranslateOrderStructure::<Runtime>::on_runtime_upgrade();
+
+            let actual = crate::Orders::<Runtime>::get(0).unwrap();
+            assert_eq!(actual, new_orders[0]);
+
+            let actual = crate::Orders::<Runtime>::get(1).unwrap();
+            assert_eq!(actual, new_orders[1]);
+        });
```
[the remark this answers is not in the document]
maltekliemann: Funny thing is - this is very difficult to figure out because `cargo test` eats `log::info!`, even with `--nocapture` enabled.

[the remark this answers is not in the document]
Chralt98: c19a2e5215b4f73d44e5c60c002599081bd78c42

zrml/orderbook/src/migrations.rs
```diff
@@ -251,20 +253,14 @@ mod tests {
...
     fn on_runtime_upgrade_is_noop_if_versions_are_not_correct() {
         ExtBuilder::default().build().execute_with(|| {
-            // Don't set up chain to signal that storage is already up to date.
-            let (_, new_orders) = construct_old_new_tuple();
-            populate_test_data::<Twox64Concat, OrderId, OrderOf<Runtime>>(
-                ORDER_BOOK,
-                ORDERS,
-                new_orders.clone(),
-            );
-            TranslateOrderStructure::<Runtime>::on_runtime_upgrade();
```
maltekliemann: You should leave the orders in to ensure they don't get messed up by the migration.

maltekliemann (approved):

zrml/orderbook/src/migrations.rs
Chralt98: I thought about this as I were making this change but came to the conclusion, that the `StorageVersion` would be changed in any case. So that's why I didn't care about the orders. If the storage root is the same before and after the migration and I know that the `StorageVersion` is changed in any case, I know that the orders are also not changed.

maltekliemann (approved):

zrml/orderbook/src/lib.rs
```diff
@@ -164,181 +178,274 @@ mod pallet {
...
-                OrderSide::Ask => order_data.base_asset_amount,
-            };
-            let maker_fill = maker_partial_fill.unwrap_or(makers_requested_total);
-            ensure!(!maker_fill.is_zero(), Error::<T>::AmountIsZero);
-            ensure!(maker_fill <= makers_requested_total, Error::<T>::AmountTooHighForOrder);
+            let maker_fill = maker_partial_fill.unwrap_or(order_data.taker_amount);
+            ensure!(
+                maker_fill >= T::AssetManager::minimum_balance(order_data.taker_asset),
+                Error::<T>::BelowMinimumBalance
+            );
```
sea212, line 403: Does it mean an order gets stuck once the remaining balance the maker wants is below the minimum balance? In that case I recommend moving this check to when the actual transferal of the funds happens, because in most cases the maker will already have some balance for the requested token and fulfill this requirement.

zrml/orderbook/src/lib.rs
```diff
@@ -164,181 +178,274 @@ mod pallet {
...
+            ensure!(maker_fill <= order_data.taker_amount, Error::<T>::AmountTooHighForOrder);
```
sea212, line 404: Just as a side note, I find the naming of the maker and taker information very confusing, as it depends on the perspective that is being looked from. I would recommend sticking to an absolute view, for example `taker_fill` and `taker_amount` always refer to the token that the maker wants.

sea212: This occurs in multiple places in this file.

zrml/orderbook/src/lib.rs
```diff
@@ -164,181 +178,274 @@ mod pallet {
...
 
+            // clone required because of mutable borrow of order_data below
```
sea212, line 406:
```suggestion
```

zrml/orderbook/README.md
```diff
@@ -1,3 +1,46 @@
...
+    are charged from the taker after the transaction is executed. In particular,
+    the maker still receives the full amount of outcome tokens.
+
+## Interface
+
+### Dispatches
+
+#### Public Dispatches
+
+- `remove_order`: Allows a user to remove their order from the order book.
```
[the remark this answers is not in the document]
Chralt98: ChatGPT: In this sentence, "user" is a singular noun, so you should use "a user" rather than "an user." "An" is used before words that begin with a vowel sound, while "a" is used before words that begin with a consonant sound. Since "user" begins with a consonant sound ("yoo"), you should use "a user."

zrml/orderbook/src/lib.rs
Chralt98: It just means that a taker, who uses `fill_order` is not allowed to partially fill an order to the extend that there is an amount missing that is below the existential deposit. We ensure somewhere else in the code that the `taker_amount` stays above the existential deposit.

Chralt98: [Here](https://github.com/zeitgeistpm/zeitgeist/blob/7052aefccd264f15d749405764777e9f2e303386/zrml/orderbook/src/lib.rs#L493-L496) we check that `taker_amount` is above the existential deposit. [Here](https://github.com/zeitgeistpm/zeitgeist/blob/7052aefccd264f15d749405764777e9f2e303386/zrml/orderbook/src/lib.rs#L311) we ensure the `taker_amount` is also above the existential deposit in a (partial) fill.

zrml/orderbook/src/lib.rs
Chralt98: I don't really understand how to achieve this. I know it's at first hard to wrap the head around it. 

```
The `maker_partial_fill` is the partial amount of what the maker wants to get filled.
```

zrml/orderbook/src/lib.rs
```diff
@@ -164,181 +178,274 @@ mod pallet {
...
+
+        fn do_remove_order(order_id: OrderId, who: AccountIdOf<T>) -> DispatchResult {
+            let order_data = <Orders<T>>::get(order_id).ok_or(Error::<T>::OrderDoesNotExist)?;
+            debug_assert!(
+                {
+                    let market = T::MarketCommons::market(&order_data.market_id)?;
+                    market.scoring_rule == ScoringRule::Orderbook
+                },
+                "The call to place_order already ensured the scoring rule order book."
+            );
```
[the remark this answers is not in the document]
Chralt98: Ok, I generally agree that I should reconsider my `debug_assert` usage. I just thought it would be a good way to tell the reader of the code that this assumption is true. `debug_assert` is also an excellent way to have no overhead in the production runtime, since its logic is not included in those builds. We also discussed exactly this code snippet already in the review from Malte. Before using `debug_assert`, I used a normal `ensure!` statement, but Malte suggested that we shouldn't include checks that wouldn't ever happen in production (because we ensure each order created does have the scoring rule order book). Your suggestion comes with the drawback that we would need to runtime overhead in production builds that we wouldn't necessarily need. So at the end, I am in favour of deleting this check overall. What would you do @maltekliemann ?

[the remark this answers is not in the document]
maltekliemann: > but Malte suggested that we shouldn't include checks that wouldn't ever happen in production

I think (hope) that I suggested that you be consistent in using or not using this pattern. In terms of software design, leaving the check out is preferred, in terms of defensive programming, leaving the check in is preferred. Either way is fine.

> The runtime review from SRLabs also expressed that debug_assert in production is a bad practice and should be avoided.

When/if SRLabs publishes a best practices, I'm happing to go over our style guide with you guys and incorporate these. I think there are some other things that we may want to add there.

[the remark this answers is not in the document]
Chralt98: > I'm happing to go over our style guide with you guys and incorporate these. I think there are some other things that we may want to add there.

Happy to apply best practices and agree on that.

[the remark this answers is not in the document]
Chralt98: 537725ec679acb73e5340c7de68ff005f9d3bee3

zrml/orderbook/README.md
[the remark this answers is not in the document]
sea212: Interesting. Learned something new today :smile:

zrml/orderbook/src/lib.rs
sea212: I would just always use `maker` if it affects the maker token and always use `taker` if it affects the taker token.

sea212 (approved):

Issue #1164, opened 2023-10-25
Allow the market creator of a trusted market to close the market at any time


Chralt98:
Related to https://github.com/zeitgeistpm/zeitgeist/pull/1127

PR #1184, opened 2023-11-09
Implement trusted market close

Chralt98:
### What does it do?

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/1164

### DIFF
--- a/zrml/prediction-markets/src/benchmarks.rs
+++ b/zrml/prediction-markets/src/benchmarks.rs
@@ -110,6 +110,46 @@ fn create_market_common<T: Config + pallet_timestamp::Config>(
     Ok((caller, market_id))
 }
 
+fn create_trusted_market<T: Config + pallet_timestamp::Config>(
+    creation: MarketCreation,
+    options: MarketType,
+    scoring_rule: ScoringRule,
+    period: Option<MarketPeriod<T::BlockNumber, MomentOf<T>>>,
+) -> Result<(T::AccountId, MarketIdOf<T>), &'static str> {
+    pallet_timestamp::Pallet::<T>::set_timestamp(0u32.into());
+    let range_start: MomentOf<T> = 100_000u64.saturated_into();
+    let range_end: MomentOf<T> = 1_000_000u64.saturated_into();
+    let creator_fee: Perbill = Perbill::zero();
+    let period = period.unwrap_or(MarketPeriod::Timestamp(range_start..range_end));
+    let caller: T::AccountId = whitelisted_caller();
+    T::AssetManager::deposit(Asset::Ztg, &caller, (100u128 * LIQUIDITY).saturated_into()).unwrap();
+    let oracle = caller.clone();
+    let deadlines = Deadlines::<T::BlockNumber> {
+        grace_period: 1_u32.into(),
+        oracle_duration: T::MinOracleDuration::get(),
+        dispute_duration: Zero::zero(),
+    };
+    let mut metadata = [0u8; 50];
+    metadata[0] = 0x15;
+    metadata[1] = 0x30;
+    let metadata = MultiHash::Sha3_384(metadata);
+    Call::<T>::create_market {
+        base_asset: Asset::Ztg,
+        creator_fee,
+        oracle,
+        period,
+        deadlines,
+        metadata,
+        creation,
+        market_type: options,
+        dispute_mechanism: None,
+        scoring_rule,
+    }
+    .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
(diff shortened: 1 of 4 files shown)


zrml/prediction-markets/src/benchmarks.rs
```diff
@@ -110,6 +110,46 @@ fn create_market_common<T: Config + pallet_timestamp::Config>(
...
+        creation,
+        market_type: options,
+        dispute_mechanism: None,
+        scoring_rule,
+    }
+    .dispatch_bypass_filter(RawOrigin::Signed(caller.clone()).into())?;
+    let market_id = <zrml_market_commons::Pallet<T>>::latest_market_id()?;
+    Ok((caller, market_id))
+}
+
```
maltekliemann, line 152: I think instead of adding this function, you should just add a `dispute_mechanism` parameter to `create_market`.l

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1184/commits/6aba68d6abecbe255b3158eabc27af8774144c88

zrml/prediction-markets/src/lib.rs
```diff
@@ -1687,6 +1687,38 @@ mod pallet {
...
+        ///
+        /// Complexity: `O(n + m)`, where `n` is the number of market ids,
+        /// which open at the same time as the specified market,
+        /// and `m` is the number of market ids,
+        /// which close at the same time as the specified market.
+        //
+        // ***** IMPORTANT *****
+        //
+        // Within the same block, operations that interact with the activeness of the same
+        // market will behave differently before and after this call.
```
maltekliemann, line 1704: I mean... yeah. Doesn't that go without saying?

```suggestion
```

zrml/prediction-markets/src/tests.rs
```diff
@@ -6342,6 +6342,94 @@ fn trusted_market_complete_lifecycle() {
...
+            PredictionMarkets::close_trusted_market(RuntimeOrigin::signed(BOB), market_id),
+            Error::<Runtime>::CallerNotMarketCreator
+        );
+
+        assert_ok!(PredictionMarkets::close_trusted_market(
+            RuntimeOrigin::signed(market_creator),
+            market_id
+        ));
+        let market = MarketCommons::market(&market_id).unwrap();
+        assert_eq!(market.status, MarketStatus::Closed);
```
maltekliemann, line 6387: Can we add a test here that verifies that the market end has been set correctly?

maltekliemann, line 6387: Do we usually test if the IDs have been cleared properly from the cache? If so, we should have these tests here, as well.

zrml/prediction-markets/src/lib.rs
```diff
@@ -2688,6 +2724,15 @@ mod pallet {
...
 
+        fn ensure_creator_of_trusted_market(
+            market: &MarketOf<T>,
+            who: &T::AccountId,
+        ) -> DispatchResult {
+            ensure!(market.creator == *who, Error::<T>::CallerNotMarketCreator);
+            ensure!(market.dispute_mechanism.is_none(), Error::<T>::MarketIsNotTrusted);
+            Ok(())
+        }
+
```
maltekliemann, line 2735: Why are we abstracting this into a separate function? Seems like a very specific thing to verify.

zrml/prediction-markets/src/tests.rs
```diff
@@ -6342,6 +6342,94 @@ fn trusted_market_complete_lifecycle() {
...
+        assert_noop!(
+            PredictionMarkets::close_trusted_market(
+                RuntimeOrigin::signed(market_creator),
+                market_id
+            ),
+            Error::<Runtime>::MarketIsNotTrusted
+        );
+    });
+}
+
```
maltekliemann, line 6432: We're still missing a test that the call fails if the market is not active.

maltekliemann (commented):
Looking good, just requesting some small changes

zrml/prediction-markets/src/lib.rs
Chralt98: 4f591cf0660d9d0c836957aeead58e1ddebcf97b

zrml/prediction-markets/src/tests.rs
Chralt98: 62237c347a9f7b9c98384cb74b8546f0a1cf14a8

zrml/prediction-markets/src/tests.rs
Chralt98: 6e4ac4715bb6a9906ebb65e63e609fb6d1a74028

maltekliemann (approved):

Issue #955, opened 2023-01-23
Add `debug_assert` statements whereever unreserves/slashes are used


maltekliemann:
Place `debug_assert` statements wherever unreserves/slashes are used to ensure that the unreserved/slashed amount was actually available. This signals to other developers that we expect that unreserve/slash to always get the entire requested amount and it is useful in testing. Instead of repeating the same (somewhat intuitive) test code involving the sentinel amount over and over again, we don't have to worry about verifying that we didn't "over"-unreserve/slash.

PR #1188, opened 2023-11-10
Add debug assertions for slashes and reserves

Chralt98:
### What does it do?

Fixes https://github.com/zeitgeistpm/zeitgeist/issues/955


maltekliemann (approved):
LGTM, but there's no consistent full stop at the end of the error message.

PR #1189, opened 2023-11-10
New Asset System - Implement asset-router

sea212:
### What does it do?
- Adds a new pallet called `zrml-asset-router`
- The asset router is configured by supplying three pallet-asset instances and one orml-tokens instance
- The asset router does implement the orml traits `MultiCurrency`, `MultiCurrencyExtended`, `MultiLockableCurrency`, `MultiReservableCurrency`, `NamedMultiReservableCurrency`
- The asset router does implement the frame-support traits `Inspect`, `Create` and `Destroy`

- pallet-assets does not provide reserve and lock functionality. While reserving is technically possible using a dispatchable call, it is not recommended to do as the whole asset class could be destroyed by the admin while the reserver assumes the funds are still in reserve
- https://github.com/zeitgeistpm/zeitgeist/issues/1198
- Rikiddo was deprecated with this PR. Resolving Rikiddo pools will not work properly, since no share holders are assumed. This restriction comes from the circumstance that the code was iterating through every share holder account, but this is not possible with pallet-assets. https://github.com/zeitgeistpm/zeitgeist/issues/1199

  - Lazy migrate existing market assets: Pallets will continue using `orml-tokens` for existing market until all those existing market assets expired. Reason: Too many accounts hold existing market assets, migrating will be expensive and cannot be executed atomically. During migration, the storage trie that hold old market asset information will most likely be manipulated by user interaction.
  - Paying fees using `CampaignAssets`.
  - Using `CampaignAssets` and `CustomAssets` as base assets in prediction markets.
  - Properly creating and destroying `MarketAssets` in pallets such as prediction markets.
  - Adjust Asset types in many pallets, such as xcm related pallets (should only allow `ZTG` and `ForeignAsset(id)` for now), orml-tokens, etc.
  - Future: Remove outcome asset classes from the old asset structure once all associated outcome tokens expired.

- Implementing reserves and locks for pallet-assets - unfeasible
- Replacing orml-currencies instead of integrating the asset-router into it. Has it's advantages, but is not necessary at this point.

- Continuation of https://github.com/zeitgeistpm/zeitgeist/pull/1181


zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+                by_amount.abs()
+            };
+
+            let by_balance = TryInto::<Self::Balance>::try_into(by_amount_abs)
+                .map_err(|_| Error::<T>::AmountIntoBalanceFailed)?;
+            if by_amount.is_positive() {
+                Self::deposit(currency_id, who, by_balance)
+            } else {
+                Self::withdraw(currency_id, who, by_balance).map(|_| ())
+            }
```
Chralt98, line 420: I don't fully get this.

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+
+        fn reserved_balance(currency_id: Self::CurrencyId, who: &T::AccountId) -> Self::Balance {
+            only_currency!(
+                currency_id,
+                Zero::zero(),
+                MultiReservableCurrency,
+                reserved_balance,
+                who
+            )
+        }
```
Chralt98, line 497: Silently suppressing the error from `only_currency` seems concerning. Sure, this is not called directly by user, rather by us developers. I am suggesting to log an error and using `debug_assert` to fail for tests.

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+
+        fn unreserve(
+            currency_id: Self::CurrencyId,
+            who: &T::AccountId,
+            value: Self::Balance,
+        ) -> Self::Balance {
+            only_currency!(currency_id, value, MultiReservableCurrency, unreserve, who, value)
+        }
```
Chralt98, line 520: Silently suppressing the error from `only_currency` seems concerning. Sure, this is not called directly by user, rather by us developers. I am suggesting to log an error and using `debug_assert` to fail for tests.

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            who: &T::AccountId,
+        ) -> Self::Balance {
+            if let Ok(currency) = T::CurrencyType::try_from(currency_id) {
+                return <T::Currencies as NamedMultiReservableCurrency<T::AccountId>>::reserved_balance_named(
+                    id, currency, who,
+                );
+            }
+
+            Zero::zero()
+        }
```
Chralt98, line 558: silently suppressed

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            value: Self::Balance,
+        ) -> Self::Balance {
+            if let Ok(currency) = T::CurrencyType::try_from(currency_id) {
+                return <T::Currencies as NamedMultiReservableCurrency<T::AccountId>>::unreserve_named(
+                    id, currency, who, value
+                );
+            }
+
+            value
+        }
```
Chralt98, line 588: silently suppressed

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            value: Self::Balance,
+        ) -> Self::Balance {
+            if let Ok(currency) = T::CurrencyType::try_from(currency_id) {
+                return <T::Currencies as NamedMultiReservableCurrency<T::AccountId>>::slash_reserved_named(
+                    id, currency, who, value
+                );
+            }
+
+            value
+        }
```
Chralt98, line 603: silently suppressed

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+        ) -> WithdrawConsequence<Self::Balance> {
+            only_asset!(
+                asset,
+                WithdrawConsequence::UnknownAsset,
+                Inspect,
+                can_withdraw,
+                who,
+                amount
+            )
+        }
```
Chralt98, line 678: silently suppressed

zrml/asset-router/src/tests/create.rs
```diff
@@ -0,0 +1,67 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        assert_ok!(AssetRouter::create(CAMPAIGN_ASSET, ALICE, true, CAMPAIGN_ASSET_MIN_BALANCE));
+        assert_ok!(AssetRouter::create(CUSTOM_ASSET, ALICE, true, CAMPAIGN_ASSET_MIN_BALANCE));
+        assert_ok!(AssetRouter::create(MARKET_ASSET, ALICE, true, CAMPAIGN_ASSET_MIN_BALANCE));
+        assert!(AssetRouter::asset_exists(MARKET_ASSET));
+        assert!(AssetRouter::asset_exists(CAMPAIGN_ASSET));
+        assert!(AssetRouter::asset_exists(CUSTOM_ASSET));
+        assert!(!AssetRouter::asset_exists(CURRENCY));
+    });
+}
```
Chralt98, line 67: Because of `create_routes_currencies_correctly` I would assume you check for different balances created for the asset classes afterwards instead of just calling `asset_exists`, if you wanna check if routing is not messed up.

zrml/prediction-markets/src/lib.rs
```diff
@@ -3442,7 +3447,11 @@ mod pallet {
                 Asset::Ztg => true,
                 #[cfg(feature = "parachain")]
                 Asset::ForeignAsset(fa) => {
-                    if let Some(metadata) = T::AssetRegistry::metadata(&Asset::ForeignAsset(fa)) {
+                    if let Some(metadata) =
+                        T::AssetRegistry::metadata(&CurrencyClass::<MarketIdOf<T>>::ForeignAsset(
+                            fa,
+                        ))
+                    {
```
Chralt98, line 3454: [Old structure](https://github.com/zeitgeistpm/zeitgeist/blob/4628f8c0af5f44eab01add840fee8e7b78aced1e/primitives/src/assets/currencies.rs#L21-L42) for orml_asset_registry used here, because we store the old structure in storage of orml_asset_registry, right?

zrml/prediction-markets/src/tests.rs
```diff
@@ -1817,7 +1817,7 @@ fn it_allows_to_sell_a_complete_set() {
         // Check the outcome balances
         let assets = PredictionMarkets::outcome_assets(0, &market);
         for asset in assets.iter() {
-            let bal = Tokens::free_balance(*asset, &BOB);
+            let bal = AssetManager::free_balance(*asset, &BOB);
```
Chralt98, line 1820: Why not using the router here, since outcome assets are assumed to be handled by pallet_assets?

Chralt98: Consequently asked for every instance in this file handling outcome assets.

Chralt98 (commented):
Pretty generic. Much to learn from that. Thanks for your work, Harald.

zrml/prediction-markets/src/tests.rs
sea212: `AssetManager` internally uses `AssetRouter` for anything that is not ZTG.

zrml/prediction-markets/src/lib.rs
sea212: `CurrencyClass` is the new and only structure used in orml-tokens. `Old` variants are only supported due to backward compatibility. 
AssetRegistry` only has `Asset::ForeignAsset(id)` (scale index 5) in it's storage. Anything that is read from the `AssetRegistry` will result in `CurrencyClass<MarketId>::ForeignAsset(id)` (or in short, `Currencies::ForeignAsset(id)` (scale index 5).

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            }
+        }
+
+        fn can_slash(
+            currency_id: Self::CurrencyId,
+            who: &T::AccountId,
+            value: Self::Balance,
+        ) -> bool {
+            // TODO
+            if let Ok(currency) = T::CurrencyType::try_from(currency_id) {
```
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/0f1374352498c7eb5c75c4b0f1ea6b41ade456bc

zrml/asset-router/src/tests/create.rs
sea212: Those conditions are covered in https://github.com/zeitgeistpm/zeitgeist/blob/0f1374352498c7eb5c75c4b0f1ea6b41ade456bc/zrml/asset-router/src/tests/multi_currency.rs#L51-L111

zrml/asset-router/src/lib.rs
sea212: the `Amount` type is [`Signed`](https://github.com/zeitgeistpm/open-runtime-module-library/blob/master/currencies/src/lib.rs#L686), but the non currency asset implementation only provide updating balances via the `Mutate` trait, which requires a signed number.

The most common signed number representation utilizes the [Two's complement](https://en.wikipedia.org/wiki/Two%27s_complement). Effectively, more negative numbers can be represented and positive numbers, to be more precise `max_neg = -max_pos - 1`. To avoid overflows, i.e. converting a negative number using `abs()` into a positive number that is out of the range of possible positive numbers of the `Signed` type, a case distinction is used. Whenever `by_amount` is `max_neg`, `Amount::max_value()` is used instead, which results in `max_pos` (`abs(max_neg) - 1`).

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+
+        fn total_issuance(currency_id: Self::CurrencyId) -> Self::Balance {
+            let total_issuance = route_call!(currency_id, total_issuance, total_issuance,);
+            total_issuance.unwrap_or(0u8.into())
+        }
+
+        fn total_balance(currency_id: Self::CurrencyId, who: &T::AccountId) -> Self::Balance {
+            let total_balance = route_call!(currency_id, total_balance, balance, who);
+            total_balance.unwrap_or(0u8.into())
+        }
```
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            } else if let Ok(asset) = T::MarketAssetType::try_from(currency_id) {
+                T::MarketAssets::reducible_balance(asset, who, false)
+            } else if let Ok(asset) = T::CampaignAssetType::try_from(currency_id) {
+                T::CampaignAssets::reducible_balance(asset, who, false)
+            } else if let Ok(asset) = T::CustomAssetType::try_from(currency_id) {
+                T::CustomAssets::reducible_balance(asset, who, false)
+            } else {
+                0u8.into()
+            }
+        }
```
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

zrml/asset-router/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/eeb43e36a666a8be3bde65f0000f78cabb1f43ce

Chralt98 (approved):

Chralt98 (approved):

zrml/asset-router/src/tests/inspect.rs
```diff
@@ -0,0 +1,99 @@
...
+        assert_eq!(AssetRouter::reducible_balance(CURRENCY, &ALICE, false), 0);
+        assert_eq!(
+            AssetRouter::can_withdraw(CURRENCY, &ALICE, CURRENCY_INITIAL_AMOUNT),
+            WithdrawConsequence::UnknownAsset
+        );
+        assert_eq!(
+            AssetRouter::can_deposit(CURRENCY, &ALICE, 1, true),
+            DepositConsequence::UnknownAsset
+        );
+    });
```
maltekliemann, line 98: I would argue that this is not correct behavior. If all functions would throw, this would be fine, but with `total_issuance` returning zero, it feels a bit odd. What's the rationale for not implementing `Inspect` by routing to `MultiCurrency` when you get a `Currency` instance?

maltekliemann (commented):
Comments on the tests can be considered optional in case we're short on time, but I feel like the tests are weird. If you want to test against hte ORML and Parity implementations, then I would suggest we copy their tests instead of creating an incomplete test suite of our own. But I would opt for testing the correct routing using a mock. Or both.

There's also the deal with the `saturated_abs` implementation which has me a little bit confused. The documentation of `num_traits` seems wrong, as it claims that `::MIN` is mapped to `::MIN` (strange choice, would have expected `::MAX`, if anything), but the implemenation seems to panick. Not sure if I'm just being dense here. Using `::MAX` seems like a bad choice since we're "rounding" and the error of the rounding effect is only small if the range of the numerical type is symmetric.

zrml/asset-router/src/tests/destroy.rs
```diff
@@ -0,0 +1,101 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        assert_ok!(<AssetRouter as orml_traits::MultiCurrency<AccountId>>::deposit(
+            CURRENCY,
+            &ALICE,
+            CURRENCY_INITIAL_AMOUNT
+        ));
+        assert_err!(AssetRouter::start_destroy(CURRENCY, None), Error::<Runtime>::Unsupported);
+        assert_err!(AssetRouter::destroy_accounts(CURRENCY, 100), Error::<Runtime>::Unsupported);
+        assert_err!(AssetRouter::destroy_approvals(CURRENCY, 100), Error::<Runtime>::Unsupported);
+        assert_err!(AssetRouter::finish_destroy(CURRENCY), Error::<Runtime>::Unsupported);
```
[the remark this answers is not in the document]
maltekliemann: I think we can replace all `assert_err` with `assert_noop`.

maltekliemann:
You should also make yourself code owner of the new pallet.

zrml/asset-router/README.md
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/ff059c19afc27defba6d93f2b080d8526c9a3631

zrml/asset-router/src/tests/destroy.rs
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/62a1bd9e841fc262a28bab86368ec0814b417d31

primitives/src/assets/currencies.rs
```diff
@@ -17,10 +17,13 @@
 
 use super::*;
 
-/// The `CurrencyClass` enum represents all non-ztg currencies.
+/// The `CurrencyClass` enum represents all non-ztg CurrencyClass
+// used in orml-tokens
```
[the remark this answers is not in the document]
sea212: Brain :spaghetti:. https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/b038088e297ba113f5e1f8ec5df721138c74bea5

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,765 @@
...
+            who: &T::AccountId,
+            amount: Self::Balance,
+        ) -> DispatchResult {
+            if let Ok(currency) = T::CurrencyType::try_from(currency_id) {
+                return <T::Currencies as MultiCurrency<T::AccountId>>::ensure_can_withdraw(
+                    currency, who, amount,
+                );
+            }
+
+            let withdraw_reason = if let Ok(asset) = T::MarketAssetType::try_from(currency_id) {
```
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/10f62f678879a06b1fc27a5c39ac0629e6469663

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,765 @@
...
+            }
+
+            if by_amount.is_zero() {
+                return Ok(());
+            }
+
+            // Ensure this doesn't overflow. There isn't any traits that exposes
+            // `saturating_abs` so we need to do it manually.
+            let by_amount_abs = if by_amount == Self::Amount::min_value() {
+                Self::Amount::max_value()
```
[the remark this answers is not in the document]
sea212: Correct, there is an assumption about how signed numbers are implemented. More about that at https://github.com/zeitgeistpm/zeitgeist/pull/1189#discussion_r1455313608

The code now returns an error instead. Tests ensure that overflow handling works properly. https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/b4d365c9b6bf375298c00ca88f57813c98884fe3

zrml/asset-router/src/tests/multi_currency.rs
```diff
@@ -0,0 +1,111 @@
...
+    assert!(AssetRouter::can_slash(asset, &ALICE, 1));
+    assert_eq!(AssetRouter::slash(asset, &ALICE, 1), 0);
+    assert_eq!(AssetRouter::free_balance(asset, &ALICE), initial_amount - min_balance - 2);
+    assert_ok!(AssetRouter::update_balance(
+        asset,
+        &ALICE,
+        <AssetRouter as MultiCurrencyExtended<AccountId>>::Amount::from(1u8)
+            - <AssetRouter as MultiCurrencyExtended<AccountId>>::Amount::from(2u8)
+    ));
+    assert_eq!(AssetRouter::free_balance(asset, &ALICE), initial_amount - min_balance - 3);
```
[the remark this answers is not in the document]
sea212: This is not about testing the underlying functionality in terms of completeness. The tests assume that the underlying functionality which is invoked is working properly.
The tests should only ensure that routing happens properly. If those tests are inspected in the context that they are invoked in, it makes perfect sense that proper routing is tested.

For example, for campaign assets those conditions are tested after execution of the referenced code above:
```rust
        assert_eq!(AssetRouter::total_issuance(CUSTOM_ASSET), 0);
        assert_eq!(AssetRouter::total_issuance(MARKET_ASSET), 0);
        assert_eq!(AssetRouter::total_issuance(CURRENCY), 0);
```

If those conditions are true, it must mean that all the calls above have been routed to the correct pallet, consequently routing works as expected. The only improvement that might be reasonable here from my perspective is to call `total_issuance` directly in the correct pallet instead of using the router.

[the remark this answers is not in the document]
sea212: Added balance tests for Bob: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/bcaef0e5273dffa3def7485281953ebee267a9fa

zrml/asset-router/src/tests/create.rs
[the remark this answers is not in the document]
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/5ebca55e826b6904789eed6863c34ccec57762f7

zrml/asset-router/src/tests/destroy.rs
```diff
@@ -0,0 +1,101 @@
...
+    assert_ok!(<AssetRouter as orml_traits::MultiCurrency<AccountId>>::deposit(
+        asset,
+        &ALICE,
+        initial_amount
+    ));
+    assert_ok!(AssetRouter::start_destroy(asset, None));
+    assert_eq!(AssetRouter::destroy_accounts(asset, 100), Ok(1));
+    assert_eq!(AssetRouter::destroy_approvals(asset, 100), Ok(1));
+    assert_ok!(AssetRouter::finish_destroy(asset));
+    assert!(!AssetRouter::asset_exists(asset));
```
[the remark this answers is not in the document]
sea212: I think this is fine. We know through other tests that `create` creates the desired asset only in the correct pallet, i.e. routing works properly. Based on that assumption, I think it's perfectly reasonable if the routing of `Destroy` leads to the end result that the asset does not exist anymore, given that is was verified previously that it exists.
What exactly are you missing here?

zrml/asset-router/src/tests/multi_currency.rs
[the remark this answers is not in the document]
sea212: This should improve precision of the routing tests: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/396fed6b6c3ca22a5406388f3d69109bd395dd03 . Since it is now verified directly using the correct pallet that the total issuance is indeed correct for every pallet, the `MultiCurrency` routing tests can be considered as working properly.

Please note that the tests build upon each other, e.g. other tests assume that routing for the `MultiCurrency` trait have properly been verified.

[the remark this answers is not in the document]
sea212: Also https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/f82addfe1684ed68cdce372a26f310a6f4ee7bc3

zrml/asset-router/src/tests/inspect.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1189/commits/cf6af0770a1e666f956c994db2fab0a0d0d9d8a4

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+        type Balance = T::Balance;
+
+        fn total_issuance(asset: Self::AssetId) -> Self::Balance {
+            route_call!(asset, total_issuance, total_issuance,).unwrap_or(Zero::zero())
+            // only_asset!(asset, Zero::zero(), Inspect, total_issuance,)
+        }
+
+        fn minimum_balance(asset: Self::AssetId) -> Self::Balance {
+            route_call!(asset, minimum_balance, minimum_balance,).unwrap_or(Zero::zero())
+            //only_asset!(asset, Zero::zero(), Inspect, minimum_balance,)
```
maltekliemann, line 658:
```suggestion
```

zrml/asset-router/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+        }
+
+        fn balance(asset: Self::AssetId, who: &T::AccountId) -> Self::Balance {
+            route_call!(asset, total_balance, balance, who).unwrap_or(Zero::zero())
+            //only_asset!(asset, Zero::zero(), Inspect, balance, who)
```
maltekliemann, line 663:
```suggestion
```

maltekliemann (approved):

Issue #1086, opened 2023-08-28
Make prediction-markets easy to configure (or split it)


maltekliemann:
Bit of a meta issue. When I want to test a pallet that depends on some small part of prediction-markets, I have two options:

- Implement a fake version of the parts of prediction-markets that I need.
- Add prediction-markets, court, authorized, global-disputes, simple-disputes, swaps and liquidity-mining to the mock. In the case of testing the new swaps pallet, I don't need any of these.

I'm not a huge fan of the first option as I like having integration tests for our pallets, but the second option really sucks.

sea212:
Ideally we'd have those tests separated, as in prediction-markets unit tests (within the crate) mock the functionality of dependencies and can work solely by itself, thus effectively testing the unit itself. Integration tests on the other hand should reside in the `runtime` crate, where each specific dependency is known an accessible.

Chralt98:
> Ideally we'd have those tests separated, as in prediction-markets unit tests (within the crate) mock the functionality of dependencies and can work solely by itself, thus effectively testing the unit itself. Integration tests on the other hand should reside in the runtime crate, where each specific dependency is known an accessible.

This explains the basic ideal (as you said) for the difference of unit tests and integration tests. However, it does not solve the fundamental problem Malte mentioned. I fully understand Malte's concern and I agree, it's a mess to replicate all the dependencies of another dependency. This is why I am personally in favour of testing not against a mock, but against the actual runtime, or if we want to have a mock, then test against it, but with all dependencies that we have (a full runtime representation). This is kind of the idea of integration tests, but it also allows unit testing and it is closer to the actual production runtime behaviour. So this would get rid of all the mocks that we have and replace them with either the real runtime or a general mock close to runtime. 

This comes with the implication that we would need to figure out how to run benchmark test, since those rely on a mock inside each pallet.

Chralt98:
> Implement a fake version of the parts of prediction-markets that I need.

The fake part is always flawed, since it doesn't replicate the real functionality.

maltekliemann:
Solution here is to have both:

- Let's have proper unit tests which use well configured mocks. I'll try to demonstrate how to do this in substrate when testing the liquidity tree.
- Let's put integration tests (which already make up most of our test suite) into the runtime code.

maltekliemann:
[...]

PR #1191, opened 2023-11-13
Fix failing court benchmark

Chralt98:
### What does it do?

Fixes main branch CI https://github.com/zeitgeistpm/zeitgeist/actions/runs/6848052314/job/18617476375

### DIFF
--- a/zrml/court/src/mock.rs
+++ b/zrml/court/src/mock.rs
@@ -127,8 +127,11 @@ impl DisputeResolutionApi for MockResolution {
 
     fn remove_auto_resolve(market_id: &Self::MarketId, resolve_at: Self::BlockNumber) -> u32 {
         <mock_storage::MarketIdsPerDisputeBlock<Runtime>>::mutate(resolve_at, |ids| -> u32 {
-            ids.retain(|id| id != market_id);
-            ids.len() as u32
+            let ids_len = ids.len() as u32;
+            if let Some(pos) = ids.iter().position(|i| i == market_id) {
+                ids.swap_remove(pos);
+            }
+            ids_len
         })
     }
 }
--- a/zrml/court/src/benchmarks.rs
+++ b/zrml/court/src/benchmarks.rs
@@ -533,6 +533,7 @@ benchmarks! {
 
         let caller: T::AccountId = whitelisted_caller();
         deposit::<T>(&caller);
+        // adds resolve_at for the created market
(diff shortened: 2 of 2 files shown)


maltekliemann:
What exactly happened here? Why wasn't this caught by the CI?

Chralt98:
> What exactly happened here? Why wasn't this caught by the CI?

It was caught by the CI, but because we usually expect the `No space left on device` error, it accidentally wasn't caught.

After integrating verify checks for court, one verify check was only visible for the actual execution of the benchmarking logic and couldn't be caught by the benchmark test suite. I don't know exactly why this happened, maybe because we use a mock and therefore also mock our auto resolve dispute api :D.

maltekliemann:
Yeah, the `DisputeResolutionApi` mock incorrectly mirrors the behavior of the implementation in prediction-markets. The former removes _all_ occurences of the market ID while the latter only appears to remove the first item. Could you change the mock real quick so this sort of problem doesn't occur again?

Chralt98:
2d2c39dae1cd43efbdbcab78ca08ea6987bf6f9f

Chralt98:
I think this example emphasises the strong need of a general mock with all the dependencies as in the actual runtime. By putting in the whole prediction markets pallet configuration into the court mock, we could have prevented my "current mirroring change".

Chralt98:
https://github.com/zeitgeistpm/zeitgeist/issues/1086 We could have prevented this mistake by using either a full mock of all pallets or the runtime itself as a test runtime.

zrml/court/src/mock.rs
```diff
@@ -127,8 +127,11 @@ impl DisputeResolutionApi for MockResolution {
 
     fn remove_auto_resolve(market_id: &Self::MarketId, resolve_at: Self::BlockNumber) -> u32 {
         <mock_storage::MarketIdsPerDisputeBlock<Runtime>>::mutate(resolve_at, |ids| -> u32 {
-            ids.retain(|id| id != market_id);
-            ids.len() as u32
+            let ids_len = ids.len() as u32;
+            if let Some(pos) = ids.iter().position(|i| i == market_id) {
+                ids.swap_remove(pos);
```
maltekliemann, line 132:
```suggestion
                let _ = ids.swap_remove(pos);
```

maltekliemann (approved):

PR #1192, opened 2023-11-17
Integrate Moonwall

Chralt98:
### What does it do?

https://moonsong-labs.github.io/moonwall/

Adds moonwall to our integration test suite.

I looked at the way how tanssi integrated moonwall https://github.com/moondance-labs/tanssi/tree/master/test

Add integration tests.

Just using zombienet, but test suite is problematic here, since custom ts scripts are hard to write. Look [here](https://github.com/zeitgeistpm/zeitgeist/blob/chralt98-add-integration-tests/integration-tests/zombienet/0001-balance-transfer.ts) for an zndsl custom ts script example called by [this](https://github.com/zeitgeistpm/zeitgeist/blob/a2d0a57312834afdba6ebf3fcf24469e869c2849/integration-tests/zombienet/produce-blocks.zndsl#L35). 

### References

https://github.com/Moonsong-Labs/moonwall


maltekliemann (changes requested):
Seems to me that this contains a bunch of copyrighted material without proper attribution. This should be rectified. Since we're dealing with JSON files which can't handle comments, adding a separate LICENSE file to the integration-tests/ folder seems fine to me. However, depending on the license, changes to files may have to be tracked, as well.

maltekliemann (approved):

integration-tests/scripts/download-polkadot.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Check if the operating system is macOS
+if [[ $(uname) == "Darwin" ]]; then
+    echo "This script is not intended for MacOS. But keep in mind you need to have 'polkadot', 'polkadot-execute-worker', 'polkadot-prepare-worker' in any case! So compile those yourself! Exiting..."
+    exit 1
+fi
+
+# Grab Polkadot version
+branch=$(egrep -o '/polkadot.*#([^\"]*)' $(dirname $0)/../../Cargo.lock | head -1 | sed 's/.*release-//#')
+polkadot_release=$(echo $branch | sed 's/#.*//' | sed 's/\/polkadot-sdk?branch=tanssi-polkadot-v//')
```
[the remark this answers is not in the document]
Chralt98: It is done here https://github.com/zeitgeistpm/zeitgeist/pull/1194/files#diff-754c1ca397e1f6bffb5e02524e5ef26316533fdc6ff0791151be2193095142beL8-R14

However without the TODO

integration-tests/specs/zeitgeist-parachain-2101.json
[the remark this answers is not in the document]
Chralt98: Changes made in the [other](https://github.com/zeitgeistpm/zeitgeist/pull/1194) PR [ecb38d10e3587e2ce24572a2ac50fd779edcf0a0](https://github.com/zeitgeistpm/zeitgeist/pull/1194/commits/ecb38d10e3587e2ce24572a2ac50fd779edcf0a0)

integration-tests/scripts/download-polkadot.sh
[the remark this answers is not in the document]
Chralt98: 4e41a90a28962458d596e06e8287ca06d977d2ff

integration-tests/scripts/download-polkadot.sh
```diff
@@ -0,0 +1,65 @@
...
+# polkadot_release=$(echo $branch | sed 's/#.*//' | sed 's/\/polkadot-sdk?branch=polkadot-v//' | sed 's/-.*//')
+polkadot_release="1.1.0"
+
+# Always run the commands from the "integration-tests" dir
+cd $(dirname $0)/..
+
+if [[ -f tmp/polkadot ]]; then
+	POLKADOT_VERSION=$(tmp/polkadot --version)
+	if [[ $POLKADOT_VERSION == *$polkadot_release* ]]; then
+		exit 0
```
sea212, line 41: Feels strange that the script just silently finishes whenever the binaries exist in the correct version.

```suggestion
		echo "Polkadot binary has correct version"
		exit 0
```

integration-tests/scripts/download-polkadot.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Check if the operating system is macOS
+if [[ $(uname) == "Darwin" ]]; then
+    echo "This script is not intended for MacOS, since the prebuild binaries are meant to be executed on Linux. But keep in mind you need to have 'polkadot', 'polkadot-execute-worker', 'polkadot-prepare-worker' in any case! So compile those yourself! Exiting..."
+    exit 1
+fi
+
+# TODO: Use following line instead of the one above once Zeitgeist uses >=polkadot-v1.1.0
+# Note: Version 1.4.0 of relaychain didn't allow the parachain to produce blocks
+# polkadot_release=$(echo $branch | sed 's/#.*//' | sed 's/\/polkadot-sdk?branch=polkadot-v//' | sed 's/-.*//')
```
sea212, line 32:
```suggestion
# branch=$(egrep -o '/polkadot.*#([^\"]*)' $(dirname $0)/../../Cargo.lock | head -1 | sed 's/.*release-//#')
# polkadot_release=$(echo $branch | sed 's/#.*//' | sed 's/\/polkadot-sdk?branch=polkadot-v//' | sed 's/-.*//')
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,269 @@
+name: Integration Tests
+
+on:
+  push:
+    branches: [ main, chralt98-add-it-tests ]
```
sea212, line 5: This is a pretty heavy workload. I propose we save resources and execute this only when a release candidate or release tag is set. In any case, `chralt98-add-it-tests` should be removed (for testing `dispatch` is a useful target). I can see at https://github.com/zeitgeistpm/zeitgeist/actions/runs/7448080700 that the CI works.

```suggestion
on:
  push:
    tags:
      - '^v[0-9]+.[0-9]+.[0-9]+(-rc[0-9]+)?$'
```

integration-tests/moonwall.config.json
```diff
@@ -0,0 +1,129 @@
...
+   "environments": [
+      {
+         "name": "zombienet_zeitgeist_upgrade",
+         "testFileDir": [
+            "tests/rt-upgrade-zombienet"
+         ],
+         "runScripts": [
+            "build-zeitgeist-spec.sh",
+            "download-polkadot.sh"
+         ],
```
sea212, line 14: I think a script that builds zeitgeist in release mode should be contained within that list.

sea212 (commented):
First of all, awesome much needed features. Happy to see more robustness against some of the most critical errors. Moonwall was a good choice.

When I execute Zombienet as per documentation:
```
./scripts/download-polkadot.sh
cd ../
./integration-tests/scripts/deploy-zombienet.sh
```

I get the error: 
> Error:  	 Error: Command failed with exit code 127: bash -c ./target/release/zeitgeist build-spec  --disable-default-bootnode > /tmp/zombie-efcb0d184977ebbca8a8740e4b279d5a_-43492-yqbrUc3SnlaC/cfg/undefined-plain.json
bash: line 1: ./target/release/zeitgeist: No such file or directory

I suggest to create another script to build the zeitgeist node for a specific `CHAIN` (default zeitgeist), `PROFILE` (default release), and `FEATURES`. It should be invoked properly whenever Zeitgeist binaries are needed, such as in the case above.

`deploy-zombienet.sh` requires `jq`, but it seems on my default Ubuntu 22.04 LTS it was not present. We should ensure it's present.

integration-tests/scripts/deploy-zombienet.sh
```diff
@@ -67,6 +67,7 @@ function download_zombienet {
     fi
 
     mkdir -p ./tmp/
+    mkdir -p ./integration-tests/tmp/
```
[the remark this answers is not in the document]
Chralt98: cbc663bb69a598d895f76fb02c8fbc720201e472

integration-tests/moonwall.config.json
Chralt98: The reason why I didn't do this is because the CI should only build the node once (for all tests), so that we can save CI runner resources and ultimately fasten the CI process.

sea212: I think the we should fix that, Moonwall errors if the node is not build. You can check in the build script whether the script is in an CI environment by verifying the state of the [`GITHUB_ACTIONS`](https://docs.github.com/en/actions/learn-github-actions/variables) environment variable (`=true` if in CI).

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1192/commits/9e9a268d80005e27f136aeddb2335c81787eaaac

Chralt98: wait it's not ready. the script needs to be executed from the root folder..

Chralt98: d6585ded8d53f27d68d8cb6de254ba23d1e49664

Chralt98: I will watch the CI running and if everything works, I will push another commit to remove the CI trigger for this branch and then I will request your review again.

Chralt98: Unfortunately I get a `No space left on device` CI :-( So I rely on your advanced review.

sea212: Good to know. I'll add the job to the self-hosted section once the new servers are ready.

sea212 (approved):

PR #1193, opened 2023-11-23
Ensure MinBetSize after fee

Chralt98:
### What does it do?

Ensure the amount minus fees is above the MinBetSize for parimutuels.

Adds `require_transactional`, because this PR breaks the rule `verify first, write last`.

An error because of the existential deposit should not happen as long as the `MinBetSize` minus the market creator fees is above the existential deposits.

Malte noticed that amount_minus_fees should be checked for MinBetSize. The concern was that `amount_minus_fees` could drop below the existential deposit, if `MinBetSize` was not high enough after the deduction of external fees. The implementation of `transfer` and `deposit` itself would throw an internal error, if the situation ever occurs, that this case happens. But in order for us to clarify the error message if it happens, we now throw our own error message to clarify the mitigation actions for the front end and polkadot js users.

After https://github.com/zeitgeistpm/zeitgeist/pull/1183 is merged, we should use consistent `verify first, write last` and mitigate the TODO in this PR.

### DIFF
--- a/zrml/parimutuel/src/lib.rs
+++ b/zrml/parimutuel/src/lib.rs
@@ -33,6 +33,7 @@ mod pallet {
     use frame_support::{
         ensure, log,
         pallet_prelude::{Decode, DispatchError, Encode, TypeInfo},
+        require_transactional,
         traits::{Get, IsType, StorageVersion},
         PalletId, RuntimeDebug,
     };
@@ -173,10 +174,6 @@ mod pallet {
         /// Action cannot be completed because an unexpected error has occurred. This should be
         /// reported to protocol maintainers.
         InconsistentState(InconsistentStateError),
-        /// The amount minus fees is below the existential deposit of the base asset.
-        AmountMinusFeesBelowBaseAssetExistentialDeposit,
-        /// The amount minus fees is below the existential deposit of the parimutuel share asset.
-        AmountMinusFeesBelowParimutuelShareExistentialDeposit,
     }
 
     // NOTE: these errors should never happen.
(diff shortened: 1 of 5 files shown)


zrml/parimutuel/src/lib.rs
```diff
@@ -336,6 +340,19 @@ mod pallet {
             let external_fees = T::ExternalFees::distribute(market_id, base_asset, &who, amount);
             let amount_minus_fees =
                 amount.checked_sub(&external_fees).ok_or(Error::<T>::Unexpected)?;
+
+            let existential_deposit_asset = T::AssetManager::minimum_balance(asset);
+            let existential_deposit_base_asset = T::AssetManager::minimum_balance(base_asset);
+            // TODO: distribute already charges the fees / modifies storage (verify first, write last rule broken)
+            // TODO: mitigate this by using `get_fee` and then `distribute` after the merge with PR https://github.com/zeitgeistpm/zeitgeist/pull/1183
```
maltekliemann, line 347: We don't have to observe verify first/write last where it's not convenient. If you want to be used, it's probably better to add `require_transactional` to the `do_*` functions. Using `get_fee` has it's own issues.

```suggestion
```

zrml/parimutuel/src/lib.rs
```diff
@@ -173,6 +173,10 @@ mod pallet {
         /// Action cannot be completed because an unexpected error has occurred. This should be
         /// reported to protocol maintainers.
         InconsistentState(InconsistentStateError),
+        /// The amount minus fees is below the existential deposit of the base asset.
+        AmountMinusFeesBelowBaseAssetExistentialDeposit,
+        /// The amount minus fees is below the existential deposit of the parimutuel share asset.
+        AmountMinusFeesBelowParimutuelShareExistentialDeposit,
```
maltekliemann, line 179: Seems like we don't have tests that ensure that these trigger.

maltekliemann (changes requested):

maltekliemann (approved):

PR #1205, opened 2023-12-08
Add env_logger and add force-debug feature

Chralt98:
### What does it do?

Adds the env_logger to the mocks so that logs like `log::info`, `log::error`, `log::warn` can be seen in tests. 

It also adds the `force-debug` feature to easily identify the causes of problems in benchmarks or in release builds, because all logs are expanded and not `wasm::stripped` in this case.

### DIFF
--- a/zrml/rikiddo/src/mock.rs
+++ b/zrml/rikiddo/src/mock.rs
@@ -148,6 +148,9 @@ impl ExtBuilder {
     pub fn build(self) -> sp_io::TestExternalities {
         let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
 
+        // see the logs in tests when using `cargo test -- --nocapture`
+        let _ = env_logger::builder().is_test(true).try_init();
+
         pallet_balances::GenesisConfig::<Runtime> { balances: self.balances }
             .assimilate_storage(&mut t)
             .unwrap();
--- a/zrml/styx/src/mock.rs
+++ b/zrml/styx/src/mock.rs
@@ -111,6 +111,9 @@ impl ExtBuilder {
     pub fn build(self) -> sp_io::TestExternalities {
         let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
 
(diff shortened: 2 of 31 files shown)


zrml/styx/src/mock.rs
```diff
@@ -1,3 +1,4 @@
+// Copyright 2022-2023 Forecasting Technologies LTD.
```
maltekliemann, line 1:
```suggestion
// Copyright 2023 Forecasting Technologies LTD.
```

zrml/rikiddo/src/mock.rs
```diff
@@ -1,3 +1,4 @@
+// Copyright 2022-2023 Forecasting Technologies LTD.
```
maltekliemann, line 1:
```suggestion
// Copyright 2023 Forecasting Technologies LTD.
```

maltekliemann (changes requested):

zrml/rikiddo/src/mock.rs
```diff
@@ -148,6 +149,9 @@ impl ExtBuilder {
     pub fn build(self) -> sp_io::TestExternalities {
         let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
 
+        // see the logs in tests when using `cargo test -- --nocapture`
```
maltekliemann, line 152: We should mention that `RUST_LOG=debug cargo test` is the required call to see infos/warnings.

Chralt98:
@maltekliemann and I found out that `RUST_LOG=debug` can be used as the compiler flag to see warnings and infos.

zrml/rikiddo/src/mock.rs
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1205/commits/66e69565728c008794799cb2b6e5d6ea00ec6dff

maltekliemann (approved):
Strictly speaking, `-- --nocapture` isn't required here, but I'll let that slide. ;)

PR #1233, opened 2024-01-16
Reduce benchmark runs of Zeitgeist pallets

sea212:
Running benchmarks of Zeitgeist pallets on the Zeitgeist reference machine currently takes four days.

### What does it do?
- Reduces the benchmark iterations for Zeitgeist pallets (currently benchmarking those pallets takes 4 days on the reference machine)

### DIFF
--- a/scripts/benchmarks/configuration.sh
+++ b/scripts/benchmarks/configuration.sh
@@ -30,8 +30,8 @@ export ZEITGEIST_PALLETS=(
     zrml_authorized zrml_court zrml_global_disputes zrml_liquidity_mining zrml_neo_swaps \
     zrml_orderbook zrml_parimutuel zrml_prediction_markets zrml_swaps zrml_styx \
 )
-export ZEITGEIST_PALLETS_RUNS="${ZEITGEIST_PALLETS_RUNS:-1000}"
-export ZEITGEIST_PALLETS_STEPS="${ZEITGEIST_PALLETS_STEPS:-10}"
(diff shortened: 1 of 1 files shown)


scripts/benchmarks/configuration.sh
```diff
@@ -30,8 +30,8 @@ export ZEITGEIST_PALLETS=(
     zrml_authorized zrml_court zrml_global_disputes zrml_liquidity_mining zrml_neo_swaps \
     zrml_orderbook zrml_parimutuel zrml_prediction_markets zrml_swaps zrml_styx \
 )
-export ZEITGEIST_PALLETS_RUNS="${ZEITGEIST_PALLETS_RUNS:-1000}"
-export ZEITGEIST_PALLETS_STEPS="${ZEITGEIST_PALLETS_STEPS:-10}"
+export ZEITGEIST_PALLETS_RUNS="${ZEITGEIST_PALLETS_RUNS:-20}"
+export ZEITGEIST_PALLETS_STEPS="${ZEITGEIST_PALLETS_STEPS:-50}"
```
Chralt98, line 34: I always see the default to be runs: 50 and steps: 20. Seems like you could have meant that.

sea212: Please elaborate. All other pallets within our repo use `RUNS=20` and `STEPs=50` and Substrate pallets in general also use this configuration, as can be seen at https://github.com/paritytech/polkadot-sdk/blob/a42a47f81d91ccf2f885ff01749185cd4c793278/substrate/frame/balances/src/weights.rs#L30-L31

Chralt98 (approved):

PR #1235, opened 2024-01-18
Implement `force_pool_exit` and disable other zrml-swaps functions

maltekliemann:
### What does it do?

Implements `force_pool_exit`, which allows any user to forcibly withdraw another user's funds held in a Balancer-style pool. This will be used to defund all Balancer-style pools after the migration to LMSR that will be executed in 0.5.0.

- I've installled call filters for all functions of zrml-swaps except `force_pool_exit`. Seemed appropriate. It's not strictly necessary because all pools will be closed and no pool can be created, but it just seemed reasonable to me.
- I've not benchmarked `force_pool_exit` because it's virtually equivalent to `pool_exit` in terms of computation and just used the `pool_exit` benchmark instead.

### DIFF
--- a/runtime/battery-station/src/lib.rs
+++ b/runtime/battery-station/src/lib.rs
@@ -58,9 +58,9 @@ use zrml_prediction_markets::Call::{
 };
 use zrml_rikiddo::types::{EmaMarketVolume, FeeSigmoid, RikiddoSigmoidMV};
 use zrml_swaps::Call::{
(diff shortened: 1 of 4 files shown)


sea212 (approved):
LGTM. Since function call serves chain maintenance, it would be appropriate to use `Pays::No` for fees whenever the result is that account exit. I don't think this is necessary though, as Zeitgeist will execute this for all accounts.

PR #1240, opened 2024-01-21
Reorganize prediction-markets tests

maltekliemann:
### What does it do?

Restructures prediction-market tests into separate files, eliminates duplicates, makes `get()` calls through `Config` instead of directly calling mock constants and adds TODOs pointing to missing tests, which will be filled in in future PRs.


maltekliemann:
Closing because I'm pushing in the wrong branch.

PR #1242, opened 2024-01-21
Reorganize prediction-market tests

maltekliemann:
### What does it do?

Restructures prediction-market tests into separate files, eliminates duplicates, makes `get()` calls through `Config` instead of directly calling mock constants and adds TODOs pointing to missing tests, which will be filled in in future PRs.


Chralt98 (approved):
Finally the output of CodeCov is quite useful again:

> Comparison is base [(cb6a053)](https://app.codecov.io/gh/zeitgeistpm/zeitgeist/commit/cb6a0531d8be6844472b69cbfe5093ec4c899143?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zeitgeistpm) 92.63% compared to head [(a5f24ef)](https://app.codecov.io/gh/zeitgeistpm/zeitgeist/pull/1242?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zeitgeistpm) 92.71%.
Report is 4 commits behind head on main.

Lines added and removed 😏 ^^ It's all about perception 😂

> makes get() calls through Config instead of directly calling mock constants

What is that?

maltekliemann:
> > makes get() calls through Config instead of directly calling mock constants
> 
> What is that?

Replace `MinCategories::get()` with `<Runtime as Config>::MinCategories::get()`, etc. A bit more typing/a few more line breaks, but the advantage is clarity _and_ tests are independent of whether they're run against the mock or the actual runtime. The latter will be useful when we move the integration tests to `runtime/`.

sea212:
@mergifyio requeue

PR #1246, opened 2024-01-23
Improve fee payment management

sea212:
### What does it do?
- https://www.notion.so/zeitgeistpm/Fee-improvement-77f0479ba0b3430397a2cb5fbefd78e4?pvs=4


zrml/prediction-markets/src/lib.rs
```diff
@@ -832,14 +832,15 @@ mod pallet {
...
-                return Ok(Some(T::WeightInfo::redeem_shares_scalar()).into());
-            }
+                Some(T::WeightInfo::redeem_shares_scalar())
+            } else {
+                None
+            };
 
-            let default_weight: Option<Weight> = None;
-            Ok((default_weight, Pays::No).into())
+            Ok((weight, Pays::No).into())
```
maltekliemann, line 843: `OutcomeReport` only has two variants, which are covered by the two `let if` cases. So we might as well use a `match` statement, which would save us the unreachable `else` branch and avoid early returns. This also makes the final return of the function unnecessary. I also find it odd that this final return uses `Pays::No` (not that it matters, but it is odd).

```suggestion
            let weight = match resolved_outcome {
                OutcomeReport::Categorical(_) => T::WeightInfo::redeem_shares_categorical(),
                OutcomeReport::Scalar(_) => T::WeightInfo::redeem_shares_scalar(),
            };

            Ok(Some(weight).into())
```

maltekliemann (changes requested):

zrml/prediction-markets/src/lib.rs
sea212: Agreed on the match.
I also agree `Pays::No` is debatable here. I am quite confused now, as [previously](https://github.com/zeitgeistpm/zeitgeist/blob/761cc598a6eb168a3377831d0addc69f36069f65/zrml/prediction-markets/src/lib.rs#L843-L844) fee payment was only remitted in the unreachable `else` branch. Strange ... Besides that, I think one can argue that we had to outsource token redemption to a dispatchable call because it was not technically feasible to automatically redeem the shares on resolution. Consequently, the user does this maintenance task and thus should be reimbursed for it. I thought that was the idea behind the initial `Pays::No`. Plus, dust rewards that are less in value than the TX fee for calling this dispatchable might be ignored and congest the chain. Personally I am fine with both solutions. Any preference?

maltekliemann: I'm generally weary of using `Pays::No` for obvious reasons, but I understand the reasoning here. Either way's fine by me.

sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1246/commits/78444eeeff32c43e22b5a0870d85abd8c4f924ef

maltekliemann (approved):

PR #1247, opened 2024-01-24
Fix Rust and Discord badge

sea212:
### What does it do?
Fixes the URLs associated with the Rust and Discord badge on the main page of this repository.


Chralt98 (approved):
Reminds me of removing the test instead of fixing the underlying problem XD

PR #1251, opened 2024-02-07
Use `Balance` instead of `u128` as type for Balancer weights

### LINKED ISSUE #1213: [Swaps] Use a specific type for CPMM pool weights
Replace the hard-coded `u128` with a user-configurable type, either `Balance` or `Weight: From<Balacne> + Into<Balance>`. Probably the former.

### LINKED ISSUE #1216: [Primitives] Add `Weight` type to `SwapsApi`
Same as #1213, just for the `SwapsApi` trait.

maltekliemann:
### What does it do?

We replace the hardcoded `u128` for Balancer weights with the chain's balance type. We also replace a (benign) panicker in the CPMM math. Fixes #1213. Fixes #1216.

### DIFF
--- a/zrml/swaps/src/fixed.rs
+++ b/zrml/swaps/src/fixed.rs
@@ -26,7 +26,9 @@ use frame_support::dispatch::DispatchError;
 use zeitgeist_primitives::{
     constants::BASE,
     math::{
-        checked_ops_res::{CheckedAddRes, CheckedDivRes, CheckedMulRes, CheckedSubRes},
+        checked_ops_res::{
+            CheckedAddRes, CheckedDivRes, CheckedMulRes, CheckedRemRes, CheckedSubRes,
+        },
         fixed::{FixedDiv, FixedMul},
     },
 };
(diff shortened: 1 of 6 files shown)


zrml/swaps/src/fixed.rs
```diff
@@ -53,7 +55,7 @@ pub fn bsub_sign(a: u128, b: u128) -> Result<(u128, bool), DispatchError> {
 }
 
 pub fn bpowi(a: u128, n: u128) -> Result<u128, DispatchError> {
-    let mut z = if n % 2 != 0 { a } else { BASE };
```
Chralt98, line 56: Oh, severe that we used `%` before 😰. But good reminder that `number % 0` produces a panic.

Chralt98: I wonder, if we have other places using the `%` operator.. In court we have one for sure, but no zero division is possible there..

Chralt98 (approved):

zrml/swaps/src/fixed.rs
Chralt98: Could you update the court pallet regarding this change too in this PR?

maltekliemann: #1254; reverted changes from this PR

sea212 (approved):

Chralt98 (approved):

PR #1253, opened 2024-02-07
Migrate neo-swaps pools to bounded storage

### LINKED ISSUE #1214: [NeoSwaps] In `Pool`, replace `BTreeMap` with `BoundedBTreeMap` to avoid manual implementations of `MaxEncodedLen`; then make `MaxAssets` a config parameter
In particular, we can then make `MaxAssets` a config parameter.

maltekliemann:
### What does it do?

Migrates neo-swaps pools from `BTreeMap` to `BoundedBTreeMap` to avoid potentially faulty manual `MaxEncodedLen` implementations. Fixes #1214.


Chralt98 (dismissed):

zrml/neo-swaps/src/consts.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) const EXP_NUMERICAL_LIMIT: u128 = 10;
 /// Numerical lower limit for ln arguments (fixed point number).
 pub(crate) const LN_NUMERICAL_LIMIT: u128 = BASE / 10;
 /// The maximum number of assets allowed in a pool.
-pub(crate) const MAX_ASSETS: u16 = 128;
+pub(crate) const MAX_ASSETS: u32 = 128;
```
sea212, line 25: Why is that necessary? `u32` seems overkill and also deviates from `MaxCategories` type in PM.

zrml/neo-swaps/src/lib.rs
```diff
@@ -834,7 +834,7 @@ mod pallet {
             ensure!(market.scoring_rule == ScoringRule::Lmsr, Error::<T>::InvalidTradingMechanism);
             let asset_count = spot_prices.len();
             ensure!(asset_count as u16 == market.outcomes(), Error::<T>::IncorrectVecLen);
-            ensure!(market.outcomes() <= MAX_ASSETS, Error::<T>::AssetCountAboveMax);
+            ensure!(market.outcomes() as u32 <= MaxAssets::get(), Error::<T>::AssetCountAboveMax);
```
sea212, line 837: Casting might result in unexpected results, for example `300u16 as u8` results in `44`.

```suggestion
            ensure!(market.outcomes().saturated_into() <= MaxAssets::get(), Error::<T>::AssetCountAboveMax);
```

zrml/neo-swaps/src/types/max_assets.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::consts::MAX_ASSETS;
+use sp_runtime::traits::Get;
+
+pub(crate) struct MaxAssets;
+
+impl Get<u32> for MaxAssets {
+    fn get() -> u32 {
+        MAX_ASSETS
+    }
+}
```
sea212, line 27: Why is that necessary? Why don't we use a `Config` type for that?

zrml/neo-swaps/src/types/max_assets.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2023-2024 Forecasting Technologies LTD.
```
sea212, line 1:
```suggestion
// Copyright 2024 Forecasting Technologies LTD.
```

zrml/prediction-markets/src/mock.rs
```diff
@@ -44,6 +44,8 @@ use sp_runtime::{
     DispatchError, DispatchResult,
 };
 use std::cell::RefCell;
+#[cfg(feature = "parachain")]
+use zeitgeist_primitives::types::XcmAsset;
```
sea212, line 48: We already have that feature above and could combine all conditional imports.

zrml/neo-swaps/src/migration.rs
```diff
@@ -15,6 +15,239 @@
...
+    T: Config,
+    LSM: LiquiditySharesManager<T>,
+{
+    pub account_id: T::AccountId,
+    pub reserves: BTreeMap<AssetOf<T>, BalanceOf<T>>,
+    pub collateral: AssetOf<T>,
+    pub liquidity_parameter: BalanceOf<T>,
+    pub liquidity_shares_manager: LSM,
+    pub swap_fee: BalanceOf<T>,
+}
```
sea212, line 65: `BTreeMap` to `BoundedBTreeMap` does not require any migration because the underlying scale encoded data remains the same. The new generic parameter that specifies the boundary is only used on the logical layer, e.g. to ensure that nothing is added beyond bounds or decoding fails if boundaries are surpassed, but the data layout remains the same.

https://docs.rs/bounded-collections/0.2.0/src/bounded_collections/bounded_btree_map.rs.html#34

sea212: Other than that, migration looks good.

zrml/neo-swaps/src/consts.rs
maltekliemann: Took me a while to remember. It's because the generic size parameter of the `BoundedBTreeMap` expects a `Get<u32>` implementor. I guess I could write an adapter instead.

maltekliemann: The deviation actually makes sense to me. In combo, we're not combining markets, we're just combining their pools. So combo pools will, in general, have more assets than markets.

zrml/neo-swaps/src/lib.rs
maltekliemann: I get that, but we're casting to `u32` here, so that should always be fine. That being said, is there any argument against just using saturated or checked conversions instead? Seems like these casts are a risk in case we ever change the type of `market.outcomes()`. Also, clippy doesn't appear to warn on narrow casts.

zrml/prediction-markets/src/mock.rs
maltekliemann: You mean with `cfg_if`?

zrml/neo-swaps/src/migration.rs
maltekliemann: Hmm, know I feel silly. I think in that case it makes more sense to just remove the entire migration.

zrml/neo-swaps/src/types/max_assets.rs
maltekliemann: I was worried about a misconfiguration. The thing is that there's a (hard-coded!) minimum spot price required when creating a pool, so there should be a maximum number of assets allowed, because if there are too many assets, it becomes impossible to keep all spot prices above the minimum spot price. I'll add a comment to the code.

zrml/neo-swaps/src/lib.rs
sea212: The difference is that the cast's proper functionality relies on an assumption, whereas the saturated conversion does not, as it just maxes out when an overflow occurs.

zrml/prediction-markets/src/mock.rs
sea212: Like https://github.com/zeitgeistpm/zeitgeist/blob/2d0277bea7ba5796733b1db4db484dc73f30e23c/zrml/prediction-markets/src/lib.rs#L58-L62

zrml/neo-swaps/src/migration.rs
sea212: I'm fine with either way. But it seems you have some additional useful sanity checks that already seem to make it worth executing it https://github.com/zeitgeistpm/zeitgeist/blob/8415c118c8fe564683f4868fb04f77def3d84f35/zrml/neo-swaps/src/migration.rs#L90-L93

zrml/neo-swaps/src/consts.rs
sea212: Potentially more than 65535 assets? Sounds like a lot of effort, curious to see the clever shortcuts to do the math that you'll find :)

zrml/neo-swaps/src/lib.rs
sea212: Really doesn't matter in this case, I agree. But I tend to opt for the side of less conditions to remember.

zrml/neo-swaps/src/consts.rs
maltekliemann: :) Anything over 1,000 is pushing it. When I said the deviation makes sense I meant it makes sense to allow a different value rather than a different type. The `u32` is really only there to satisfy the trait bounds of `Bounded*`.

zrml/prediction-markets/src/mock.rs
maltekliemann: 9c7d94edeca54627f07b3a77f52e8ca82752ac25

zrml/neo-swaps/src/lib.rs
maltekliemann: Went a little crazy and cleaned up all the conversions. The `NarrowingConversion` error is to catch vectors that are too large and would cause saturation. I understand this cannot happen for various other technical reasons, but somehow I felt this was necessary in terms of correctness.

maltekliemann: 5d75af2b1f81d6cf0581d45b6fe00cc6df1f9ea1

zrml/neo-swaps/src/types/max_assets.rs
sea212: Seems a bit overkill to me to create a whole new file for that. I was thinking along the lines of putting the following line next to the constant:
```rust
pub type MaxAssets = frame_support::pallet_prelude::ConstU32<MAX_ASSETS>;
```

sea212 (dismissed):

zrml/neo-swaps/src/lib.rs
```diff
@@ -832,9 +846,11 @@ mod pallet {
...
             ensure!(market.status == MarketStatus::Active, Error::<T>::MarketNotActive);
             ensure!(market.scoring_rule == ScoringRule::Lmsr, Error::<T>::InvalidTradingMechanism);
-            let asset_count = spot_prices.len();
-            ensure!(asset_count as u16 == market.outcomes(), Error::<T>::IncorrectVecLen);
-            ensure!(market.outcomes() <= MAX_ASSETS, Error::<T>::AssetCountAboveMax);
+            let asset_count_u16: u16 =
+                spot_prices.len().try_into().map_err(|_| Error::<T>::NarrowingConversion)?;
+            let asset_count_u32: u32 = asset_count_u16.into();
+            ensure!(asset_count_u16 == market.outcomes(), Error::<T>::IncorrectVecLen);
+            ensure!(asset_count_u32 <= MaxAssets::get(), Error::<T>::AssetCountAboveMax);
```
Chralt98, line 853: You can use shadowing here if you like. I mean this:

```rust
            let asset_count: u16 =
                spot_prices.len().try_into().map_err(|_| Error::<T>::NarrowingConversion)?;
            ensure!(asset_count_u16 == market.outcomes(), Error::<T>::IncorrectVecLen);
            let asset_count: u32 = asset_count.into();
            ensure!(asset_count_u32 <= MaxAssets::get(), Error::<T>::AssetCountAboveMax);
```

Let me know if this doesn't work.

Chralt98: Ah, you use it below... mhhh... then it's fine I guess.

Chralt98 (dismissed):

sea212 (approved):

PR #1254, opened 2024-02-09
Remove unchecked modulo operations from production

maltekliemann:

### DIFF
--- a/primitives/src/math/checked_ops_res.rs
+++ b/primitives/src/math/checked_ops_res.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 Forecasting Technologies LTD.
+// Copyright 2023-2024 Forecasting Technologies LTD.
 //
 // This file is part of Zeitgeist.
 //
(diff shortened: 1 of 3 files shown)


Chralt98 (approved):

sea212 (commented):
Seems good to me, however there are still two occurrences of unchecked modulo in production code:
https://github.com/zeitgeistpm/zeitgeist/blob/33503dbb789e274f50e9e5c689da90f890039323/zrml/court/src/lib.rs#L1416
https://github.com/zeitgeistpm/zeitgeist/blob/33503dbb789e274f50e9e5c689da90f890039323/zrml/court/src/lib.rs#L1795-L1798

maltekliemann:
Okay, those are in `debug_assert` statements though.

sea212:
Right, but as far as I can tell the node will panic due to the modulo by zero instead of the assertion. Fine to leave like that from my perspective though, as the node would crash anyways (another topic).

sea212 (approved):

maltekliemann:
The problem is that it's not easy to have really good error handling without changing the return value of the function, We can just remove the `debug_assert` statement?

sea212:
I think this would be a good example for the `unimplemented_not_terminating` macro plus a test.

Issue #735, opened 2022-08-07
Make `MaxMarketPeriod`, `MaxInRatio` and `MaxOutRatio` hardcoded constants


maltekliemann:

maltekliemann:
Partially solved by #1255.

maltekliemann:
Closing because I don't see the point anymore.

PR #1255, opened 2024-02-10
Refactor swaps

maltekliemann:
### What does it do?

- Hardcodes `Max*Ratio` constants to avoid misconfigurations. Partially solves #735.
- Fixes documentation of all modules related to swaps.
- Changes the return value of single-asset operations. There's no need for these to return their consumed weights.
- Fixes `pool_exit` behavior with respect to accidentally dusted pool accounts (IDs 7, 9, 11 and 14 on mainnet) and ED issues (burns funds from the pool if they're smaller than ED instead of erroring).
- Eliminates unused, duplicate or obsolete errors and events.
- Makes all extrinsics `transactional`.
- Applies `do_*` pattern to all extrinsics.
- Removes unnecessary post-execution weight corrections.
- Removes unnecessary `clone()` calls.
- Uses `withdraw` instead of `slash` for safer burning of pool shares and pool destruction.

- Split `tests.rs`.
- Clean up `Swaps` trait.
- Split `Swaps` implementation from `lib.rs`.
- Solve various issues already raised on GitHub.


primitives/src/traits/swaps.rs
```diff
@@ -50,68 +46,66 @@ pub trait Swaps<AccountId> {
...
         asset_amount: Self::Balance,
         min_pool_amount: Self::Balance,
-    ) -> Result<Weight, DispatchError>;
+    ) -> DispatchResult;
 
-    /// Swap - Exact amount in
+    /// Buy the `asset_out`/`asset_in` pair from the pool for an exact amount of `asset_in`.
     ///
-    /// Swaps a given `asset_amount_in` of the `asset_in/asset_out` pair to `pool_id`.
+    /// This function will error is both `min_asset_amount_out` and `max_price` are `None`.
```
sea212, line 96:
```suggestion
    /// This function will error if both `min_asset_amount_out` and `max_price` are `None`.
```

primitives/src/traits/swaps.rs
```diff
@@ -120,23 +114,23 @@ pub trait Swaps<AccountId> {
...
         min_asset_amount_out: Option<Self::Balance>,
         max_price: Option<Self::Balance>,
-    ) -> Result<Weight, DispatchError>;
+    ) -> DispatchResult;
 
-    /// Swap - Exact amount out
+    /// Buy the `asset_out`/`asset_in` pair from the pool, receiving an exact amount of `asset_out`.
     ///
-    /// Swaps a given `asset_amount_out` of the `asset_in/asset_out` pair to `origin`.
+    /// This function will error is both `min_asset_amount_out` and `max_price` are `None`.
```
sea212, line 121:
```suggestion
    /// This function will error if both `min_asset_amount_out` and `max_price` are `None`.
```

zrml/swaps/src/lib.rs
```diff
@@ -836,6 +1034,7 @@ mod pallet {
             Ok(())
         }
 
+        // TODO Fix the issue that clean pools can't be drained.
```
sea212, line 1037: `// TODO(#issue): description`

sea212 (approved):
LGTM

zrml/swaps/src/lib.rs
maltekliemann: bc2e707

sea212 (approved):

PR #1257, opened 2024-02-14
Migration: Add Market ID to Market

maltekliemann:
### What does it do?

Migrates markets from their old version (including no market ID) to the new version (from branch `mkl-market-id`).


zrml/market-commons/src/migrations.rs
```diff
@@ -15,3 +15,323 @@
...
+#[cfg(feature = "try-runtime")]
+use frame_support::migration::storage_key_iter;
+
+#[cfg(any(feature = "try-runtime", feature = "test"))]
+use frame_support::Blake2_128Concat;
+
+#[cfg(any(feature = "try-runtime", test))]
+const MARKET_COMMONS: &[u8] = b"MarketCommons";
+#[cfg(any(feature = "try-runtime", test))]
+const MARKETS: &[u8] = b"Markets";
```
sea212, line 51: Can be grouped in two groups, but irrelevant here as we'll remove that migration in a couple of days again.

sea212 (approved):

PR #1261, opened 2024-02-21
Implement Hybrid AMM-CDA Router

Chralt98:


zrml/orderbook/src/lib.rs
```diff
@@ -463,7 +464,10 @@ mod pallet {
         ) -> DispatchResult {
             let market = T::MarketCommons::market(&market_id)?;
             ensure!(market.status == MarketStatus::Active, Error::<T>::MarketIsNotActive);
-            ensure!(market.scoring_rule == ScoringRule::Orderbook, Error::<T>::InvalidScoringRule);
```
Robiquet, line 466: Wouldn't `ScoringRule::Orderbook` still be valid?

Robiquet: If every market is going to be initialised with `ScoringRule::AmmCdaHybrid` what will happen to parimutuel markets?

maltekliemann: From https://hackmd.io/NBVKUY1QT--YnotR5KUx2w#Scoring-Rules: "The Lmsr and Orderbook variants are removed from the ScoringRule enum and replaced by AmmCdaHybrid, which allows a market to use both the order book and the AMM."

> If every market is going to be initialised with ScoringRule::AmmCdaHybrid what will happen to parimutuel markets?

The `ScoringRule::Parimutuel` object will not be deleted, so the parimutuel markets stay as they are.

Chralt98:
Abandoned in favor of multiple smaller PRs!

PR #1263, opened 2024-02-23
Only run copyright CI when merging into `main`

maltekliemann:
### What does it do?

Only runs copyright CI when merging into `main`. We're making the assumption that all changes that are part of a PR actually "happened" in this PR. But in case of merges of `main` into feature branches, that is not the case.


sea212 (approved):
LGTM

PR #1267, opened 2024-02-26
Update toolchain to fix codecov errors

maltekliemann:
### What does it do?

Updates toolchain to rustc nightly-2023-02-01. Not going any further right now because of the [removal of stdsimd](https://github.com/rust-lang/rust/commit/ea37e8091fe87ae0a7e204c034e7d55061e56790) which some of our dependencies use.

The second commit fixes the usual "unnecessary `clone()`" and string formatting errors that clippy produces after an update.

Please ensure that all tests and benchmarks pass in case the CI fails to verify this.


maltekliemann:
Closed because I don't know what year it is.

PR #1269, opened 2024-02-27
Update toolchain to fix codecov and fuzz errors

maltekliemann:
### What does it do?

Updates toolchain to rustc nightly-2024-01-31. Not going any further right now because of the [removal of stdsimd](https://github.com/rust-lang/rust/commit/ea37e8091fe87ae0a7e204c034e7d55061e56790) which some of our dependencies use.


sea212 (approved):

sea212:
Setting PR to approved although code owner reviews are still outstanding because logic was not changed.

PR #1278, opened 2024-03-04
AMM-CDA-5: Add tests for Hybrid Router

Chralt98:


zrml/hybrid-router/src/mock.rs
```diff
@@ -0,0 +1,502 @@
...
+    fn deploy_pool(
+        _who: Self::AccountId,
+        _market_id: Self::MarketId,
+        _amount: Self::Balance,
+        _swap_prices: Vec<Self::Balance>,
+        _swap_fee: Self::Balance,
+    ) -> DispatchResult {
+        Ok(())
+    }
+}
```
maltekliemann, line 120:
```suggestion
```

zrml/hybrid-router/src/mock.rs
```diff
@@ -0,0 +1,502 @@
...
+    type CloseEarlyDisputeBond = CloseEarlyDisputeBond;
+    type CloseEarlyTimeFramePeriod = CloseEarlyTimeFramePeriod;
+    type CloseEarlyProtectionBlockPeriod = CloseEarlyProtectionBlockPeriod;
+    type CloseEarlyProtectionTimeFramePeriod = CloseEarlyProtectionTimeFramePeriod;
+    type CloseEarlyRequestBond = CloseEarlyRequestBond;
+    type CloseMarketEarlyOrigin = EnsureSignedBy<Sudo, AccountIdTest>;
+    type CloseOrigin = EnsureSignedBy<Sudo, AccountIdTest>;
+    type Court = Court;
+    type Currency = Balances;
+    type DeployPool = DeployPoolNoop;
```
maltekliemann, line 233:
```suggestion
    type DeployPool = NeoSwaps;
```

zrml/hybrid-router/src/tests/trade.rs
```diff
@@ -0,0 +1,159 @@
...
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Zeitgeist. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg(test)]
+
+use super::*;
+
+/// These tests are for shared logic of the buy and sell functions.
```
maltekliemann, line 22: Bit confused here. These tests are all testing functionality of `buy`. The `do_trade` function is an implementation detail and does not require tests. Instead, the public functions like `buy` and `sell` should each have their correct functionality verified separately.

Chralt98: Whereas everything in buy.rs and sell.rs is unique and specific to their respective extrinsics.

maltekliemann: > The reason for having the trade.rs file was that both buy and sell have the same behavior, since do_trade is called by both of them.

Yeah, but you're not testing `do_trade` here, you're testing `buy`. If you would call `do_trade` here, it would be a different story.

> Whereas everything in buy.rs and sell.rs is unique and specific to their respective extrinsics.

The way testing is usually done is that the specification of the function under test defines the behavior: If the input is x, then the FUT does this and returns y. If the input is u, then the FUT does that and return v. Etc. And then the tests verify each of these behaviors.

In this example, the spec might say that on success, both `buy` and `sell` emit an event. This would require two tests: One that verifies that `buy` emits an event, and one that verifies that `sell` emits an event. Only testing this for `buy` and then appealing to the fact that both `buy` and `sell` are implemented using `do_trade` is, I think, asking for trouble: If the next refactor changes the implementation of `sell` to no longer use `do_trade` maybe this will break the behavior of `sell`, but this wouldn't be caught by the tests. Long story short, right now, the execution path `sell` > `do_trade` > ... > emit event doesn't appear to be tested at all.

So I would suggest duplicating all these tests, once for `buy` and once for `sell`.

Chralt98: 0b7c241eb6291669a86960bb293162ec173f3e49

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1278/commits/be344b904fb4b7560db8e9fcff8e2ce5752fb203

maltekliemann (approved):

Chralt98:
Changes added here https://github.com/zeitgeistpm/zeitgeist/commit/2633895b9364d4677155302f00714edb1bff6ea5

This is no normal merge because this would result in changes being reverted.

PR #1280, opened 2024-03-06
Add external fee amount to fill order event

Chralt98:


Chralt98:
I will use another PR about the Hybrid Orderbook AMM event that adds the external fee amount to the event.

Chralt98:
Abandoned in favor of https://github.com/zeitgeistpm/zeitgeist/pull/1288

PR #1283, opened 2024-03-07
New Asset System - Integrate into Orderbook/Court/GlobalDisputes

sea212:
### What does it do?
- Orderbook now uses `reserve` when the `maker_asset` supports it, otherwise an account that's unique per order is used to store the funds of the order.


zrml/orderbook/src/lib.rs
```diff
@@ -401,17 +394,23 @@ mod pallet {
...
+            } else {
+                T::AssetManager::repatriate_reserved_named(
+                    &Self::reserve_id(),
+                    maker_asset,
+                    &maker,
+                    &taker,
+                    taker_fill,
+                    BalanceStatus::Free,
+                )?;
+            }
```
Chralt98, line 411: Assume someone knows the order_account address and sends the maker_asset to it. The pallet would now try to fill the order using the transfer, although the maker has the maker_asset reserved. 

You could use the `reserved_balance` api function to check for the maker's balance instead.

zrml/orderbook/src/lib.rs
```diff
@@ -298,40 +280,45 @@ mod pallet {
...
-                maker,
-                order_data.maker_amount,
-                missing,
-            );
+            if T::AssetManager::ensure_can_withdraw(asset, &order_account, amount).is_ok() {
+                T::AssetManager::transfer(asset, &order_account, maker, amount)?;
+            } else {
+                let missing =
+                    T::AssetManager::unreserve_named(&Self::reserve_id(), asset, maker, amount);
+
```
Chralt98, line 309: Same as already commented.

zrml/orderbook/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1283/commits/e8cbf4111de693d759a51f52a50d27c50299c822

zrml/orderbook/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1283/commits/e8cbf4111de693d759a51f52a50d27c50299c822

zrml/orderbook/src/lib.rs
```diff
@@ -298,40 +280,45 @@ mod pallet {
...
-                order_data.maker_asset,
-                maker,
-                order_data.maker_amount,
-                missing,
-            );
+            let order_account = Self::order_account(order_id);
+            let asset = order_data.maker_asset;
+            let amount = order_data.maker_amount;
+
+            if !T::AssetManager::reserved_balance(asset, &maker).is_zero() {
```
Chralt98, line 304: Why not `reserved_balance_named` here instead? In that way we wouldn't have conflicts with potential other pallets.

Additionally I think the asset could be an outcome asset, which includes the market id, then it's fine, but it could also be the base asset. Other problem we have here then is, that if you have another order on the order book with the same base asset, you would go into this branch, right?

One way to solve this is by putting a more fine-grained reserve identifier onto the reserve, which directly identifies the asset additionally to the pallet identifier.

zrml/orderbook/src/lib.rs
```diff
@@ -401,17 +394,22 @@ mod pallet {
...
+
+            if !T::AssetManager::reserved_balance(maker_asset, &maker).is_zero() {
+                T::AssetManager::repatriate_reserved_named(
+                    &Self::reserve_id(),
+                    maker_asset,
+                    &maker,
+                    &taker,
+                    taker_fill,
+                    BalanceStatus::Free,
+                )?;
```
Chralt98, line 407: Why not `reserved_balance_named` here instead? In that way we wouldn't have conflicts with potential other pallets.

Additionally I think the asset could be an outcome asset, which includes the market id, then it's fine, but it could also be the base asset. Other problem we have here then is, that if you have another order on the order book with the same base asset, you would go into this branch, right?

One way to solve this is by putting a more fine-grained reserve identifier onto the reserve, which directly identifies the asset additionally to the pallet identifier.

zrml/orderbook/src/lib.rs
Chralt98: Additionally you could also check for the amount in question. Like, we have to have at least `amount` available, otherwise it's the other case.

Chralt98: Can't you just check if the asset is created using orml_tokens which has reserves? This way would be the safest approach. Because this is exactly what we want to know here. If the asset was either created by pallet_assets or orml_tokens. Maybe it's worth to break the abstraction of the asset router here in favor of clarity?

Chralt98: But I don't see any case right now how this `reserved_balance_named` could fail right now so that it accidentally uses the reserve case instead of the transfer case. So I suppose it's fine.

zrml/orderbook/src/lib.rs
sea212: Both cases (named and same base asset) don't really matter. The only thing this lines of code are trying to achieve is to check if a asset is reservable in general or not and then execute the correct branch. This builds up on the pre-condition that during order creation, an asset that is reservable is always reserved and an asset that is not reservable is always transferred to the order account.

Assume we use non named reserves. If there is reserved balance, the asset is reservable. Thus `repatriate_reserve_named` will be called which fails if there is no named reserve for that pallet, account and asset.
Assume we create two orders with the same reservable base asset. During order creation the base asset was reserved. When someone wants to fill the order or when the owner wants to remove the order, the order is fetched from the DB and the maker is extracted. If there is reserved balance for that maker, it is unreserved. The reserve of the other order is untouched, assuming logic works properly. I think proper tests for those scenarios are sufficient.

Name reserved balance check: https://github.com/zeitgeistpm/zeitgeist/pull/1283/commits/5185e92e6585f2d332dc42f542de2080ffc7e1d4

zrml/orderbook/src/lib.rs
sea212: https://github.com/zeitgeistpm/zeitgeist/pull/1283/commits/5185e92e6585f2d332dc42f542de2080ffc7e1d4

PR #1284, opened 2024-03-07
New Asset System - Integrate into remaining pallets

sea212:
### What does it do?
- Adjust all remaining pallets to work with the new asset system.


zrml/neo-swaps/src/tests/join.rs
```diff
@@ -208,14 +208,15 @@ fn join_fails_on_insufficient_funds() {
             vec![_1_2, _1_2],
             CENT,
         );
+
```
maltekliemann, line 211:
```suggestion
```

zrml/neo-swaps/src/tests/join.rs
```diff
@@ -208,14 +208,15 @@ fn join_fails_on_insufficient_funds() {
...
         assert_noop!(
             NeoSwaps::join(
                 RuntimeOrigin::signed(ALICE),
                 market_id,
                 _100,
                 vec![u128::MAX, u128::MAX]
             ),
-            orml_tokens::Error::<Runtime>::BalanceTooLow
+            pallet_assets::Error::<Runtime, MarketAssetsInstance>::NoAccount
```
maltekliemann, line 219: Do you know off the top of your head why this raises `NoAccount` but a failed `deploy_pool` raises `BalanceLow`?

zrml/neo-swaps/src/tests/deploy_pool.rs
maltekliemann, line 1: What's the point of the changes to this file?

sea212: It's related to the existence of the assets. We can't deposit tokens for market assets before they are created now, so the change rearrange the code to first create the market before depositing the tokens and to check for the last events before new events are emitted.

zrml/neo-swaps/src/tests/join.rs
sea212: This raises `NoAccount` because `ALICE`'s account for the assets in question was never created, i.e. she has no balance at all. In the `deploy_pool` case, `ALICE`'s account exists because it was created when she received tokens, but the balance she tries to transfer exceeds her available balance.

zrml/neo-swaps/src/tests/join.rs
sea212: nitpicky

maltekliemann (approved):

PR #1285, opened 2024-03-07
Add previous stake information after rejoin

Chralt98:
### What does it do?

It rewards inflation based on the stake information before the current inflation period..

### DIFF
--- a/zrml/court/src/lib.rs
+++ b/zrml/court/src/lib.rs
@@ -29,6 +29,7 @@ use crate::{
 };
 use alloc::{
     collections::{BTreeMap, BTreeSet},
+    format,
     vec::Vec,
 };
 use core::marker::PhantomData;
--- a/zrml/court/src/lib.rs
+++ b/zrml/court/src/lib.rs
@@ -1228,7 +1228,6 @@ mod pallet {
 
             let mut active_lock = BalanceOf::<T>::zero();
             let mut consumed_stake = BalanceOf::<T>::zero();
-            let mut joined_at = now;
 
             if let Some(prev_p_info) = <Participants<T>>::get(who) {
                 ensure!(amount >= prev_p_info.stake, Error::<T>::AmountBelowLastJoin);
@@ -1238,7 +1237,6 @@ mod pallet {
                 {
                     active_lock = prev_p_info.active_lock;
                     consumed_stake = pool_item.consumed_stake;
-                    joined_at = pool_item.joined_at;
 
                     pool.remove(index);
                 } else {
@@ -1251,7 +1249,7 @@ mod pallet {
                 pool = remove_weakest_if_full(pool)?;
             }
 
-            let (active_lock, consumed_stake, joined_at) = (active_lock, consumed_stake, joined_at);
+            let (active_lock, consumed_stake) = (active_lock, consumed_stake);
 
             match pool.binary_search_by_key(&(amount, who), |pool_item| {
                 (pool_item.stake, &pool_item.court_participant)
@@ -1271,7 +1269,7 @@ mod pallet {
                             stake: amount,
                             court_participant: who.clone(),
                             consumed_stake,
-                            joined_at,
+                            joined_at: now,
                         },
                     )
                     .map_err(|_| {
--- a/zrml/court/src/lib.rs
+++ b/zrml/court/src/lib.rs
@@ -61,7 +61,7 @@ use sp_runtime::{
     DispatchError, Perbill, SaturatedConversion,
 };
 use zeitgeist_primitives::{
-    math::checked_ops_res::CheckedRemRes,
+    math::checked_ops_res::{CheckedAddRes, CheckedRemRes, CheckedSubRes},
     traits::{DisputeApi, DisputeMaxWeightApi, DisputeResolutionApi},
     types::{
         Asset, GlobalDisputeItem, Market, MarketDisputeMechanism, MarketStatus, OutcomeReport,
@@ -1191,36 +1191,25 @@ mod pallet {
     where
         T: Config,
     {
-        fn reassign_participant_reward(
+        fn get_uneligible_stake(
             pool_item_opt: Option<&CourtPoolItemOf<T>>,
             amount: BalanceOf<T>,
-            now: BlockNumberFor<T>,
-        ) -> Result<(BalanceOf<T>, BlockNumberFor<T>), DispatchError> {
+            current_period_index: BlockNumberFor<T>,
+        ) -> Result<BalanceOf<T>, DispatchError> {
             let pool_item = match pool_item_opt {
                 Some(pool_item) => pool_item,
-                None => return Ok((amount, now)),
+                None => return Ok(amount),
             };
 
-            let current_period_index = now
-                .checked_div(&T::InflationPeriod::get())
-                .ok_or(Error::<T>::Unexpected(UnexpectedError::InflationPeriodIsZero))?;
-            let last_reward_period_index = pool_item
-                .pre_period_join_at
-                .checked_div(&T::InflationPeriod::get())
-                .ok_or(Error::<T>::Unexpected(UnexpectedError::InflationPeriodIsZero))?;
-            let last_join_period_index = pool_item
-                .last_join_at
-                .checked_div(&T::InflationPeriod::get())
-                .ok_or(Error::<T>::Unexpected(UnexpectedError::InflationPeriodIsZero))?;
-            if current_period_index != last_reward_period_index {
-                if current_period_index != last_join_period_index {
-                    return Ok((pool_item.stake, pool_item.last_join_at));
-                } else {
-                    return Ok((pool_item.pre_period_join_stake, pool_item.pre_period_join_at));
-                }
+            if current_period_index != pool_item.uneligible_index {
+                let uneligible_stake = amount.checked_sub_res(&pool_item.stake)?;
+                Ok(uneligible_stake)
+            } else {
+                let additional_uneligible_stake = amount.checked_sub_res(&pool_item.stake)?;
+                let uneligible_stake =
+                    pool_item.uneligible_stake.checked_add_res(&additional_uneligible_stake)?;
+                Ok(uneligible_stake)
             }
-
-            Ok((amount, now))
         }
 
         fn get_initial_joined_at(
@@ -1300,7 +1289,7 @@ mod pallet {
             let now = <frame_system::Pallet<T>>::block_number();
 
             let (prev_pool_item_opt, active_lock, consumed_stake) =
-                match <Participants<T>>::get(who) {
+                match <Participants<T>>::get(who) {<
                     Some(prev_p_info) => {
                         let (pruned_pool, old_consumed_stake, prev_pool_item_opt) =
                             Self::handle_existing_participant(who, amount, pool, &prev_p_info)?;
@@ -1313,10 +1302,16 @@ mod pallet {
                     }
                 };
 
-            let (pre_period_join_stake, pre_period_join_at) =
-                Self::reassign_participant_reward(prev_pool_item_opt.as_ref(), amount, now)?;
+            let current_period_index = now
+                .checked_div(&T::InflationPeriod::get())
+                .ok_or(Error::<T>::Unexpected(UnexpectedError::InflationPeriodIsZero))?;
+
+            let uneligible_stake = Self::get_uneligible_stake(
+                prev_pool_item_opt.as_ref(),
+                amount,
+                current_period_index,
+            )?;
             let joined_at = Self::get_initial_joined_at(prev_pool_item_opt.as_ref(), now);
-            let last_join_at = now;
 
             match pool.binary_search_by_key(&(amount, who), |pool_item| {
                 (pool_item.stake, &pool_item.court_participant)
@@ -1337,9 +1332,8 @@ mod pallet {
                             court_participant: who.clone(),
                             consumed_stake,
                             joined_at,
-                            last_join_at,
-                            pre_period_join_stake,
-                            pre_period_join_at,
+                            uneligible_index: current_period_index,
+                            uneligible_stake,
                         },
                     )
                     .map_err(|_| {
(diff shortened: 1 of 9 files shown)


zrml/court/src/lib.rs
```diff
@@ -1356,26 +1424,44 @@ mod pallet {
...
-            for CourtPoolItem { stake, court_participant, joined_at, .. } in pool {
-                if !at_least_one_inflation_period(joined_at) {
+            for CourtPoolItem {
+                court_participant,
+                last_join_at,
+                stake,
+                pre_period_join_at,
+                pre_period_join_stake,
+                ..
+            } in pool
```
sea212, line 1450: Just so it does not get lost: Future optimization to avoid iterating over all pools twice is to cache the total stake.

zrml/court/src/lib.rs
```diff
@@ -1188,6 +1191,101 @@ mod pallet {
...
+            } else {
+                let consumed_stake = prev_p_info.active_lock;
+
+                let pool = Self::remove_weakest_if_full(pool, amount)?;
+
+                Ok((pool, consumed_stake, None))
+            }
+        }
+
         fn do_join_court(
```
sea212, line 1289: Just realizing now that joining and extending stake are intermingled here which leads to a bit more case distinctions and also not intuitive interaction (as an existing Juror has to join twice with the full amount to extend existing stake). Low prio, but should be optimized at some point in the future.

zrml/court/src/lib.rs
```diff
@@ -1188,6 +1191,101 @@ mod pallet {
...
+        }
+
+        fn handle_existing_participant(
+            who: &T::AccountId,
+            amount: BalanceOf<T>,
+            mut pool: CourtPoolOf<T>,
+            prev_p_info: &CourtParticipantInfoOf<T>,
+        ) -> Result<(CourtPoolOf<T>, BalanceOf<T>, Option<CourtPoolItemOf<T>>), DispatchError>
+        {
+            ensure!(amount >= prev_p_info.stake, Error::<T>::AmountBelowLastJoin);
```
sea212, line 1271: Does this imply that jurors can never decrease their stake (as there is no function for specifically that) and exiting and rejoining is the only option?

zrml/court/src/lib.rs
```diff
@@ -1272,6 +1337,9 @@ mod pallet {
                             court_participant: who.clone(),
                             consumed_stake,
                             joined_at,
+                            last_join_at,
+                            pre_period_join_stake,
+                            pre_period_join_at,
```
sea212, line 1342: I am really having a hard time understanding why we need three different join dates. I think we could achieve the same with two and only storing `last_join_inflation_period_index` and `last_join_stake` additionally to existing fields. Then, whenever `pool_join` is called, 
```rust
if last_join_inflation_period_index != cur_idx { 
   last_join_inflation_period_index = cur_idx; 
   last_join_stake = amount - amount_previous ;
} else {
  last_join_stake += amount - amount_previous;
}
```

Now, whenever an inflation payout happens, payout is calculated as follows:
```rust
// cur_idx is the inflation period index for the period that is evaluated.
let payout = if last_join_inflation_period_index != cur_idx { 
  amount;
} else {
  amount - last_join_stake;
}

**Review by sea212 — COMMENTED** [review]
Review still in progress, but so far

File: zrml/court/src/lib.rs (hunk 4 above)
Comment (Chralt98) [reply zrml/court/src/lib.rs:1342]: First of all, let me explain my approach. It uses `pre_period_join_stake` and `pre_period_join_at` to store the relevant information about the stake that already has passed one inflation period, which is the requirement for inflation emission.

Assume there are three buckets A, B and C. Each bucket represents one inflation period. At the end of each inflation period, but before the next inflation period begins, the inflation distribution happens. Your first join happens in A (pre_period_stake = amount; pre_period_join_at = now). Inflation period A ends and no reward is distributed to you. Your second join happens in B (pre_period_stake unchanged; pre_period_join_at unchanged). At the end of B the inflation reward according to your _first_ join is distributed (using pre_period_* information). You don't join in C and the inflation period ends. Now you would expect to get inflation rewards based on your second join. But how would we know the current `stake` was actually already at least one inflation period old? In order to get `last_join_at`, we need to save it for each join (last_join_at is always reassigned to now at every join). If we didn't have `last_join_at`, we wouldn't know if one inflation period passed for the `pool_item.stake`. Here is another edge case: Say you would have joined in C. In this case you would overwrite pre_period_stake and pre_period_join_at by the information of the join in B. But for overwriting `pre_period_join_at` you need to have `last_join_at` and need to check if it was in a previous inflation period. If it was in a previous inflation period, then overwrite `pre_period_join_at`, otherwise use the old `pre_period_join_at`.

Now about your approach:

```rust
if last_join_inflation_period_index != cur_idx { 
  last_join_inflation_period_index = cur_idx;
  ...
else {
  ...
}
```

Ok, this would the last_join_inflation_period_index to be updated at every join.

```rust
if last_join_inflation_period_index != cur_idx { 
  last_join_inflation_period_index = cur_idx;
  last_join_stake = amount - amount_previous;
else {
  last_join_stake = last_join_stake + amount - amount_previous;
}
```

> last_join_stake = amount - amount_previous;

I assume `amount_previous` is the `pool_item.stake` of the last join? In that case `last_join_stake` would be only the **additional** stake since the last join. Why should the additional stake be relevant? We currently reward based on the full stake at a particular point in time.

> last_join_stake = last_join_stake + amount - amount_previous;

The above is done if last_join_inflation_period_index == cur_idx. I can't make sense of this formula. Why would you add the current join `amount` onto the last_join_stake?

```rust
// cur_idx is the inflation period index for the period that is evaluated.
let payout = if last_join_inflation_period_index != cur_idx { 
  amount;
} else {
  amount - last_join_stake;
}
```

I can follow the first case. There you want to use the current `pool_item.stake` for the inflation reward calculation, if the last join happened in a previous inflation period. That is fine.

I am confused about `amount - last_join_stake`, because it would only account for the additional stake since the last join.

In general the idea about storing inflation period indices instead of block numbers can be done.

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1285/commits/ed69c2c9cfc13501458071938fcbb7c7479bae8a

Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1285/commits/8ef4484002ae88f04faaba5aa522fe6dd23478f4

zrml/court/src/migrations.rs
```diff
@@ -14,3 +14,220 @@
...
+                            consumed_stake: old_pool_item.consumed_stake,
+                            joined_at: old_pool_item.joined_at,
+                            uneligible_index: old_pool_item
+                                .joined_at
+                                .checked_div(&T::InflationPeriod::get())
+                                // because inflation period is not zero checked_div is safe
+                                // however, if not use the current period index
+                                // 5_184_000 is the block of the inflation reward distribution
+                                // 5_184_000 / 216_000 (blocks per 30 day inflation period) = 24
+                                .unwrap_or(24u64.saturated_into::<BlockNumberOf<T>>()),
```
sea212, line 88: I think we should error in favor of the user. Setting it to the current period index would result in nobody getting any inflation payouts.

```suggestion
                                .checked_div(&T::InflationPeriod::get())
                                // because inflation period is not zero checked_div is safe
                                // however, if not use the current period index
                                // 5_184_000 is the block of the inflation reward distribution
                                // 5_184_000 / 216_000 (blocks per 30 day inflation period) = 24
                                .unwrap_or(0u64.saturated_into::<BlockNumberOf<T>>()),
```

sea212: Besides that, this would also work for any chain using that pallet.

zrml/court/src/migrations.rs
```diff
@@ -14,3 +14,220 @@
...
+                            // to be uneligible, which is exactly what we want
+                            // if using zero, all joins of period 24 would be eligible,
+                            // but only joins in 23 waited the full period yet
+                            // to understand the calculation of the eligible stake look into handle_inflation
+                            uneligible_stake: old_pool_item.stake,
+                        })
+                        .collect::<Vec<_>>(),
+                )
+            })
+        });
```
sea212, line 99: Interesting, wondering how to benchmark potentially heavy operations on vectors within a storage here. Clearly we only read and write once, but the actual effort scales with the size of the vector. I think in our case it's irrelevant, but generally, what would be the approach? The only solution I see is benchmarking the mutation of the vector.

zrml/court/src/migrations.rs
```diff
@@ -14,3 +14,220 @@
...
+        match res {
+            Ok(_) => log::info!("MigrateCourtPoolItems: Success!"),
+            Err(e) => log::error!("MigrateCourtPoolItems: Error: {:?}", e),
+        }
+
+        total_weight = total_weight.saturating_add(T::DbWeight::get().writes(1));
```
sea212, line 105: One more read for fetching the old court pool vector.

```suggestion
        total_weight = total_weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
```

zrml/court/src/migrations.rs
```diff
@@ -14,3 +14,220 @@
...
+
+    #[cfg(feature = "try-runtime")]
+    fn post_upgrade(_previous_state: Vec<u8>) -> Result<(), &'static str> {
+        let court_pool = crate::CourtPool::<T>::get();
+        log::info!(
+            "MigrateCourtPoolItems: post-upgrade executed. Migrated {:?} pool items",
+            court_pool.len()
+        );
+        Ok(())
+    }
```
sea212, line 127: We should implement this logic to verify that the migrations are successful on Battery Station and Zeitgeist chains.

sea212 (changes requested):

zrml/court/src/lib.rs
```diff
@@ -1188,6 +1191,90 @@ mod pallet {
...
+        fn remove_weakest_if_full(
+            mut p: CourtPoolOf<T>,
+            amount: BalanceOf<T>,
+        ) -> Result<CourtPoolOf<T>, DispatchError> {
+            if p.is_full() {
+                let lowest_item = p.first();
+                let lowest_stake = lowest_item
+                    .map(|pool_item| pool_item.stake)
+                    .unwrap_or_else(<BalanceOf<T>>::zero);
+                debug_assert!(Self::is_sorted_and_all_greater_than_lowest(&p, lowest_stake));
```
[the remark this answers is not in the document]
Chralt98: 3a9520a60139b30bf7176d41460a005b8ac00788

zrml/court/src/lib.rs
```diff
@@ -1188,6 +1191,90 @@ mod pallet {
...
+        #[inline]
+        fn is_sorted_and_all_greater_than_lowest(
+            p: &CourtPoolOf<T>,
+            lowest_stake: BalanceOf<T>,
+        ) -> bool {
+            let mut sorted = p.clone();
+            sorted.sort_by_key(|pool_item| (pool_item.stake, pool_item.court_participant.clone()));
+            p.len() == sorted.len()
+                && p.iter().zip(sorted.iter()).all(|(a, b)| lowest_stake <= a.stake && a == b)
+        }
```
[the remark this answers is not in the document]
Chralt98: 3a9520a60139b30bf7176d41460a005b8ac00788

zrml/court/src/lib.rs
```diff
@@ -1355,34 +1418,41 @@ mod pallet {
...
-            let at_least_one_inflation_period =
-                |joined_at| now.saturating_sub(joined_at) >= T::InflationPeriod::get();
-            let total_stake = pool
-                .iter()
-                .filter(|pool_item| at_least_one_inflation_period(pool_item.joined_at))
-                .fold(0u128, |acc, pool_item| {
-                    acc.saturating_add(pool_item.stake.saturated_into::<u128>())
+            debug_assert!(!inflation_period.is_zero());
+            let current_period_index =
+                now.checked_div(&inflation_period.saturating_add(One::one()));
```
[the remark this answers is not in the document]
Chralt98: good catch!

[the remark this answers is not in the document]
Chralt98: cc03b3d1731b73b5c755f2f16b776c2e1feb713a

zrml/court/src/migrations.rs
Chralt98: 9067fbc17dffd7bd1952b3d76e689e466b5ab515

zrml/court/src/migrations.rs
Chralt98: 9067fbc17dffd7bd1952b3d76e689e466b5ab515

zrml/court/src/tests.rs
```diff
@@ -3089,30 +3111,335 @@ fn handle_inflation_works() {
...
+        let free_charlie_after = Balances::free_balance(CHARLIE);
+        let charlie_reward = 600024000;
+        assert_eq!(free_charlie_after - free_charlie_before, charlie_reward);
+
+        assert_eq!(bob_reward, charlie_reward * 3);
+    });
+}
+
+#[test]
+fn reward_inflation_correct_for_multiple_rejoins() {
```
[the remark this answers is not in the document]
Chralt98: c69d5b9d069e8628d372a1cc9f5ec9bac4aa4929

zrml/court/src/migrations.rs
Chralt98: This is the only solution I see too (benchmarking the migration code in general). I would also always account for the worst case in migration benchmarks, because the migrations are not paid by a user directly as with extrinsics where refunds make sense. In general we never actually accounted for the computation in migrations, only for storage reads and writes, but the clean approach would have been to do exactly this. Apart from the computation made in this mapping here, there is also a bigger encoding and decoding effort for bigger vectors stored in storage. I think Nodle network once made a migration, which exceeded the block limit and led to a chain halt. https://medium.com/nodle-io/latest-parachain-upgrade-what-everyone-should-know-9d5563e97a70

In the article above I see the following solution:

```rust
<Runtime as frame_system::Config>::BlockWeights::get().max_block;
```

This is the true worst weight possible :1st_place_medal: 

It seems using [try-runtime](https://github.com/paritytech/try-runtime-cli/pull/22) allows to see the migration time it took. This means we can watch out for this number if it exceeds the block time.

zrml/court/src/migrations.rs
Chralt98: https://github.com/zeitgeistpm/zeitgeist/pull/1285/commits/3b3762090538b2a5d83ba15f38161e801e905df4

zrml/court/src/migrations.rs
Chralt98: in our case the actual weight was around 0.1% for the try runtime execution

zrml/court/src/lib.rs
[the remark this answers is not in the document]
sea212: Would be good to know from the log message which pool is causing the issue, would significantly reduce the debugging effort.

sea212 (approved):

PR #1288, opened 2024-03-14
AMM-CDA-7 Add event information for indexer for Hybrid Orderbook AMM

Chralt98:
### What does it do?

It adds necessary event information for the indexer and front end.

It contains aggregated amounts for the asset out, the related AMM trades and order book trades plus an optional order amount information, if an order was placed at the end.


zrml/hybrid-router/src/lib.rs
```diff
@@ -122,13 +125,27 @@ mod pallet {
     {
         /// A trade was executed.
         HybridRouterExecuted {
+            /// The type of transaction (Buy or Sell).
```
maltekliemann, line 128: Are the fee amounts contained in the `amount_*` values? This should be clarified in the documentation.

maltekliemann (commented):
I think we should clarify what the fee info in the events means. The tests need to reflect the changes made here.

zrml/hybrid-router/src/lib.rs
Chralt98: 2e58bedbc7609d4cb833c2f2501974fc1d729ec8

Chralt98:
Dropped in favor of this PR https://github.com/zeitgeistpm/zeitgeist/pull/1293/ This was done, because the commit history here led to merged changes would have been needed to be reviewed again.

PR #1291, opened 2024-03-26
Fix Hybrid Router clippy and tests

Chralt98:
### What does it do?

I can successfully run `./scripts/tests/clippy.sh` and `cargo test --all-features` now.


zrml/hybrid-router/Cargo.toml
```diff
@@ -9,36 +9,70 @@ sp-runtime = { workspace = true }
...
+xcm-builder = { workspace = true, optional = true }
+zrml-authorized = { workspace = true, optional = true }
+zrml-court = { workspace = true, optional = true }
+zrml-global-disputes = { workspace = true, optional = true }
+zrml-liquidity-mining = { workspace = true, optional = true }
+zrml-neo-swaps = { workspace = true, optional = true }
+zrml-orderbook = { workspace = true, optional = true }
+zrml-prediction-markets = { workspace = true, optional = true }
+zrml-simple-disputes = { workspace = true, optional = true }
+zrml-swaps = { workspace = true, optional = true }
```
maltekliemann, line 33: We definitely don't need all of these dependencies. Please reduce this to a minimal set and remove/mark everything else as dev dependency.

Chralt98: Your zrml-neo-swaps pallet uses the setup with feature mock too. The reason why I did this was because of compiling the crate `zrml-hybrid-router` with all features. I got this in the mock:

```rust
#[cfg(feature = "parachain")]
zrml_prediction_markets::impl_mock_registry! {
    MockRegistry,
    CurrencyId,
    Balance,
    zeitgeist_primitives::types::CustomMetadata
}
```

`impl_mock_registry` uses feature `mock` and `parachain` together [here](https://github.com/zeitgeistpm/zeitgeist/blob/ed11a332aafdadb90828881ddfe5a00d94265082/zrml/prediction-markets/src/orml_asset_registry.rs#L19)

Chralt98: You meant using the dev dependencies, right?

maltekliemann: Oh, my bad, you're right! Marking them as optional is actually correct. Still though: At the very least you don't need zrml-swaps in the dependencies.

maltekliemann (approved):

PR #1300, opened 2024-04-03
Don't require copyright CI when merging into feature branch

sea212:
### What does it do?
- Creates two distinct auto merge rules for merging into main and merging into another branch.
- Removes the requirement of passing copyright CI when merging into a branch that's not main.


.mergify.yml
```diff
@@ -21,6 +22,28 @@ pull_request_rules:
...
+      - check-success=Test parachain build
+      - check-success=Fuzz
+      - label=s:accepted
+    actions:
+      label:
+        remove:
+          - s:in-progress
+          - s:review-needed
+      queue:
+        merge_method: squash
```
maltekliemann, line 46: Looks fine to me, although I'm wondering if we should really squash when merging into feature branches.

sea212: I think squashing is fine. It will show a reference to the PR primarily and as additional logs all the commits are preserved, so after all no data is lost and in fact it improves readability.

maltekliemann (approved):

PR #1306, opened 2024-04-05
Update lib.rs

ch1kanutt:
Comment or remove:
#![feature(proc_macro_hygiene)]
As it requires rust nightly and barely compatible with arm devices (Termux on Android 11/12)
Hope this attribute is not necessary?


ch1kanutt:
Nightly is necessary anyway :( ?

PR #1294, opened 2024-03-29
AMM-CDA-9 Adds soft and hard failure distinction for AMM and order book errors

Chralt98:

PR #1279, opened 2024-03-04
AMM-CDA-6: Add benchmarks and weights for Hybrid Router

Chralt98:
### What important points should reviewers know?

Maximum Weights:

TODO

at the parameter maximum

The current max block weight for one second is:

```
  perClass: {
    normal: {
      baseExtrinsic: {
        refTime: 99,840,000
        proofSize: 0
      }
      maxExtrinsic: {
        refTime: 324,900,160,000
        proofSize: 3,407,872
      }
      maxTotal: {
        refTime: 375,000,000,000
        proofSize: 3,932,160
      }
      reserved: {
        refTime: 0
        proofSize: 0
      }
    }
```

However my machine is really fast, so it is recommended to reconsider the MaxOrders parameter after we have the weights for our reference machine.

PR #1274, opened 2024-03-01
AMM-CDA-1: Switch to AmmCdaHybrid enum field

Chralt98:
### What does it do?

Instead of order book and amm separate scoring rules, we use a shared enum field for the hybrid approach.

This also includes the resulting storage migration.

Migration was successful on ``try-runtime` for battery station:

```bash
2024-03-04 16:31:49.940  INFO main zrml_market_commons::migrations: All 755 markets could successfully be decoded.    
2024-03-04 16:31:49.948  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Starting...    
2024-03-04 16:31:50.040  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Upgraded 755 markets.    
2024-03-04 16:31:50.041  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Done!    
2024-03-04 16:31:50.215  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Market counter post-upgrade is 755! 
```

And main-net:

```bash
2024-03-04 16:38:56.388  INFO main zrml_market_commons::migrations: All 360 markets could successfully be decoded.    
2024-03-04 16:38:56.391  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Starting...    
2024-03-04 16:38:56.436  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Upgraded 360 markets.    
2024-03-04 16:38:56.436  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Done!    
2024-03-04 16:38:56.517  INFO main zrml_market_commons::migrations: MigrateScoringRuleAmmCdaHybrid: Market counter post-upgrade is 360!
```

PR #1307, opened 2024-04-08
Implement Hybrid Router

Chralt98:
### What does it do?

It implements a router that manages the neo-swaps AMM and the order book for executing buy and sell trades.

The following PRs were incorporated into this PR:

https://github.com/zeitgeistpm/zeitgeist/pull/1274
https://github.com/zeitgeistpm/zeitgeist/pull/1275
https://github.com/zeitgeistpm/zeitgeist/pull/1276
https://github.com/zeitgeistpm/zeitgeist/pull/1277
https://github.com/zeitgeistpm/zeitgeist/pull/1278
https://github.com/zeitgeistpm/zeitgeist/pull/1279
https://github.com/zeitgeistpm/zeitgeist/pull/1288
https://github.com/zeitgeistpm/zeitgeist/pull/1293
https://github.com/zeitgeistpm/zeitgeist/pull/1294


sea212 (dismissed):

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+            external_fee_amount: BalanceOf<T>,
+            /// The swap fee amount paid in the base asset.
+            swap_fee_amount: BalanceOf<T>,
+        },
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// The specified amount is zero.
+        AmountIsZero,
```
maltekliemann, line 189: Can't find a test using this error.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+        /// The price limit is too high.
+        PriceLimitTooHigh,
```
maltekliemann, line 191: Can't find a test using this error.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+        /// The asset of an order is not equal to the maker asset of the order book.
+        AssetNotEqualToOrderBookMakerAsset,
+        /// The asset of an order is not equal to the taker asset of the order book.
+        AssetNotEqualToOrderBookTakerAsset,
+        /// The strategy "immediate or cancel" was applied.
+        CancelStrategyApplied,
+        /// The asset count does not match the markets asset count.
+        AssetCountMismatch,
+        /// The maximum number of orders was exceeded.
+        MaxOrdersExceeded,
```
maltekliemann, line 205: Can't find a test using this error.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+        /// The `orders` vector **must** be sorted in ascending order by the price of their associated
+        /// orders. Failing this, the behavior of `buy` is undefined.
+        ///
+        /// If the maximum price is reached before the entire buy order is filled, the `strategy` parameter
+        /// decides if the order is rolled back (`Strategy::ImmediateOrCancel`) or if a limit order for the
+        /// remaining amount is placed (`Strategy::LimitOrder`).
+        ///
+        /// Complexity: `O(n)`
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::buy(*asset_count as u32, orders.len() as u32))]
```
maltekliemann, line 236: I think we should use `saturating_into` here. This is something that recently popped up in neo-swaps as well. The cast `usize` to `u32` is unsafe on 64 bit systems. This probably has absolutely zero bearing on actual function, but I think using saturating conversion is cleaner.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+        /// The `orders` vector **must** be sorted in ascending order by the price of their associated
+        /// orders. Failing this, the behavior of `sell` is undefined.
+        ///
+        /// If the maximum price is reached before the entire buy order is filled, the `strategy` parameter
+        /// decides if the order is rolled back (`Strategy::ImmediateOrCancel`) or if a limit order for the
+        /// remaining amount is placed (`Strategy::LimitOrder`).
+        ///
+        /// Complexity: `O(n)`
+        #[pallet::call_index(1)]
+        #[pallet::weight(T::WeightInfo::sell(*asset_count as u32, orders.len() as u32))]
```
maltekliemann, line 292: I think we should use saturating_into here. This is something that recently popped up in neo-swaps as well. The cast usize to u32 is unsafe on 64 bit systems. This probably has absolutely zero bearing on actual function, but I think using saturating conversion is cleaner.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+                        return Ok((amount_in, None));
+                    }
+                    T::Amm::calculate_sell_amount_until(market_id, asset, price_limit)?
+                }
+            };
+
+            let amm_amount_in = amm_amount_in.min(amount_in);
+
+            if amm_amount_in.is_zero() {
+                return Ok((amount_in, None));
```
maltekliemann, line 397: This execution path appears to be untested.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+            market_id: MarketIdOf<T>,
+            asset: AssetOf<T>,
+            amount_in: BalanceOf<T>,
+            min_amount_out: BalanceOf<T>,
+        ) -> Result<Option<AmmTradeOf<T>>, DispatchError> {
+            match tx_type {
+                TxType::Buy => {
+                    match T::Amm::buy(who, market_id, asset, amount_in, min_amount_out) {
+                        Ok(amm_trade) => Ok(Some(amm_trade)),
+                        Err(ApiError::SoftFailure(AmmSoftFail::Numerical)) => Ok(None),
```
maltekliemann, line 424: The soft failure execution path appears to be untested. If you go with a super small buy amount for a low price outcome, you should be able to reliably create a soft failure.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+                        Err(ApiError::HardFailure(dispatch_error)) => Err(dispatch_error),
+                    }
+                }
+                TxType::Sell => {
+                    match T::Amm::sell(who, market_id, asset, amount_in, min_amount_out) {
+                        Ok(amm_trade) => Ok(Some(amm_trade)),
+                        Err(ApiError::SoftFailure(AmmSoftFail::Numerical)) => Ok(None),
```
maltekliemann, line 431: This execution path appears to be untested.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+            market_id: MarketIdOf<T>,
+            base_asset: BaseAsset,
+            asset: AssetOf<T>,
+            price_limit: BalanceOf<T>,
+        ) -> Result<OrderAmmTradesInfo<T>, DispatchError> {
+            let mut amm_trades = Vec::new();
+            let mut order_trades = Vec::new();
+            for &order_id in orders {
+                if remaining.is_zero() {
+                    break;
```
maltekliemann, line 466: This execution path appeears to bee untested.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+                }
+
+                let order = match T::OrderBook::order(order_id) {
+                    Ok(order) => order,
+                    Err(_) => continue,
```
maltekliemann, line 471: This execution path appears to be untested.

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+            order_id: OrderId,
+            maker_fill: BalanceOf<T>,
+        ) -> Result<Option<OrderTradeOf<T>>, DispatchError> {
+            match T::OrderBook::fill_order(who, order_id, Some(maker_fill)) {
+                Ok(order_trade) => Ok(Some(order_trade)),
+                Err(ApiError::SoftFailure(OrderbookSoftFail::BelowMinimumBalance))
+                | Err(ApiError::SoftFailure(
+                    OrderbookSoftFail::PartialFillNearFullFillNotAllowed,
+                )) => Ok(None),
+                Err(ApiError::HardFailure(dispatch_error)) => Err(dispatch_error),
```
maltekliemann, line 540: Both of these paths look like they're untested.

zrml/hybrid-router/Cargo.toml
```diff
@@ -0,0 +1,101 @@
...
+]
+try-runtime = [
+    "frame-support/try-runtime",
+]
+
+[package]
+authors = ["Zeitgeist PM <contact@zeitgeist.pm>"]
+edition = "2021"
+name = "zrml-hybrid-router"
+version = "0.4.3"
```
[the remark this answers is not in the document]
Chralt98: a2c71a71b4227996e8b65ddc289c8dbca6c46c07

zrml/hybrid-router/src/lib.rs
```diff
@@ -0,0 +1,761 @@
...
+                Asset = AssetOf<Self>,
+                Balance = BalanceOf<Self>,
+            >;
+
+        /// The maximum number of orders that can be used to execute a trade.
+        #[pallet::constant]
+        type MaxOrders: Get<u32>;
+
+        /// The API to handle the order book.
+        type OrderBook: HybridRouterOrderbookApi<
```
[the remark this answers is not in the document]
Chralt98: b9689966251c53a2e25583682a3e5fe6a99e5ab4

zrml/hybrid-router/src/lib.rs
Chralt98: ba390a211e78700de75a68006d00c516d5d66ffc

zrml/hybrid-router/src/lib.rs
Chralt98: adc78fa35b0fb5d7dcaecf4d0d2ce9f414916d70

zrml/hybrid-router/src/lib.rs
Chralt98: c6f8fc82ca729ce9982933379f312e99808f1632

zrml/hybrid-router/src/lib.rs
Chralt98: Can you elaborate a little bit more on what you mean by `unsafe`? 

> if the length of orders exceeds the maximum value that can be represented by a u32 (which is 4,294,967,295), then the conversion will result in truncation. This means that any values beyond the maximum u32 value will be lost.

In this case, the fees for this transaction would exceed the limit anyways..

Chralt98: fa1a9753edadb9b59f9d9e3f3aa0b7cd0a4077ae

zrml/hybrid-router/src/lib.rs
Chralt98: c84cf22739b4a951789fb3f49d5b38db668f52f1

zrml/hybrid-router/src/lib.rs
Chralt98: 7fb7705781e1efbe08f8398e4df0d9158e8392cc

zrml/hybrid-router/src/lib.rs
Chralt98: dba485133f06b42ed10fcf41a728fd6f2ccd2a0b

zrml/hybrid-router/src/lib.rs
Chralt98: Frankly, I couldn't reproduce a scenario in which this path is executed, since the outer code prevents this. What should we do now? Should we delete the execution path? What I found interesting is that no error was thrown with this test:

```rust
#[test]
fn buy_not_from_amm_since_amm_amount_in_is_zero() {
        let liquidity = _10;
        let spot_prices = vec![_1_2, _1_2];
        let swap_fee = CENT;
        let asset_count = 2u16;
        let market_id = create_market_and_deploy_pool(
            ALICE,
            BASE_ASSET,
            MarketType::Categorical(asset_count),
            liquidity,
            spot_prices.clone(),
            swap_fee,
        );

        let asset = Assets::CategoricalOutcome(market_id, 0);
        let amount = 1;

        let order_maker_amount = _2;
        assert_ok!(AssetManager::deposit(asset, &CHARLIE, order_maker_amount));
        assert_ok!(Orderbook::place_order(
            RuntimeOrigin::signed(CHARLIE),
            market_id,
            Assets::CategoricalOutcome(market_id, 0),
            order_maker_amount,
            BASE_ASSET,
            _1,
        ));

        let order_ids = Orders::<Runtime>::iter().map(|(k, _)| k).collect::<Vec<_>>();
        assert_eq!(order_ids.len(), 1);
        let order_id = order_ids[0];

        let max_price = (_1_2 + 1).saturated_into::<BalanceOf<Runtime>>();
        let orders = vec![order_id];
        let strategy = Strategy::LimitOrder;
        assert_ok!(HybridRouter::buy(
            RuntimeOrigin::signed(ALICE),
            market_id,
            asset_count,
            asset,
            amount,
            max_price,
            orders,
            strategy,
        ));

        let order = Orders::<Runtime>::get(order_id).unwrap();
        assert_eq!(
            order,
            Order {
                market_id,
                maker: CHARLIE,
                maker_asset: Assets::CategoricalOutcome(market_id, 0),
                maker_amount: _2,
                taker_asset: BASE_ASSET,
                taker_amount: _1,
            }
        );
    });
}
```

Here the max price is just plus 1 as compared to the amm spot price which means that the calculated buy amount is just 1u128 in and 2u128 out. Isn't this concerning? It seems like this should be below a threshold, but the AMM just executed the trade. Do you think this is concerning?

maltekliemann: Ah, I See. In that case, I think removing the _outer_ `is_zero` check is the cleaner option because means that `fill_fromyamm` still ensures that that the parameters it is called with are correct.

zrml/hybrid-router/src/lib.rs
Chralt98: d5a36ba563b7ae32411f69c107f0aca67e3b5da9

zrml/hybrid-router/src/lib.rs
Chralt98: 6b3c930a47a2a8c227ccd0e4b629e0a7bc730c21

zrml/hybrid-router/src/lib.rs
maltekliemann: Key word here is _truncation_, as opposed to _saturation_.

```rust
fn main() {
    let x: u64 = u32::MAX as u64 + 1;
    let y: u32 = x as u32;
    println!("{:?}", x); // 4294967296
    println!("{:?}", y); // 0
    println!("{:?}", u32::MAX); // 4294967295
}
```

As you can see `x as u32` does not saturate to `u32::MAX`, which is what we want in this situation, but instead cuts off half of the bit representation of `x` to obtain the cast to `u32`. Since the lower half of the bit representation contains only zeroes, `y` is zero. Which is actually quite dangerous when we're talking about weights.

When I said "unsafe", that was technically wrong, of course, there's no risk of undefined behavior. The behavior is defined, it's just not what we want.

zrml/hybrid-router/src/lib.rs
maltekliemann: Looks good, although I think we need to work on the logic that sets the price of the limit order (not in this PR though). I think if the purchase doesn't move the needle on the AMM, it shouldn't place a limit order at a significantly higher price.

zrml/hybrid-router/src/lib.rs
maltekliemann: bump

zrml/hybrid-router/src/lib.rs
Chralt98: Rust's behavior is wrapping around as a default. Means modulo operation is used. Thanks for the clarification.

zrml/hybrid-router/src/lib.rs
Chralt98: 73c4f578f4de74e85976ae5200808f394b41bd0f

maltekliemann (approved):

PR #1320, opened 2024-04-18
Fix integration tests for tag push

Chralt98:

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -3,8 +3,7 @@ name: Integration Tests
 on:
   workflow_dispatch:
   push:
-    tags:
-      - "^v[0-9]+.[0-9]+.[0-9]+(-rc[0-9]+)?$"
+    branches: [ chralt98-fix-integration-tests ]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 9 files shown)


Chralt98:
I currently have no idea to move forward here https://github.com/zeitgeistpm/zeitgeist/actions/runs/8738968950/job/23982386333 

And after using an extended timeout, I got another error `Error: No response received from RPC endpoint in 60s` [here](https://github.com/zeitgeistpm/zeitgeist/actions/runs/8750122775/job/24015119467?pr=1320). So I changed the RPC endpoint to be dwellir instead of Zeitgeists..

I tried to execute the chopsticks integration tests in consecutive order, but no success. I tried to increase the timeout. I tried to understand the chopsticks logs. I verified the expected file structure. It may be an idea to update moonwall.

Chralt98:
I guess the endpoint of dwellir fixed the problem... /).- You can't understand how happy I am about finally fixing this..

sea212:
> I guess the endpoint of dwellir fixed the problem... /).- You can't understand how happy I am about finally fixing this..

Perhaps some rate limiting issue. We might want to start to work with API keys.

sea212 (approved):

PR #1324, opened 2024-04-27
Integrate new xcm-emulator

sea212:
### What does it do?
Updates the code to work with the new xcm-emulator API and updates the tests accordingly.


runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
```diff
@@ -356,47 +387,51 @@ fn transfer_eth_sibling_to_zeitgeist() {
...
             AssetManager::free_balance(ETH_ID.into(), &ZeitgeistTreasuryAccount::get()),
             // Align decimal fractional places
-            treasury_initial_balance + adjusted_balance(eth(1), eth_fee())
+            expected_treasury
         )
     });
 }
 
 #[test]
 fn transfer_eth_zeitgeist_to_sibling() {
```
maltekliemann, line 407: Maybe this should test that truncation works as expected? Completely random comment, but would it make sense to open an issue about this?

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
```diff
@@ -405,200 +440,147 @@ fn transfer_eth_zeitgeist_to_sibling() {
...
-        let fee_adjusted = adjusted_balance(eth(1), eth_fee());
-        let expected = transfer_amount - fee_adjusted;
+        let expected = bob_initial_balance + transfer_amount - adjusted_balance(eth(1), eth_fee());
+        let expected_sovereign = initial_sovereign_balance - transfer_amount;
 
         // Verify that Bob now has initial balance + amount transferred - fee
-        assert_eq!(Balances::free_balance(&BOB), sibling_bob_initial_balance + expected,);
-
+        assert_eq!(Balances::free_balance(&bob()), expected);
         // Verify that the amount transferred is now subtracted from the zeitgeist account at sibling
```
maltekliemann, line 454: I think we're missing the treasury balance check here.

maltekliemann (approved):

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
sea212: The decimal adjustments are implicitly tested by verifying adjusted balances on accounts using `adjusted_balance`

runtime/zeitgeist/src/integration_tests/xcm/tests/transfers.rs
sea212: I think it's not needed here because from the perspective of `Sibling` it is just receiving it's native token back (ZTG).

Cargo.toml
```diff
@@ -352,6 +357,10 @@ panic = "unwind"
 # Commits:
 #  - cb803be8cdc52d020890308a2076d752a8c4ce5c
 pallet-assets = { git = "https://github.com/zeitgeistpm/polkadot-sdk", branch = "zeitgeist-polkadot-v1.1.0" }
+# xcm-emulator incompatible block number type fixed
+# Commits:
+# -
```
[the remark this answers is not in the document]
sea212: Oh, it's missing the commit hash of the changes. The comments should indicate what the patch is for and which [commits](https://github.com/zeitgeistpm/polkadot-sdk/commit/84163279ce450f608de56b9d5151e65fe4bec51a) deviate from the original repository.

PR #1334, opened 2024-05-06
Update version to v0.5.3

maltekliemann:


runtime/battery-station/src/lib.rs
```diff
@@ -104,8 +104,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("zeitgeist"),
     impl_name: create_runtime_str!("zeitgeist"),
     authoring_version: 1,
-    spec_version: 55,
-    impl_version: 1,
+    spec_version: 56,
+    impl_version: 0,
```
Chralt98, line 108: Why you did change the `impl_version`? Can't we just leave it as the default, which is one?

runtime/zeitgeist/src/lib.rs
```diff
@@ -93,8 +93,8 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("zeitgeist"),
     impl_name: create_runtime_str!("zeitgeist"),
     authoring_version: 1,
-    spec_version: 55,
-    impl_version: 1,
+    spec_version: 56,
+    impl_version: 0,
```
Chralt98, line 97: Why you did change the `impl_version`? Can't we just leave it as the default, which is one?

runtime/battery-station/src/lib.rs
Chralt98: Oh I didn't read `This number can be reverted to 0 after a [spec_version](https://paritytech.github.io/polkadot-sdk/master/cumulus_test_client/sc_executor/struct.RuntimeVersion.html#structfield.spec_version) bump.`

runtime/zeitgeist/src/lib.rs
Chralt98: Oh I didn't read `This number can be reverted to 0 after a [spec_version](https://paritytech.github.io/polkadot-sdk/master/cumulus_test_client/sc_executor/struct.RuntimeVersion.html#structfield.spec_version) bump.`

Chralt98 (approved):

runtime/battery-station/src/lib.rs
maltekliemann: Yeah, it's fairly irrelevant, but it kind of bothered me that it's at 1 for no (?) reason.

Chralt98 (approved):

PR #1335, opened 2024-05-08
Allow HydraDX atomic swap XCMs

Chralt98:


Chralt98:
This PR is abandoned since the current plan is to use USDC from AssetHub instead of a swap solution via HydraDX.

PR #1344, opened 2024-06-11
Remove zrml-simple-disputes and migrate markets

maltekliemann:


Chralt98 (approved):
I don't think the Pallet ID of simple-disputes is still required [here](https://github.com/zeitgeistpm/zeitgeist/blob/43eeda0076984795cbd927357449669b52e05bf4/primitives/src/constants.rs#L117-L118). Changelog for devs addition: Remove simple-disputes from the code base..

Chralt98 (approved):

Chralt98 (approved):

PR #1349, opened 2024-07-22
Merge release v0.5.3

maltekliemann:


Chralt98 (dismissed):
Just a side note. It makes total sense to isolate different products in different repos (e. g. using Solana programs across different repos). Here you can see that all changes of the asset system affected the parimutuel code, although the parimutuel code is not used in production yet. This significantly increases developer time, since all this code needs to be maintained and reviewed too (technical debt). I assume you know this already, but I couldn't be reminded about this more than doing these reviews here.

Chralt98 (approved):
Benchmark weights looking good.

PR #1352, opened 2024-08-12
Remove old migrations and clear pallet storage

maltekliemann:
### What does it do?

Removes old migrations and clears the storage of all previously removed pallets.


Chralt98 (approved):
LGTM. Tested `try-runtime` on Battery Station to figure out how much weight it took. It was about 2% which is fine.

PR #1355, opened 2024-08-13
Update weights

Chralt98:


maltekliemann (approved):

maltekliemann:
`reject_market` and `report_market_...` have lost dependency on parameters.

maltekliemann:
But this seems to happens all the time, and weights always fluctuate pretty wildly so...

PR #1356, opened 2024-08-13
Fix migration & it-tests for release-v0.5.4

Chralt98:
### What does it do?

- Fix migration for battery station to remove markets, which fail to decode
- Replace main-net RPC endpoint for try-runtime
- Correct the pallet storage version for market-commons to match the newest version
- added `bool` flag to `VERBOSE_LOG` to satisfy moonwall

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -788,7 +788,7 @@ checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 
 [[package]]
 name = "battery-station-runtime"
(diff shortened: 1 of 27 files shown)


maltekliemann (commented):
This is currently missing tests. Can we add tests or is this scenario difficult to reproduce?

Also, I think if you're going to hardcode the market IDs, which I understand is a practical approach, you should make this a separate migration which we then _only_ apply to Battery Station using the `with-battery-station-runtime` feature.

Please also change the target branch to `main` and then merge `main` into the release branch after merging this feature branch.

PR #1360, opened 2024-08-22
Fix incorrect documentation of `XcmMetadata::fee_factor`

maltekliemann:
### What does it do?

The documentation incorrectly states that the `fee_factor` uses the same number of decimals that the foreign asset uses. It actually uses the same number of decimals that every asset uses on Zeitgeist - ten.


Chralt98 (approved):
LGTM

spazcoin:
Thanks for clearing this up.

PR #1363, opened 2024-08-29
Fix Rikiddo migration

maltekliemann:
### What does it do?

The pallet prefix for zrml-rikiddo was specified incorrectly in the migration which clears the pallet. This is now mitigated. You can see that clearing the rikiddo pallet now actually deletes a key. That key is for the storage version of the pallet.

It may seem strange that the pallet for custom assets, for instance, doesn't have any keys. This seems to be due to the very different ways these two pallets were initialized, with the `CustomAssets` instance falling victim to the substrate bug that doesn't set the storage version when a pallet is added to a live chain. You can test this by finding what the storage key for the storage version of `CustomAssets` is (`0x97ca552d2b66546c8271d80839408bfd4e7b9012096b41c4eb3aaf947f6ea429`) and then querying it on Battery Station or Zeitgeist. The result is `0x`, i.e. there is no storage under this key. By contrast, the storage key for the storage version of the Rikiddo pallet returns `0x0000`, the correct zero-initialized value.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -96,7 +96,7 @@ macro_rules! decl_common_types {
             pub const CustomAssetsPalletStr: &'static str = "CustomAssets";
             pub const MarketAssetsPalletStr: &'static str = "MarketAssets";
             pub const LiquidityMiningPalletStr: &'static str = "LiquidityMining";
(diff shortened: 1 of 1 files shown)


Chralt98 (approved):
Interesting, good that you caught this. Otherwise the RemovePallet would have been a noop.

PR #1397, opened 2024-12-02
Increase existential deposit

maltekliemann:

### DIFF
--- a/docs/changelog_for_devs.md
+++ b/docs/changelog_for_devs.md
@@ -25,8 +25,8 @@ APIs/RPC interface.
 
 ## v0.5.4
 
-- ⚠️  Remove zrml-liquidity-mining from code base.
-- ⚠️  Remove zrml-rikiddo from the code base.
+- ⚠️ Remove zrml-liquidity-mining from code base.
+- ⚠️ Remove zrml-rikiddo from the code base.
 
 ## v0.5.2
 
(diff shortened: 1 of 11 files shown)


docs/changelog_for_devs.md
```diff
@@ -12,6 +12,16 @@ As of 0.3.9, the changelog's format is based on
...
 
+## v0.5.6
+
+- Increase existential deposit of all non-foreign assets to `BASE`.
+
+## v0.5.5
+
+- Fix Rikiddo migration (#1363)
+- Fix incorrect documentation of XcmMetadata::fee_factor (#1360)
+- Remove old migrations and clear pallet storage (#1352)
```
Chralt98, line 23: Why did you put this doc in this PR?

maltekliemann: Because it was missing, so I thought I'd add it here.

runtime/common/src/lib.rs
```diff
@@ -83,15 +83,79 @@ macro_rules! decl_common_types {
...
+            pub trie_id: TrieId,
+            pub code_hash: CodeHash,
+            pub storage_bytes: u32,
+            pub storage_items: u32,
+            pub storage_byte_deposit: Balance,
+            pub storage_item_deposit: Balance,
+            pub storage_base_deposit: Balance,
+            pub delegate_dependencies:
+                BoundedBTreeMap<CodeHash, Balance, ContractsMaxDelegateDependencies>,
+        }
```
Chralt98, line 111: Verifiable [here](https://github.com/paritytech/polkadot-sdk/blob/65a434a0ed474c14f692dcf9f69f8da66a99d401/substrate/frame/contracts/src/storage.rs#L45-L73) and [here](https://github.com/paritytech/polkadot-sdk/blob/65a434a0ed474c14f692dcf9f69f8da66a99d401/substrate/frame/contracts/src/lib.rs#L156-L157). I see the visibility level only of `trie_id`, `code_hash`, `storage_byte_deposit` to be public, but not the other fields.

Chralt98 (approved):
Looks good to me.

PR #1418, opened 2025-03-21
Fix pool account bug

Chralt98:


Chralt98:
This doesn't matter as long as `deploy_combinatorial_pool` is used from now on. This was Maltes' plan anyways, to use `combo_buy` instead of `buy` and `combo_sell` instead of `sell`. And also use `split` instead of `buy_complete_set` and `merge` instead of `sell_complete_set`.

Chralt98:
I reopen the PR, because if we wanted to have a single pool for one market, it could still be done with `deploy_pool`. Also the fact that Malte left the function and updated it in the Combo & Futarchy PR speaks for its desired usage in the future.

